PR #12, opened 2018-11-28
Integrate gtest and ctest

SgStapleton:


kberryUSGS:
This isn't a very useful comment, but I looked at this and I have no idea why this isn't building on Travis on OSX. It looks like the cmake setup and .travis.yml here is virtually identical to what we used for CSM-CameraModel.

jlaura:
Is this clang not using c++11?

jlaura:
ehhh I see the CXX11 standard in the cmake lists. Another option might be how gtest is being pulled. Is the commit that came in as a submodule bleeding edge or a release?

kberryUSGS:
@jlaura The pulling the bleeding edge of gtest was it! Pulling release-1.8.1 fixes this problem.

kberryUSGS:
I wasn't sure of the best place to fix this, so I put in a separate PR with just the gtest submodule changes: PR #22

SgStapleton:
So, it looks like travis failed to build for macos, but it is actually passing with @kberryUSGS 's changes (https://github.com/USGS-Astrogeology/eal/pull/12/checks?check_run_id=36105146).

Kelvinrr:
Looks good :+1:

PR #13, opened 2018-11-28
Add .travis.yml to do builds

### LINKED ISSUE #3: Setup Continuous Integration

kberryUSGS:
This initial .travis.yml results in successful builds on Travis. Testing will need to be added after the gtest PR #12 comes in.

It was necessary to modify the .gitmodules to use https rather than ssh to resolve a permission denied error on Travis.

This may fix #3 or we may want to keep it open until tests are added.


SgStapleton:
Good catch on updating the submodule to https! Everything looks good.

PR #26, opened 2018-11-29
Simple JSON input

SgStapleton:
Very simple implementation that can be modified while format is being established. Related to issue https://github.com/USGS-Astrogeology/eal/pull/26.


SgStapleton:
@kberryUSGS It is in the include/json directory that is part of this PR. This is the same set-up they had established for the CSM-CameraModel.

SgStapleton:
This code may not be needed alongside @acpaquette 's implementation of eal.cpp?

PR #27, opened 2018-11-29
Basic Library API

### LINKED ISSUE #25: Skeleton Code
Adds basic files/classes/functions for the api

acpaquette:
Basic definition of the EAL api and a few name changes. Closes #25.

### DIFF
--- a/include/eal.h
+++ b/include/eal.h
@@ -1,0 +1,15 @@
+#ifndef EAL_H
+#define EAL_H
+
+#include <vector>
+#include <string.h>
+
+using namespace std;
+
+vector<double> getPosition(vector<vector<double>> coords, vector<double> coeffs, char *interp, double time);
+vector<double> getVelocity(vector<vector<double>> coords, vector<double> coeffs, char *interp, double time, bool interpolation);
+
+vector<double> getPosition(vector<double> coeffs, char *interp, double time);
+vector<double> getVelocity(vector<double> coeffs, char *interp, double time, bool interpolation);
+
+#endif // EAL_H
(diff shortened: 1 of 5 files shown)


include/eal.h
```diff
@@ -0,0 +1,21 @@
+#ifndef EAL_H
+#define EAL_H
+
+#include <vector>
+#include <string.h>
+
+using namespace std;
+
+vector<double> getPosition(vector<vector<double>> coords, vector<double> coeffs, char *interp, double time);
+vector<double> getVelocity(vector<vector<double>> coords, vector<double> coeffs, char *interp, double time, bool interpolation);
```
jessemapel: Why are both coordinates and coefficients being passed in? I'd expect coordinates and times to be passed in. Also, the input data vectors should be passed as constant references to avoid excessive copying.

include/eal.h
```diff
@@ -0,0 +1,21 @@
...
+
+vector<double> getPosition(vector<double> coeffs, char *interp, double time);
+vector<double> getVelocity(vector<double> coeffs, char *interp, double time, bool interpolation);
```
jessemapel: We should pass the interpolation as an enumeration instead of a c-string. This will help avoid string parsing and allow us to use a switch statement in the implementation if we want to.

include/eal.h
```diff
@@ -0,0 +1,21 @@
...
+
+vector<double> getRotation(char *from, char *to, vector<vector<double>> rotations, vector<double> times, char *interp, double time);
+vector<double> getAngularVelocity(char *from, char *to, vector<vector<double>> rotations, vector<double> times, char *interp, double time);
+
+vector<double> getRotation(char *from, char *to, vector<double> coefficients, char *interp, double time);
+vector<double> getAngularVelocity(char *from, char *to, vector<double> coefficients, char *interp, double time);
```
jessemapel: Is there a reason for using c-strings instead of std::string?

PR #29, opened 2018-11-30
Cleaned up API

jessemapel:


include/eal.h
```diff
@@ -1,21 +1,25 @@
...
+  vector<double> getVelocity(vector<vector<double>> coords, vector<double> times, string interp, double time, bool interpolation);
 
-vector<double> getRotation(char *from, char *to, vector<vector<double>> rotations, vector<double> times, char *interp, double time);
-vector<double> getAngularVelocity(char *from, char *to, vector<vector<double>> rotations, vector<double> times, char *interp, double time);
+  vector<double> getPosition(vector<double> coeffs, string interp, double time);
+  vector<double> getVelocity(vector<double> coeffs, string interp, double time, bool interpolation);
 
-vector<double> getRotation(char *from, char *to, vector<double> coefficients, char *interp, double time);
-vector<double> getAngularVelocity(char *from, char *to, vector<double> coefficients, char *interp, double time);
+  vector<double> getRotation(string from, string to, vector<vector<double>> rotations, vector<double> times, string interp, double time);
```
kberryUSGS: Do we want a `bool interpolation` on this one?

kberryUSGS: I guess I'm not sure what we would do if =false. "Sorry, no rotation data at this point?"

jessemapel: I think the getAngularVelocity need bool interpolation options. I don't think this one does.

kberryUSGS: Agreed, that makes more sense.

jessemapel: I'll make that change and push it up

PR #45, opened 2018-12-04
implemented getVelocity

Kelvinrr:
Moved some things to reduce code duplication.

### DIFF
--- a/include/eal.h
+++ b/include/eal.h
@@ -5,21 +5,25 @@
 #include <string>
 #include <vector>
 
+#include <gsl/gsl_interp.h>
+
 namespace eal {
 
-  enum interpolation {
-    linear,
-    spline
+  enum derivation {
+    none,
+    first,
+    second
   };
 
   nlohmann::json constructStateFromIsd(const std::string positionRotationData);
 
   std::vector<double> getPosition(std::vector<std::vector<double>> coords,
                                   std::vector<double> times,
-                                  interpolation interp, double time);
+                                  double time, const gsl_interp_type* interp);
+
   std::vector<double> getVelocity(std::vector<std::vector<double>> coords,
                                   std::vector<double> times,
-                                  interpolation interp, double time);
+                                  double time, const gsl_interp_type* interp);
 
   std::vector<double> getPosition(std::vector<double> coeffs, double time);
   std::vector<double> getVelocity(std::vector<double> coeffs, double time);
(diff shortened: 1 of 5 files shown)


include/eal.h
```diff
@@ -5,41 +5,46 @@
...
 namespace eal {
 
-  enum interpolation {
-    linear,
-    spline
+  enum derivation {
+    none,
+    first,
+    second
   };
```
Kelvinrr: Enum for derivations might be unnecessary? Just use numbers directly?

include/eal.h
```diff
@@ -5,41 +5,46 @@
...
 
   std::vector<double> getPosition(std::vector<std::vector<double>> coords,
                                   std::vector<double> times,
-                                  interpolation interp, double time);
+                                  double time, const gsl_interp_type* interp);
+
   std::vector<double> getVelocity(std::vector<std::vector<double>> coords,
                                   std::vector<double> times,
-                                  interpolation interp, double time);
+                                  double time, const gsl_interp_type* interp);
```
Kelvinrr: Since we are using GSL, I imagine it is unnecessary to redefine an enum if it already exists in gsl.

src/eal.cpp
```diff
@@ -106,59 +105,45 @@ namespace eal {
...
-    }
-    if (time < times.front() || time > times.back()) {
-      throw invalid_argument("Invalid interpolation time, outside of input times.");
-    }
-
-    // GSL setup
-    gsl_interp *interpolator = gsl_interp_alloc(gsl_interp_cspline, numPoints);
-    gsl_interp_init(interpolator, &times[0], &points[0], numPoints);
-    gsl_interp_accel *acc = gsl_interp_accel_alloc();
+ double interpolate(vector<double> points, vector<double> times, double time, const gsl_interp_type* interp, derivation d) {
```
Kelvinrr: Smashed the different interpolation functions into one since they all have the same code with the exception of the enum val you pipe into GSL.

Kelvinrr:
Is it really necessary to test these functions if they all dispatch to GSL? Shouldn't it only test our own integrity checks (e.g. `if (numPoints < 2) {throw something}`)?

include/eal.h
kberryUSGS: I would prefer to keep GSL-specific enums out of our API, especially since Jesse mentioned that GSL doesn't support hermite splines, and we'll need to add that. @jessemapel What do you think? 

I was picturing the enum getting extended to something like: 

```
enum interpolation {
linear,
spline,
hermite
}
```

And maybe more!

jessemapel: I agree. I consumers should not need to include the GSL headers with our library.

jessemapel: This also makes it so if we end up using something other than GSL, the API does not need to change.

src/eal.cpp
jessemapel: :+1:

include/eal.h
jessemapel: I think using numbers is fine. 0, 1, 2, ...

include/eal.h
kberryUSGS: Agreed.

jessemapel:
I think that we do need to test that we are properly dispatching to GSL or whatever library we are using. I'm not sure how to do that despite checking what numbers get sent back.

Kelvinrr:
If we insist on making sure we hit the GSL calls, could we at least attempt to mock out GSL and other rmath library calls first?

jessemapel:
@Kelvinrr I think that's a totally fine way to go.

Issue #36, opened 2018-11-30
Implement get velocity from function coefficients


jessemapel:

kberryUSGS:
Are the function coefficients we pass in to this function coefficients for a velocity function, or are we passing the coefficients for a position function in to this function? I'm asking because if they are coefficients to a velocity function, does this basically do the same thing as the `getPosition` function?

jessemapel:
They are position function coefficients, so we need to evaluate the derivative.

kberryUSGS:
@jessemapel :+1 Thanks, I will go ahead and do this. That does make more sense -- I wasn't sure why / how the consumer of this library was likely to be able to provide coefficients for a velocity polynomial?

jessemapel:
Done by #52

PR #46, opened 2018-12-04
Add getPosition with coefficients

### LINKED ISSUE #33: Implement get position from function coefficients

kberryUSGS:
This should fix #33 .

Please ignore the getVelocity function for now. I'll update it to take coeffs for a position fn and calculate the derivative in a new PR for #36 

I changed the coefficients vector to be a vector of vectors so coefficients could be provided for x, y, z.

### DIFF
--- a/src/eal.cpp
+++ b/src/eal.cpp
@@ -2,6 +2,7 @@
 
 #include <json.hpp>
 
+#include <iostream>
 #include <gsl/gsl_interp.h>
 #include <gsl/gsl_spline.h>
 
@@ -75,12 +76,20 @@ namespace eal {
   //                y = cy_n * t^n + cy_n-1 * t^(n-1) + ... + cy_0
   //                z = cz_n * t^n + cz_n-1 * t^(n-1) + ... + cz_0
   vector<double> getPosition(vector<vector<double>> coeffs, double time) {
+    if ( time < 0.0) {
+      throw invalid_argument("Invalid input time, must be non-negative.");
+    }
 
     if (coeffs.size() != 3) {
       throw invalid_argument("Invalid input coeffs, expected three vectors.");
     }
 
     // make sure all coeffs sizes are equal, else throw error...
+    if ((coeffs[0].size() != coeffs[1].size()) || 
+        (coeffs[1].size() != coeffs[2].size())) {
+      throw invalid_argument("Invalid input coeffs, each must represent the same degree polynomial.");
+    }
+
     int degree = coeffs[0].size() - 1; 
 
     vector<double> coordinate = {0.0, 0.0, 0.0};
@@ -94,8 +103,19 @@ namespace eal {
     return coordinate;
   }
 
+
+  // Velocity Function Functions
+  // vector<double> coeffs = [[cvx_n, cvx_n-1, cvx_n-2, ... cvx_0],
+  //                          [cvy_n, cvy_n-1, cvy_n-2, ... cvy_0],
+  //                          [cvz_n, cvz_n-1, cvz_n-2, ... cvz_0]]
+  // The equations evaluated by this function are:
+  //                v_x = cvx_n * t^n + cvx_n-1 * t^(n-1) + ... + cvx_0
+  //                v_y = cvy_n * t^n + cvy_n-1 * t^(n-1) + ... + cvy_0
+  //                v_z = cvz_n * t^n + cvz_n-1 * t^(n-1) + ... + cvz_0
   vector<double> getVelocity(vector<vector<double>> coeffs, double time) {
-    vector<double> coordinate = {0.0, 0.0, 0.0};
+    // This is the same calculation, for now? A polynomial is a polynomial -- should this be
+    // something else? 
+    vector<double> coordinate = getPosition(coeffs, time);
     return coordinate;
   }
 
--- a/src/eal.cpp
+++ b/src/eal.cpp
@@ -4,6 +4,7 @@
 
 #include <gsl/gsl_interp.h>
 #include <gsl/gsl_spline.h>
+#include <gsl/gsl_poly.h>
 
 #include <string>
 #include <stdexcept>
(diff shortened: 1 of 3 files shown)


kberryUSGS:
Fixes #33.

src/eal.cpp
```diff
@@ -68,13 +67,54 @@ namespace eal {
...
+    }
+
+    if (coeffs.size() != 3) {
+      throw invalid_argument("Invalid input coeffs, expected three vectors.");
+    }
+
+    // make sure all coeffs sizes are equal, else throw error...
+    if ((coeffs[0].size() != coeffs[1].size()) || 
+        (coeffs[1].size() != coeffs[2].size())) {
+      throw invalid_argument("Invalid input coeffs, each must represent the same degree polynomial.");
```
kberryUSGS: This is kind of arbitrary, but if you had a 2nd degree solution for x, but 4 for z, and 0 for y, we could say: use the highest degree solution to pick the size, and then put in 0s for everything else.

src/eal.cpp
```diff
@@ -68,13 +67,54 @@ namespace eal {
...
+    }
+
+    int degree = coeffs[0].size() - 1; 
+
     vector<double> coordinate = {0.0, 0.0, 0.0};
+    
+    for (unsigned i=0; i < coeffs[0].size(); i++) {
```
kberryUSGS: Could use `gsl_poly_eval` instead. I did this 'by-hand' because it was so straightforward -- any opinions?

src/eal.cpp
```diff
@@ -68,13 +67,54 @@ namespace eal {
   }
 
   // Postion Function Functions
-  vector<double> getPosition(vector<double> coeffs, double time) {
+  // vector<double> coeffs = [[cx_n, cx_n-1, cx_n-2, ... cx_0],
```
kberryUSGS: Any opinions about the order I put these in? It's the opposite of the order that gsl uses. My reason for picking this order was just that I'm used to seeing equations written in this order: `2x^2 + 3x + 1 = 0`.

SgStapleton: I have to say, this is what I am used to seeing/using as well, but reading through GSL's documentation yesterday I saw they were opposite as well and found this interesting. I wonder if there is some standard...?

src/eal.cpp
```diff
@@ -68,13 +67,54 @@ namespace eal {
...
+  // Velocity Function Functions
+  // vector<double> coeffs = [[cvx_n, cvx_n-1, cvx_n-2, ... cvx_0],
+  //                          [cvy_n, cvy_n-1, cvy_n-2, ... cvy_0],
+  //                          [cvz_n, cvz_n-1, cvz_n-2, ... cvz_0]]
+  // The equations evaluated by this function are:
+  //                v_x = cvx_n * t^n + cvx_n-1 * t^(n-1) + ... + cvx_0
+  //                v_y = cvy_n * t^n + cvy_n-1 * t^(n-1) + ... + cvy_0
+  //                v_z = cvz_n * t^n + cvz_n-1 * t^(n-1) + ... + cvz_0
+  vector<double> getVelocity(vector<vector<double>> coeffs, double time) {
+    // This is the same calculation, for now? A polynomial is a polynomial -- should this be
```
kberryUSGS: Please ignore this for now. I'll update this to accept coeffs for the position function and calculate the derivative in a separate PR.

src/eal.cpp
jessemapel: One reason to use `1 + 3x + 2x^2` is because as you're iterating over the coefficients, you can use the index as your exponent.

```
for (int i = 0; i < coeff.size(), i++) {
  result += coeff[i] * pow(time, i);
}
```

src/eal.cpp
```diff
@@ -68,13 +67,54 @@ namespace eal {
...
+  // vector<double> coeffs = [[cx_n, cx_n-1, cx_n-2, ... cx_0],
+  //                          [cy_n, cy_n-1, cy_n-2, ... cy_0],
+  //                          [cz_n, cz_n-1, cz_n-2, ... cz_0]]
+  // The equations evaluated by this function are:
+  //                x = cx_n * t^n + cx_n-1 * t^(n-1) + ... + cx_0
+  //                y = cy_n * t^n + cy_n-1 * t^(n-1) + ... + cy_0
+  //                z = cz_n * t^n + cz_n-1 * t^(n-1) + ... + cz_0
+  vector<double> getPosition(vector<vector<double>> coeffs, double time) {
+    if ( time < 0.0) {
+      throw invalid_argument("Invalid input time, must be non-negative.");
```
jessemapel: Why enforce non-negative time?

src/eal.cpp
jessemapel: Another reason to put the constant first is you can arbitrarily pad with 0s without changing anything.

src/eal.cpp
jessemapel: I think we can allow for any size, just require that they are all non-empty

src/eal.cpp
jessemapel: I really don't have an opinion either way.

src/eal.cpp
```diff
@@ -68,13 +67,54 @@ namespace eal {
...
+  //                          [cvy_n, cvy_n-1, cvy_n-2, ... cvy_0],
+  //                          [cvz_n, cvz_n-1, cvz_n-2, ... cvz_0]]
+  // The equations evaluated by this function are:
+  //                v_x = cvx_n * t^n + cvx_n-1 * t^(n-1) + ... + cvx_0
+  //                v_y = cvy_n * t^n + cvy_n-1 * t^(n-1) + ... + cvy_0
+  //                v_z = cvz_n * t^n + cvz_n-1 * t^(n-1) + ... + cvz_0
+  vector<double> getVelocity(vector<vector<double>> coeffs, double time) {
+    // This is the same calculation, for now? A polynomial is a polynomial -- should this be
+    // something else? 
+    vector<double> coordinate = getPosition(coeffs, time);
```
jessemapel: This should compute the derivative, so almost the same computation.

src/eal.cpp
kberryUSGS: :+1: will do

src/eal.cpp
kberryUSGS: @jessemapel Good point. Especially with the switch to allow any size, I'll go ahead and swap the order.

src/eal.cpp
kberryUSGS: :+1:

kberryUSGS:
@jessemapel :+1: Can do with the polynomial evaluation.

src/eal.cpp
```diff
@@ -68,13 +67,54 @@ namespace eal {
...
+  // vector<double> coeffs = [[cvx_n, cvx_n-1, cvx_n-2, ... cvx_0],
+  //                          [cvy_n, cvy_n-1, cvy_n-2, ... cvy_0],
+  //                          [cvz_n, cvz_n-1, cvz_n-2, ... cvz_0]]
+  // The equations evaluated by this function are:
+  //                v_x = cvx_n * t^n + cvx_n-1 * t^(n-1) + ... + cvx_0
+  //                v_y = cvy_n * t^n + cvy_n-1 * t^(n-1) + ... + cvy_0
+  //                v_z = cvz_n * t^n + cvz_n-1 * t^(n-1) + ... + cvz_0
+  vector<double> getVelocity(vector<vector<double>> coeffs, double time) {
+    // This is the same calculation, for now? A polynomial is a polynomial -- should this be
+    // something else?
```
SgStapleton: Oh - hahaha!

tests/EalTest.cpp
```diff
@@ -142,3 +142,37 @@ TEST(SplineInterpTest, Extrapolate) {
   EXPECT_THROW(eal::splineInterpolate(data, times, 4.0),
                invalid_argument);
 }
+
+TEST(PoisitionCoeffTest, ValidInput) {
+  double time = 2.0; 
+  vector<vector<double>> coeffs = {{3.0, 2.0, 1.0},
+                                   {2.0, 3.0, 1.0},
+                                   {1.0, 2.0, 3.0}};
```
SgStapleton: Would we maybe want to check negative coefficient values...?

src/eal.cpp
kberryUSGS: Naively, it didn't seem meaningful to have a negative time

jessemapel: CSM uses negative times, because it does image time relative to the center of the image. We should support that usage.

src/eal.cpp
kberryUSGS: Using `gsl_poly_eval` after discussion in standup.

src/eal.cpp
kberryUSGS: :+1:

tests/EalTest.cpp
kberryUSGS: @SgStapleton :+1:

PR #48, opened 2018-12-06
getRotation and getAngularVelocity from points implementation 

Kelvinrr:
First draft using Eigen, only using linear interpolation.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,7 +35,6 @@ target_include_directories(eal
                            ${GSL_INCLUDE_DIRS}
                            ${EIGEN3_INCLUDE_DIR}
                            PUBLIC
-                           ${EIGEN3_INCLUDE_DIR}
                            ${EAL_INCLUDE_DIRS})
 
 # Setup for GoogleTest
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,8 +20,6 @@ set(CMAKE_CXX_STANDARD 11)
 # Third Party Dependencies
 find_package(GSL)
 
-message(STATUS ${GSL_INCLUDE_DIRS})
-
 # Library setup
 add_library(eal SHARED
             ${CMAKE_CURRENT_SOURCE_DIR}/src/eal.cpp)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,6 +20,8 @@ set(CMAKE_CXX_STANDARD 11)
 # Third Party Dependencies
 find_package(GSL)
 
+message(STATUS ${GSL_INCLUDE_DIRS})
+
 # Library setup
 add_library(eal SHARED
             ${CMAKE_CURRENT_SOURCE_DIR}/src/eal.cpp)
--- a/src/eal.cpp
+++ b/src/eal.cpp
@@ -31,7 +31,7 @@ namespace eal {
 
   // Position Data Functions
   vector<double> getPosition(vector<vector<double>> coords, vector<double> times, double time,
-                             const gsl_interp_type* interp) {
+                             interpolation interp) {
     // Check that all of the data sizes are okay
     // TODO is there a cleaner way to do this? We're going to have to do this a lot.
     if (coords.size() != 3) {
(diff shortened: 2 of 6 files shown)


src/eal.cpp
```diff
@@ -80,32 +83,62 @@ namespace eal {
...
+    // memory. Should be easy to switch to a contiguous column-major format
+    // if we stick with Eigen.
+    double data[] = {0,0,0,0};
+    for (size_t i = 0; i<rotations[0].size(); i++) {
+      Eigen::Quaterniond quat(rotations[0][i], rotations[1][i], rotations[2][i], rotations[3][i]);
+      quat.normalize();
+      rotations[0][i] = quat.w();
+      rotations[1][i] = quat.x();
+      rotations[2][i] = quat.y();
+      rotations[3][i] = quat.z();
```
Kelvinrr: Bad implementation, total GSD edition. I believe the correct way to do this would be to use column-major vectors if we do decide to use Eigen for matrix operations so we can avoid unnecessary allocs and copies.

src/eal.cpp
```diff
@@ -80,32 +83,62 @@ namespace eal {
...
   vector<double> getVelocity(vector<vector<double>> coeffs, double time) {
     vector<double> coordinate = {0.0, 0.0, 0.0};
     return coordinate;
   }
 
-  // Rotation Data Functions
 
   // Rotation Data Functions
-  vector<double> getRotation(string from, string to, vector<vector<double>> rotations,
+  vector<double> getRotation(vector<vector<double>> rotations,
```
Kelvinrr: From what I understand, this should be similar to positional interpolation except for a 4th dimension and extra normalization steps. Could be wrong.

kberryUSGS: I don't have a complete grasp of this either, but it looks like eigen also has a specific method for interpolating between quaternions called `slerp` for spherical linear interpolation: 
http://eigen.tuxfamily.org/dox/classEigen_1_1QuaternionBase.html#ac840bde67d22f2deca330561c65d144e
See: https://en.wikipedia.org/wiki/Slerp

kberryUSGS: Googling around, there are some discussions on stack overflow about how doing a normal linear interpolation can have undesired results, but these are all in the context of using quaternions for graphics, so I'm not sure whether or not this applies to our usecase.

Kelvinrr: Yes, so there are two ways to perform linear interpolation on quaternions: slerp and lerp, which are two different forms of interpolation (and from what I know, this form of straight linear interpolation is lerp, if we want spherical, then slerp will do). The gist of what I want to know: Do we one only one or the other for whatever reason,

Kelvinrr: http://web.mit.edu/2.998/www/QuaternionReport1.pdf 

pg. 40-43 seem to explain the two and the results

kberryUSGS: :+1: On the link to the QuaternionReport resource. I don't know the answer to your question, but would lean toward both?

Kelvinrr: From reading, the main reason to not care about slerp is if our angles have predominately small deltas. In which case, the slower slerp won't really win us anything.

Kelvinrr: Another good write up on differences http://number-none.com/product/Hacking%20Quaternions/.  Quantifies how inaccurate lerp actually is to slerp along with other pros in cons. Keep in mind that this is from a game developers perspective that care way more about real-time efficiency than we do.

jessemapel: I would lean towards allowing both methods. For rigorous scientific computations I'd be tempted to use slerp. For quick rough computations lerp may be worth the loss in efficiency.

I think implementing slerp interpolation can be handled in another PR.

CMakeLists.txt
```diff
@@ -28,10 +29,13 @@ set_target_properties(eal PROPERTIES
     SOVERSION 1)
 set(EAL_INCLUDE_DIRS "${CMAKE_CURRENT_SOURCE_DIR}/include/"
                      "${CMAKE_CURRENT_SOURCE_DIR}/include/json")
+
 target_include_directories(eal
                            PRIVATE
                            ${GSL_INCLUDE_DIRS}
+                           ${EIGEN3_INCLUDE_DIR}
                            PUBLIC
+                           ${EIGEN3_INCLUDE_DIR}
```
jessemapel: These include directories don't need to be both public and private.

src/eal.cpp
```diff
@@ -80,34 +83,89 @@ namespace eal {
...
+    // GSL setup
+    vector<double> coordinate = {0.0, 0.0, 0.0, 0.0};
+
+    coordinate = { interpolate(rotations[0], times, time, interp, 1),
+                   interpolate(rotations[1], times, time, interp, 1),
+                   interpolate(rotations[2], times, time, interp, 1),
+                   interpolate(rotations[2], times, time, interp, 1)};
+
+    Eigen::Quaterniond quat(coordinate.data());
+    quat.normalize();
```
jessemapel: Does this modify the internal data array of coordinate?

I'd expect the Eigen quaternion constructor to copy the contents of the array.

Kelvinrr: If I understand he docs, it uses the array as the internal buffer, so it should work with the pointer directly instead of copying it. It's why I suggested changing our vector format, we can stuff a std::vector into an eigen::Matrix with only copying a pointer. In all fairness I need to test it first since I am only going off what I read.

src/eal.cpp
Kelvinrr: :+1:

src/eal.cpp
Kelvinrr: Needed to use Eigen::Map to get this functionality, why only copying the pointer isn't the default behavior for Eigen matrices I don't know.

jessemapel:
Fixes #37 and #32

PR #58, opened 2018-12-11
Added first steps at cube reading notebook

jessemapel:
To run the notebook you'll need the following:

jupyter, numba, quaternion, and pvl


SgStapleton:
This is a great start!

PR #59, opened 2018-12-11
ISD IO

acpaquette:
Adds mini version of pfeffernusse to eal. Only serious dependency is on the config files to point to spice kernels.


Kelvinrr:
Tests and env file still need to be carried over, but that can be done in a different PR.

PR #60, opened 2018-12-12
Moved in pfeffer tests, name change to ale in the codebase 

Kelvinrr:
Also updated environment file and removed autogenerated code that came over from pfeffernusse. 

Name change of repo should happen after this is merged.


environment.yml
```diff
@@ -1,11 +1,16 @@
...
+  - python=3.7.1
   - cmake>=3.10
   - gsl
   - openblas>=0.3.0
   - eigen
+  - spiceypy
+  - numpy
+  - python-dateutil
+  - pip:
+    - pvl
```
jlaura: We should be able to get PVL from: https://anaconda.org/conda-forge/pvl

CMakeLists.txt
```diff
@@ -1,13 +1,13 @@
...
-#      The main build file for building EAL using CMake.
+#      The main build file for building ale using CMake.
 #===============================================================================
 # CMake initialization
 
 # Specify the required version of CMake.
 # cmake 3.10 required for ctest/gtest integration
 cmake_minimum_required(VERSION 3.10)
-project(eal VERSION 0.0.1 DESCRIPTION "Ephemeris Abstraction Library")
+project(ale VERSION 0.0.1 DESCRIPTION "Ephemeris Abstraction Library")
```
jlaura: Below the name was changed to Abstraction Library for Ephemerides I think - we should be consistent.

PR #61, opened 2018-12-12
Name change in README

Kelvinrr:
missed this in the last PR.


jlaura:
I think you need to pull in the merged changes and rebase. Then this will just touch the README?

PR #62, opened 2018-12-12
Fixed angular velocity interpolation

jessemapel:
We were only returning the derivative of the quaternion. I got the corrected formula from ftp://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/req/rotation.html#Obtaining%20angular%20velocity%20from%20quaternions


tests/EalTest.cpp
```diff
@@ -22,19 +24,6 @@ TEST(PositionInterpTest, LinearInterp) {
...
-                                 {0, 1, 2, 3},
-                                 {0, 2, 1, 0}};
-
-  vector<double> coordinate = eal::getPosition(data, times, 0.5, eal::spline);
-
-  ASSERT_EQ(3, coordinate.size());
-  EXPECT_DOUBLE_EQ(0,      coordinate[0]);
-  EXPECT_DOUBLE_EQ(0.5,   coordinate[1]);
-  EXPECT_DOUBLE_EQ(2.8 * 0.5 - 0.8 * 0.125, coordinate[2]);
-}
```
jessemapel: I took this test out because we are testing the spline interpolation separately now

tests/EalTest.cpp
```diff
@@ -233,23 +235,40 @@ TEST(VelocityCoeffTest, InvalidInput) {
...
-  vector<vector<double>> rots({{1,1,1,1}, {0,0,0,0}, {1,1,1,1}, {0,0,0,0}});
+  vector<vector<double>> rots({{0,1,0,0}, {0,0,1,0}, {0,0,0,1}, {1,0,0,0}});
   vector<double> r = eal::getRotation(rots, times, 2, eal::linear);
+  Eigen::Quaterniond quat(r[0], r[1], r[2], r[3]);
 
-  ASSERT_NEAR(0.707107,  r[0], 0.000001);
-  EXPECT_DOUBLE_EQ(0,  r[1]);
-  ASSERT_NEAR(0.707107, r[2], 0.000001);
-  EXPECT_DOUBLE_EQ(0, r[3]);
+  EXPECT_DOUBLE_EQ(1, quat.norm());
```
jessemapel: Modified to just use Eigen to check that the output is normalized.

tests/ctests/AleTest.cpp
```diff
@@ -233,23 +235,40 @@ TEST(VelocityCoeffTest, InvalidInput) {
...
+  vector<vector<double>> rots({{0,0}, {1,0}, {0,1}, {0,0}});
+  vector<double> av = ale::getAngularVelocity(rots, times, 0.5, ale::linear);
+
+  Eigen::Quaterniond quat(0, 1, 1, 0);
+  quat = quat.normalized();
+
+  Eigen::Quaterniond dQuat(0, -1, 1, 0);
+  dQuat = dQuat.normalized();
+
+  Eigen::Quaterniond avQuat = quat.conjugate() * dQuat;
```
kberryUSGS: This test looks like it performs the steps in `getAngularVelocity` without calling the function. Why?

jessemapel: I'll redo the math by hand so that we have a better comparison. It's still going to be a bit messy because of normalization.

PR #64, opened 2018-12-13
First draft of Python interface

Kelvinrr:
This should connect the python and C++ side of ale. Doesn't have a lot of debug as we probably should figure that out in a different PR. From what I understand the best way is to change how the Python side reports errors.

### DIFF
--- a/notebooks/cpp_load.ipynb
+++ b/notebooks/cpp_load.ipynb
@@ -1,0 +1,159 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#pragma cling add_include_path(\"/Users/krodriguez-pr/repos/ale/include/\")\n",
+    "#pragma cling add_include_path(\"/Users/krodriguez-pr/repos/ale/include/json\")\n",
+    "#pragma cling add_include_path(\"/Users/krodriguez-pr/anaconda3/envs/ale/include/python3.7m\")\n",
+    "\n",
+    "#pragma cling load(\"/Users/krodriguez-pr/repos/ale/build/libale.dylib\")"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#include \"Python.h\"\n",
+    "#include \"ale.h\"\n",
+    "\n",
+    "std::string test_lro_label =\n",
+    "        \"PDS_VERSION_ID                     = PDS3\"\n",
+    "        \"\"\n",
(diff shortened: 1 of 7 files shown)


Kelvinrr:
Some insight on how to best test this would be great. Not sure if there is a clean way to mock out the python function call so we can test the rest.

src/ale.cpp
```diff
@@ -266,4 +254,115 @@ namespace ale {
...
+        PyObject *pyth_val;
+        pyth_val = PyObject_CallFunctionObjArgs(pyth_func, ptype, pvalue, ptraceback, NULL);
+
+        pystr = PyObject_Str(pyth_val);
+        str = PyBytes_AS_STRING(PyUnicode_AsUTF8String(pystr));
+        full_backtrace = strdup(str);
+        Py_DECREF(pyth_val);
+
+        std::string join_cmd = "trace = ''.join(list(" + std::string(full_backtrace) + "))";
+        PyRun_SimpleString(join_cmd.c_str());
```
Kelvinrr: stacktrace is returned as a list, for pretty printing, I am using `PyRun_SimpleString` to join the list as a string. Otherwise you get a `str(list())` type string format.

src/ale.cpp
```diff
@@ -266,4 +254,115 @@ namespace ale {
    return result;
  }
 
+ std::string getPyTraceback() {
```
Kelvinrr: This function essentially uses Python's built in traceback library to get a full stack trace of a py error.

src/ale.cpp
```diff
@@ -266,4 +254,115 @@ namespace ale {
...
+    return "";
+ }
+
+ std::string load(std::string filename) {
+     static bool first_run = true;
+     if(first_run) {
+         // Initialize the Python interpreter but only once.
+         first_run = !first_run;
+         Py_Initialize();
+         atexit(Py_Finalize);
```
Kelvinrr: `Py_Initialize` and `Py_Finalize` should only be called once in a programs lifetime, this should ensure that this is enforced.

ale/drivers/base.py
```diff
@@ -19,7 +19,7 @@ class Base(ABC):
 
     """
     def __init__(self, label, *args, **kwargs):
-        self.label = pvl.loads(label)
+        self._label = pvl.loads(label, strict=False)
```
jlaura: Tight couple to PVL is probably fine for now. This will be another place to have multi-format support (for example to read an ALE derived JSON header).

notebooks/cpp_load.ipynb
```diff
@@ -0,0 +1,271 @@
+{
```
jlaura: So I got to go and look up what #pragma cling is doing - very cool. thanks!

jlaura:
This looks good to me. @Kelvinrr If you are happy with this feel free to merge.

PR #70, opened 2018-12-17
Refactoring how Drivers work

Kelvinrr:
Hopefully this makes it easier to add new drivers mixing sources (PDS3 VS PDS4 VS Spice VS ISIS Label etc.)


ale/drivers/base.py
```diff
@@ -120,12 +177,13 @@ def _compute_ephemerides(self):
...
 
     @property
-    @abstractmethod
-    def metakernel(self):
-        pass
+    def line_exposure_duration(self):
+        try:
+            return self._label['LINE_EXPOSURE_DURATION'].value * 0.001  # Scale to seconds
+        except:
+            return np.nan
```
acpaquette: Why is this in a try except?

ale/drivers/base.py
```diff
@@ -165,8 +223,19 @@ def starting_ephemeris_time(self):
         return self._starting_ephemeris_time
 
     @property
-    def _exposure_duration(self):
-        return self._label['EXPOSURE_DURATION'].value * 0.001  # Scale to seconds
+    def exposure_duration(self):
+        try:
+            return self._label['EXPOSURE_DURATION'].value * 0.001  # Scale to seconds
+        except:
+            return np.nan
```
jlaura: Stick a warning in here so that we do not silently return a NaN?

ale/drivers/base.py
```diff
@@ -94,6 +97,60 @@ def to_pfeffer_response(self):
...
+
+
+class Framer(Base):
+    @property
+    def name_model(self):
+        return "USGS_ASTRO_FRAME_SENSOR_MODEL"
+
+
+class PDS3():
     def _compute_ephemerides(self):
```
jlaura: designed as a mixin?

ale/drivers/base.py
```diff
@@ -195,13 +264,51 @@ def detector_center(self):
...
+    def starting_detector_line(self):
+        return 1
+
+    @property
+    def starting_detector_sample(self):
+        return 1
+
+    @property
+    def detector_sample_summing(self):
+        return 1
```
jlaura: Should we be parsing something here and then failing to 1 (with a warning?)

ale/drivers/base.py
```diff
@@ -195,13 +264,51 @@ def detector_center(self):
...
+        return spice.gdpool('INS{}_OD_K'.format(self.ikid),0, 3)
+
     @property
     def ikid(self):
         return spice.bods2c(self.instrument_id)
 
     @property
     def fikid(self):
-        fn = self._label.get('FILTER_NUMBER', 0)
+        fn = self.filter_number
```
jlaura: I was never enamored with this. Felt hacky, so please alter the whole fikid if that makes sense?

ale/drivers/base.py
```diff
@@ -339,50 +420,39 @@ def center_ephemeris_time(self):
...
-    @property
-    def t0_quaternion(self):
-        return self.starting_ephemeris_time - self.center_ephemeris_time
 
+class Isis3():
     @property
-    def dt_ephemeris(self):
-        return (self.ending_ephemeris_time - self.starting_ephemeris_time) / self.number_of_ephemerides
+    def start_time(self):
+        return self._label['IsisCube']['Instrument']['StartTime']
```
jlaura: Does it make sense to do this or to recursively just search for 'StartTime' in the whole label? I think `plio` has the `find_in_dict` func.

ale/drivers/base.py
```diff
@@ -339,50 +420,39 @@ def center_ephemeris_time(self):
...
     @property
     def _exposure_duration(self):
-        return self._label['LINE_EXPOSURE_DURATION'].value * 0.001  # Scale to seconds
-
-class Framer(Base):
-
-    @property
-    def name_model(self):
-        return "USGS_ASTRO_FRAME_SENSOR_MODEL"
+        return self._label['IsisCube']['Instrument']['ExposureDuration'].value * 0.001 # Scale to seconds
```
jlaura: I wonder if we should have a `scale` attribute to pull the magic number from here? We would need a scale for each attribute that we cared about.

ale/drivers/lro_driver.py
```diff
@@ -6,17 +6,22 @@
 import spiceypy as spice
 
 from ale.util import get_metakernels
-from ale.drivers.base import LineScanner
-from ale.drivers.distortion import RadialDistortion
+from ale.drivers.base import LineScanner, RadialDistortion, Spice, PDS3, Isis3
 
-class LRO_LROC(LineScanner, RadialDistortion):
+class LrocSpice(Spice, LineScanner, RadialDistortion):
```
jlaura: I like this. Not sure what others think, but having the declaration look like this, I know exactly where the data are from and what the composite parts are that make up this particular driver.

ale/drivers/base.py
Kelvinrr: yeah, by itself meaningless but can be used for any instrument driver using PDS3 Labels.

ale/drivers/base.py
Kelvinrr: That was the behavior before refactoring, I imagine any problem with the implementation of the methods themselves can be done in a different PR. We still need to figure out the when where and how of warnings and errors.

ale/drivers/base.py
Kelvinrr: Some labels do not have LINE_EXPOSURE_DURATION. We definitely need a better way to handle this case.

jessemapel: Different instruments also use different units on this :/

ale/drivers/base.py
jessemapel: We need to decide on how to handle these kinds of failures. I think a warning and returning np.NaN works

ale/drivers/base.py
jessemapel: Summing is largely a mission by missions value. There is a lot of variation in what key it is stored in and what different values mean.

I think defaulting to 1 in the base class is a good choice.

ale/drivers/base.py
```diff
@@ -195,13 +264,51 @@ def detector_center(self):
...
+    def detector_sample_summing(self):
+        return 1
+
+    @property
+    def detector_line_summing(self):
+        return self._label.get('SAMPLING_FACTOR', 1)
+
+    @property
+    def filter_number(self):
+        return self._label.get('FILTER_NUMBER', 0)
```
jessemapel: Filter Number is not something used by every instrument. Generally, filters are used by framing camera instruments.

ale/drivers/base.py
jessemapel: This is one of the keys that is explicitly required by ISIS to be at this location in the label. I think this is fine.

ale/drivers/base.py
jessemapel: I think this is something that we can work on as we add more supported instruments. Right now, I don't see enough need for this.

ale/drivers/lro_driver.py
kberryUSGS: For what it's worth, I like this, too.

ale/drivers/keys.py
```diff
@@ -0,0 +1,64 @@
+base = {
```
Kelvinrr: Someone should double check these are grouped correctly.

ale/drivers/base.py
acpaquette: Do we know if this is the case for all of the ISIS cube specific labels?

ale/drivers/base.py
```diff
@@ -1,25 +1,20 @@
...
+class Driver():
     """
-    Abstract base class for all PDS label parsing. Implementations should override
-    properties where a kernel provider deviates from the most broadly adopted
-    approach.
-
-    Methods that must be provided:
-    - instrument_id
-    - metakernel
+    Abstract base class for all Drivers.
```
jessemapel: Base class for all Drivers, it's not an ABC any more

ale/drivers/keys.py
```diff
@@ -0,0 +1,64 @@
...
+    'spacecraft_name',
+    'start_time',
+    'starting_detector_line',
+    'starting_detector_sample',
+    'starting_ephemeris_time',
+    'sun_position',
+    'sun_velocity',
+    'target_name',
+    'number_of_ephemerides',
+    'fikid',
```
jessemapel: move to filter list

ale/drivers/keys.py
```diff
@@ -0,0 +1,64 @@
...
+}
+
+linescanner = {
+    'line_exposure_duration',
+    'line_scan_rate',
+    't0_ephemeris',
+    't0_quaternion',
+    'dt_ephemeris',
+    'dt_quaternion',
+}
```
jessemapel: It's weird if there is a duplication of keys between lists, but should not cause an issue.

If implementation classes implement the same key, then there can be conflicts based on order of inheritance in base classes.

jlaura:
@Kelvinrr This looks good to go to me? If so, feel free to merge.

PR #72, opened 2018-12-19
GetRotation Implementation 

### LINKED ISSUE #34: Implement get rotation from function coefficients

acpaquette:
Adds implementation and tests for the function based getRotation function. Closes #34

### DIFF
--- a/src/ale.cpp
+++ b/src/ale.cpp
@@ -174,14 +174,25 @@ namespace ale {
   }
 
   // Rotation Function Functions
-  vector<double> getRotation(string from, string to,
-                             vector<double> coefficients, double time) {
-    vector<double> coordinate = {0.0, 0.0, 0.0};
-    return coordinate;
+  std::vector<double> getRotation(vector<vector<double>> coeffs, double time) {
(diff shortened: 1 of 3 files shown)


src/ale.cpp
```diff
@@ -174,14 +174,25 @@ namespace ale {
...
-  vector<double> getRotation(string from, string to,
-                             vector<double> coefficients, double time) {
-    vector<double> coordinate = {0.0, 0.0, 0.0};
-    return coordinate;
+  std::vector<double> getRotation(vector<vector<double>> coeffs, double time) {
+    vector<double> rotation = {0.0, 0.0, 0.0};
+
+    rotation[0] = evaluatePolynomial(coeffs[0], time, 0); // X
+    rotation[1] = evaluatePolynomial(coeffs[1], time, 0); // Y
+    rotation[2] = evaluatePolynomial(coeffs[2], time, 0); // Z
```
jessemapel: We need to check that rotation has 3 elements before making these calls. See the checks in the position version

jessemapel: Also need a test to make sure we're handling that case properly

PR #73, opened 2018-12-19
Adds spiceinit'd cube driver

jessemapel:
The output dictionary format is different, but we need to decide on a format before it's worth cleaning up.


environment.yml
```diff
@@ -11,5 +11,7 @@ dependencies:
   - eigen
   - spiceypy
   - numpy
+  - numba
```
Kelvinrr: what is numba being used for?

ale/drivers/cube_driver.py
```diff
@@ -0,0 +1,224 @@
...
+class Cube(Base):
+
+    def __init__(self, file, *args, **kwargs):
+        super(Cube, self).__init__('')
+        self.label = pvl.load(file)
+        for table in self.label.getlist('Table'):
+            binary_data = read_table_data(table, file)
+            field_data = parse_table_data(table, binary_data)
+            if table['Name'] == 'InstrumentPointing':
+                self.inst_pointing_table = parse_rotation_table(table, field_data)
```
Kelvinrr: If these tables vars are only used internally, probably best to prefix with an underscore like in other drivers.

ale/drivers/base.py
```diff
@@ -1,13 +1,11 @@
...
-
 from dateutil import parser
 import numpy as np
 import pvl
 import spiceypy as spice
 
 from ale.drivers import distortion
 
-class Base(ABC):
+class Base:
```
jlaura: This should or should not be an abstract base? @Kelvinrr?

tests/pytests/test_cube_drivers.py
```diff
@@ -0,0 +1,440 @@
...
+                                            3.57484262662076e-04, 0.0,
+                                            1.00401047146886e-05, 0.0,
+                                            1.00401047146886e-05)
+End_Object
+End
+"""
+
+    def test_table_data(table_label, file):
+        count = table_label['Records'] * len(table_label.getlist('Field'))
+        doubles = [i for i in range(count)]
```
Kelvinrr: Probably better to simply do `list(range(count))`

ale/drivers/base.py
```diff
@@ -36,7 +34,8 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         this is done, the object is out of scope and the
         kernels can be unloaded.
         """
-        spice.unload(self.metakernel)
+        if self.metakernel:
```
jlaura: Should we warn in the instance where `self.metakernel` is None?

ale/drivers/cube_driver.py
```diff
@@ -0,0 +1,224 @@
...
+
+import pvl
+import spiceypy as spice
+import numpy as np
+import quaternion
+
+from ale import config
+from ale.drivers.base import Base
+
+def read_table_data(table_label, file):
```
jlaura: We should open an issue to get numpy style docstrings onto here.

ale/drivers/cube_driver.py
```diff
@@ -0,0 +1,224 @@
...
+def field_format(field_label):
+    data_formats = {
+        'Integer' : 'i',
+        'Double'  : 'd',
+        'Real'    : 'f'
+    }
+    return data_formats[field_label['Type']] * field_label['Size']
+
+def parse_field(field_label, data):
+    if field_label['Type'] == 'Text':
```
jlaura: pass the encoding as a kwarg?

ale/drivers/cube_driver.py
```diff
@@ -0,0 +1,224 @@
...
+    if all (key in field_data for key in ('J2000SVX','J2000SVY','J2000SVZ')):
+        results['PositionCoefficients'] = np.array( [field_data['J2000SVX'][:-1],field_data['J2000SVY'][:-1],field_data['J2000SVZ'][:-1]] )
+        results['BaseTime'] = field_data['J2000SVX'][-1]
+        results['TimeScale'] = field_data['J2000SVY'][-1]
+    return results
+
+class Cube(Base):
+
+    def __init__(self, file, *args, **kwargs):
+        super(Cube, self).__init__('')
```
jlaura: I do not think you need the super here, in that you do not want to call the Base classes's init. You are overriding that here.

ale/drivers/cube_driver.py
```diff
@@ -0,0 +1,224 @@
...
+        results['ConstantRotation'] = quaternion.from_rotation_matrix(np.reshape(const_rotation_mat, (3, 3)))
+        results['ConstantFrames'] = np.array(label['ConstantFrames'])
+    if all (key in label for key in ('PoleRa','PoleDec','PrimeMeridian')):
+        results['BodyRotationCoefficients'] = np.array( [label['PoleRa'],label['PoleDec'],label['PrimeMeridian']] )
+    if all (key in label for key in ('PoleRaNutPrec','PoleDecNutPrec','PmNutPrec','SysNutPrec0','SysNutPrec1')):
+        results['SatelliteNutationPrecessionCoefficients'] = np.array( [label['PoleRaNutPrec'],label['PoleDecNutPrec'],label['PmNutPrec']] )
+        results['PlanetNutationPrecessionAngleCoefficients'] = np.array( [label['SysNutPrec0'],label['SysNutPrec1']] )
+    return results
+
+def parse_position_table(field_data):
```
Kelvinrr: Probably worth testing the outputs of these functions to make sure they are returning correct values in this PR?

ale/drivers/cube_driver.py
```diff
@@ -0,0 +1,224 @@
...
+    if 'TimeDependentFrames' in label:
+        results['TimeDependentFrames'] = np.array(label['TimeDependentFrames'])
+    if all (key in label for key in ('ConstantRotation','ConstantFrames')):
+        const_rotation_mat = np.array(label['ConstantRotation'])
+        results['ConstantRotation'] = quaternion.from_rotation_matrix(np.reshape(const_rotation_mat, (3, 3)))
+        results['ConstantFrames'] = np.array(label['ConstantFrames'])
+    if all (key in label for key in ('PoleRa','PoleDec','PrimeMeridian')):
+        results['BodyRotationCoefficients'] = np.array( [label['PoleRa'],label['PoleDec'],label['PrimeMeridian']] )
+    if all (key in label for key in ('PoleRaNutPrec','PoleDecNutPrec','PmNutPrec','SysNutPrec0','SysNutPrec1')):
+        results['SatelliteNutationPrecessionCoefficients'] = np.array( [label['PoleRaNutPrec'],label['PoleDecNutPrec'],label['PmNutPrec']] )
```
Kelvinrr: lol Im just gonna trust this code is correct.

environment.yml
jessemapel: quaternion uses numba for acceleration. If you don't have it, it complains on import.

ale/drivers/cube_driver.py
```diff
@@ -0,0 +1,224 @@
...
+    def semimajor(self):
+        return self.body_radii[0]
+
+    @property
+    def semiminor(self):
+        return self.body_radii[2]
+
+    @property
+    def reference_frame(self):
+        return self.body_orientation_table['TimeDependentFrames'][0]
```
jlaura: What about changing these to `self.appropriate_table.get('KEY', empty_default)`? I guess my real question is about error handling. The funcs above pack dicts `if all()` without any potential that `all() == False`. That is fine I think, but then this code needs to handle those instances. If only because things like the `number_of_ephemerides` properties will end up throwing a KeyError (which does not make sense unless you have seen this code and followed the execution path).

ale/drivers/mdis_driver.py
```diff
@@ -127,10 +128,35 @@ def _exposure_duration(self):
...
-        
+
     @property
     def starting_ephemeris_time(self):
         if not hasattr(self, '_starting_ephemeris_time'):
             sclock = self._label['IsisCube']['Archive']['SpacecraftClockStartCount']
             self._starting_ephemeris_time = spice.scs2e(self.spacecraft_id, sclock)
         return self._starting_ephemeris_time
+
+class isisSpicedMessenger(Cube, TransverseDistortion):
```
jlaura: Pedantic - class names should be capitalized from the start.

Kudos to all that have worked on this driver structure - this is very clean looking.

ale/drivers/mdis_driver.py
```diff
@@ -127,10 +128,35 @@ def _exposure_duration(self):
...
+        """
+        """
+        coeffs = self.label['NaifKeywords']['INS{}_FL_TEMP_COEFFS'.format(self.ikid)]
+
+        # reverse coeffs, mdis coeffs are listed a_0, a_1, a_2 ... a_n where
+        # numpy wants them a_n, a_n-1, a_n-2 ... a_0
+        f_t = np.poly1d(coeffs[::-1])
+
+        # eval at the focal_plane_tempature
+        return f_t(label['IsisCube']['Instrument']['FocalPlaneTemperature'].value)
```
jlaura: Since the calls to do the interpolation are identical in this driver, a pds driver, etc., do we pull the logic for applying the interpolation function into the module scope and then call it in the specific driver?

environment.yml
jlaura: I know this PR didn't add it, but eigen too I think.

ale/drivers/base.py
Kelvinrr: Technically no, we moved away from using ABCs to enforce keys and instead used the other approach of using a required key set.

ale/drivers/base.py
Kelvinrr: In my PR, this is moved to the Spice specific mixin. As there is no case where you use the Spice mixin and not expect to load kernels, I imagine if no metakernel is found, the property should simply error.

jlaura: :+1: That sounds totally reasonable.

notebooks/read_cube.ipynb
```diff
@@ -406,8026 +805,35 @@
...
-       "   -0.0406503006491793,\n",
-       "   -0.04074340182162566,\n",
-       "   -0.04083590103936866,\n",
-       "   -0.040928700113815566,\n",
-       "   -0.041021601072140215,\n",
-       "   -0.04111440054568199,\n",
-       "   -0.041208100159118376,\n",
-       "   -0.04129979966814669,\n",
-       "   -0.04139300041015805,\n",
-       "   -0.04148590064813
```
SgStapleton: Is there a way to clear these outputs before committing? ...Just takes up so much space...

ale/drivers/cube_driver.py
```diff
@@ -0,0 +1,224 @@
...
+    results = {field['Name']:[] for field in fields}
+    offset = 0
+    for record in range(table_label['Records']):
+        for field in fields:
+            field_data = parse_field(field, data[offset:])
+            results[field['Name']].append(field_data)
+            offset += field_size(field)
+    return results
+
+def parse_rotation_table(label, field_data):
```
Kelvinrr: Probably worth putting some docstrings for what each if statement is actually handling. Can't tell from looking what the difference in the different J2000 strings are.

jessemapel:
This is now ready for re-review.

I updated it based on the driver refactor. I also improved the tests and added numpy style doc strings.

ale/drivers/base.py
```diff
@@ -509,13 +509,13 @@ class Isis3():
...
             if isinstance(self._file, pvl.PVLModule):
-                self._label = label
+                self._label = self._file
             try:
                 self._label = pvl.loads(self._file, strict=False)
             except AttributeError:
                 self._label = pvl.load(self._file, strict=False)
             except:
-                raise Exception("{} is not a valid label".format(label))
+                raise Exception("{} is not a valid label".format(self._file))
```
jlaura: This should probably have an associated error. Options might be `SyntaxError`, `ValueError`, or a custom error type. Also, why the try/except? In this case, it might be better to do an if/else for string vs. handle because then the PVL module's error should propagate up directly.

ale/drivers/isis_spice_driver.py
```diff
@@ -0,0 +1,588 @@
...
+        The ISIS table label
+    file : file
+        The ISIS cube file
+
+    Returns
+    -------
+    bytes
+      The binary portion of the table data
+    """
+    file.seek(label['StartByte']-1) # This -1 is straight out of ISIS
```
jlaura: file is reserved, maybe input or someone other something.  I think that the -1 is likely required because the byte offsets at 1 off from the actual read start. I had to do an identical thing on another PDS style label.

ale/drivers/isis_spice_driver.py
```diff
@@ -0,0 +1,588 @@
...
+from ale.drivers.base import Driver, Isis3
+
+def read_table_data(table_label, file):
+    """
+    Helper function to read all of the binary table data
+
+    Parameters
+    ----------
+    table_label : PVLModule
+        The ISIS table label
```
jlaura: I suspect that sphinx will complain that the `The ISIS...` is not aligned with `PVLModule`, but I would be very happy if it did not.

ale/drivers/isis_spice_driver.py
```diff
@@ -0,0 +1,588 @@
...
+    single entry in a table field
+
+    Parameters
+    ----------
+    field_label : PVLModule
+        The field label
+
+    Returns
+    -------
+    str
```
jlaura: `str : ` - sphinx again...

ale/drivers/isis_spice_driver.py
```diff
@@ -0,0 +1,588 @@
...
+    data : bytes
+        The binary data for the entire table
+
+    Returns
+    -------
+    list
+      The table field entries
+    """
+    if field_label['Type'] == 'Text':
+        results[field_label['Name']].append(data[:field_label['Size']].decode(encoding=encoding))
```
jlaura: what is `results`?

ale/drivers/isis_spice_driver.py
```diff
@@ -0,0 +1,588 @@
...
+    ----------
+    table_label : PVLModule
+        The table label
+
+    data : bytes
+        The binary data for the entire table
+
+    Returns
+    -------
+    dict
```
jlaura: `dict : ` - maybe Sphinx is better at handling the missing `:`?

ale/drivers/isis_spice_driver.py
```diff
@@ -0,0 +1,588 @@
...
+      The table as a dict
+    """
+    fields = table_label.getlist('Field')
+    results = {field['Name']:[] for field in fields}
```
jlaura: Are these keys guaranteed to be unique?

ale/drivers/isis_spice_driver.py
```diff
@@ -0,0 +1,588 @@
...
+      The rotation data
+    """
+    results = {}
+    if all (key in field_data for key in ['J2000Q0','J2000Q1','J2000Q2','J2000Q3']):
+        results['Rotations'] = quaternion.as_quat_array( [ [q0, q1, q2, q3] for q0, q1, q2, q3 in zip(field_data['J2000Q0'],field_data['J2000Q1'],field_data['J2000Q2'],field_data['J2000Q3']) ] )
+    if all (key in field_data for key in ('AV1','AV2','AV3')):
+        results['AngularVelocities'] = np.array( [ [av1, av2, av3] for av1, av2, av3 in zip(field_data['AV1'],field_data['AV2'],field_data['AV3']) ] )
+    if 'ET' in field_data:
+        results['Times'] = np.array(field_data['ET'])
+    if all (key in field_data for key in ('J2000Ang1','J2000Ang2','J2000Ang3')):
```
jlaura: `in tuple` while `in list` above. doesn't matter, just style.

ale/drivers/isis_spice_driver.py
```diff
@@ -0,0 +1,588 @@
...
+      The table as a dict
+    """
+    fields = table_label.getlist('Field')
+    results = {field['Name']:[] for field in fields}
+    offset = 0
+    for record in range(table_label['Records']):
+        for field in fields:
+            field_data = parse_field(field, data[offset:])
+            results[field['Name']].append(field_data)
+            offset += field_size(field)
```
jlaura: Are the keys in each field also guaranteed to be identical? In other words, does parse_field handle the read error or does this func?

ale/drivers/isis_spice_driver.py
```diff
@@ -0,0 +1,588 @@
...
+    if all (key in label for key in ('ConstantRotation','ConstantFrames')):
+        const_rotation_mat = np.array(label['ConstantRotation'])
+        results['ConstantRotation'] = quaternion.from_rotation_matrix(np.reshape(const_rotation_mat, (3, 3)))
+        results['ConstantFrames'] = np.array(label['ConstantFrames'])
+    if all (key in label for key in ('PoleRa','PoleDec','PrimeMeridian')):
+        results['BodyRotationCoefficients'] = np.array( [label['PoleRa'],label['PoleDec'],label['PrimeMeridian']] )
+    if all (key in label for key in ('PoleRaNutPrec','PoleDecNutPrec','PmNutPrec','SysNutPrec0','SysNutPrec1')):
+        results['SatelliteNutationPrecessionCoefficients'] = np.array( [label['PoleRaNutPrec'],label['PoleDecNutPrec'],label['PmNutPrec']] )
+        results['PlanetNutationPrecessionAngleCoefficients'] = np.array( [label['SysNutPrec0'],label['SysNutPrec1']] )
+    return results
```
jlaura: @Kelvinrr Any thoughts on how this might be cleaned up some so it is not a block of if/if/if/if searching for keys. Feels like something cleaner exists, but it is escaping me.

ale/drivers/isis_spice_driver.py
jessemapel: Yes, Table field labels always contain one of each:
* Name
* Type
* Size

jessemapel: Also, the Table is invalid if two fields share the same name

ale/drivers/isis_spice_driver.py
jessemapel: What read error are you concerned about?

jlaura: Is it possible that a field mismatch exists? In other words, it is identified in the fields variable, but then not in the binary later?

ale/drivers/isis_spice_driver.py
Kelvinrr: I can't imagine anything much better, there may exists a compelling argument for taking advantage of the existing driver structure to use the different table types as mixins, but that's really just gonna shuffle the logic around. I feel in this case if blocks are fine, maybe just make the clauses easier to read (e.g. `{"key1", "key2"} <= label.keys()` assuming whatever `label` is formatted like affords it, otherwise just your above suggestion would be fine)

ale/drivers/isis_spice_driver.py
```diff
@@ -0,0 +1,588 @@
...
+        results['Times'] = np.array(field_data['ET'])
+    if all (key in field_data for key in ('J2000XV','J2000YV','J2000ZV')):
+        results['Velocities'] = np.array( [ [x, y, z] for x, y, z in zip(field_data['J2000XV'],field_data['J2000YV'],field_data['J2000ZV']) ] )
+    if all (key in field_data for key in ('J2000SVX','J2000SVY','J2000SVZ')):
+        results['PositionCoefficients'] = np.array( [field_data['J2000SVX'][:-1],field_data['J2000SVY'][:-1],field_data['J2000SVZ'][:-1]] )
+        results['BaseTime'] = field_data['J2000SVX'][-1]
+        results['TimeScale'] = field_data['J2000SVY'][-1]
+    return results
+
+class IsisSpice(Isis3):
```
Kelvinrr: Probably worth to have class level doc string explaining what the member variables are.

Kelvinrr:
Probably worth stripping logic from the notebook to use ALE library calls.

ale/drivers/isis_spice_driver.py
Kelvinrr: From talking with @scsides, the only way that could be possible is if you go around manually editing the cubes deleting binary data. Probably not something we want to worry about for now.

notebooks/read_cube.ipynb
Kelvinrr: Yeah, in the future we need to keep limited output in the cubes, in this case the notebook only existed for the sake of testing and is gone now.

Kelvinrr:
The left over comments should be easy enough to take care of, after that we should be able to merge this.

jessemapel:
@jlaura @Kelvinrr @SgStapleton Is this good to merge now? I think I've addressed most of the comments

PR #75, opened 2018-12-21
Conda build set-up

SgStapleton:
This begins to address https://github.com/USGS-Astrogeology/ale/issues/4

### DIFF
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -1,0 +1,4 @@
+mkdir build
+cd build
+cmake ..
+make
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -1,4 +1,4 @@
 mkdir build
 cd build
-cmake -DCMAKE_INSTALL_PREFIX=$PREFIX ..
+cmake -DCMAKE_INSTALL_PREFIX=$PREFIX -DCMAKE_BUILD_TYPE=Release ..
 make install
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -1,4 +1,4 @@
 mkdir build
 cd build
-cmake ..
-make
+cmake -DCMAKE_INSTALL_PREFIX=$PREFIX ..
+make install
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -16,7 +16,7 @@ build:
 # Shotgun strat on requirements until we can narrow them down
 requirements:
   build:
-  - python=3.7.1
+  - python
   - cmake>=3.10
   - gsl
   - openblas>=0.3.0
@@ -26,7 +26,7 @@ requirements:
   - python-dateutil
   - pvl
   run:
-  - python=3.7.1
+  - python
   - cmake>=3.10
   - gsl
   - openblas>=0.3.0
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,0 +1,50 @@
+{% set version = "0.0" %}
+{% set git_branch = "master" %}
+{% set build_number = "0" %}
+
+package:
+  name: ale
+  version: {{ version }}
+
+source:
+  git_url: 'https://github.com/USGS-Astrogeology/ale.git'
+  git_branch: {{ git_branch }}
+
+build:
+  number: {{ build_number }}
+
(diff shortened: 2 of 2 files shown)


SgStapleton:
The ``conda build``` command appears to run successfully, but I have not tested pushing this to anaconda, or (obviously) installing it from the usgs-astrogeology channel.

recipe/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+  git_url: 'https://github.com/USGS-Astrogeology/ale.git'
+  git_branch: {{ git_branch }}
+
+build:
+  number: {{ build_number }}
+
+# Shotgun strat on requirements until we can narrow them down
+requirements:
+  build:
+  - python=3.7.1
```
jlaura: Can we unpin the python version?

SgStapleton:
Waiting for permission to push new packages to anaconda for testing purposes.

recipe/meta.yaml
SgStapleton: It appears to be building properly without this pinned. Will confirm when I can test the uploading and installing process.

recipe/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+  version: {{ version }}
+
+source:
+  git_url: 'https://github.com/USGS-Astrogeology/ale.git'
+  git_branch: {{ git_branch }}
+
+build:
+  number: {{ build_number }}
+
+# Shotgun strat on requirements until we can narrow them down
```
kberryUSGS: I think comment this is from the ISIS3 one and not applicable here?

recipe/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+  - python-dateutil
+  - pvl
+
+# Add the tests eventually
+# test:
+#
+
+# about:
+#   home: https://developers.google.com/protocol-buffers/
+#   license: BSD 3-Clause
```
kberryUSGS: What license do we want to use for ale?

SgStapleton: Yeah, I left an example "about" section commented out for reference, but forgot to ask about specific ale implementation... Any thoughts?

kberryUSGS: Yeah! This is less a comment on your actual PR and more a question that should have gone in a ticket. I'm not sure.

recipe/build.sh
```diff
@@ -0,0 +1,4 @@
+mkdir build
+cd build
+cmake -DCMAKE_INSTALL_PREFIX=$PREFIX ..
```
jessemapel: We should add a `-DCMAKE_BUILD_TYPE=Release` here

recipe/meta.yaml
jessemapel: I think we can't use a license? My understandings is that we have to push everything we do into the public domain. @jlaura has looked into this more.

recipe/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+#   license_file: LICENSE
+#   summary: "Protocol Buffers - Google's data interchange format."
```
jessemapel: Probably should change this summary

recipe/meta.yaml
kberryUSGS: @jessemapel Isn't pushing something into the public domain still the sort of thing that requires a license to indicate that it is all in the public domain?

recipe/meta.yaml
```diff
@@ -13,37 +13,54 @@ source:
...
   number: {{ build_number }}
 
+# A number of the requirements below will have two instances: one will include a specific version
+# pinning, while the other one will not. When building a version of ALE for use specifically within
+# ISIS, use the instances with the version pinned. For all other builds, use the unpinned instances.
+# (Confirm that every requirement has one and only one instance uncommented.)
+#
+# ISIS requires pinned versions for these dependencies to avoid conflicts with other dependencies.
+# Make sure to always confirm that version numbers in this file and version numbers in the
+# meta.yaml file within the ISIS3 repo match for each edpendency before building this library
```
kberryUSGS: typo checking: edependency -> dependency

kberryUSGS:
@SgStapleton Is this ready to be merged?

PR #82, opened 2019-03-15
Getting ALE in sync with USGSCSM

Kelvinrr:
Ale now pumps out the nested JSON objects that CSM wants, also had those changes cascade in such a way to not make the drivers less inconsistent. I'm getting a successful model from CSM using spice inited cubes.


ale/drivers/base.py
```diff
@@ -1,7 +1,211 @@
...
+    cube : file
+           The ISIS cube file
+
+    Returns
+    -------
+    bytes :
+        The binary portion of the table data
+    """
+    cubehandle = open(cube, "rb")
+    cubehandle.seek(table_label['StartByte']) # This -1 is straight out of ISIS
```
jessemapel: Remove the comment since the -1 is not longer needed.

ale/drivers/base.py
```diff
@@ -57,10 +262,111 @@ def to_dict(self):
...
+            "semiminor" : self._semiminor,
+            "unit" : "m" # default to meters
+        }
 
-class LineScanner(Driver):
+    @property
+    def reference_height(self):
+        # TODO: This should be a reasonable #
+        return {
+            "maxheight" : 0,
```
jessemapel: minheight?

ale/drivers/base.py
```diff
@@ -103,15 +409,28 @@ def line_scan_rate(self):
     @property
     def number_of_ephemerides(self):
         #TODO: Not make this hardcoded
-        return 909
+        return 909 #len(self._sensor_position)
```
jessemapel: Is there a reason why this is returning the hardcoded value with the real value commented out?

ale/drivers/base.py
```diff
@@ -103,15 +409,28 @@ def line_scan_rate(self):
...
 
     @property
     def number_of_quaternions(self):
         #TODO: Not make this hardcoded
-        return 909
+        return 909 #
+        len(self._quaternions)
```
jessemapel: Same as number_of_ephemerides. Why is the real value commented out?

ale/drivers/base.py
```diff
@@ -439,21 +697,42 @@ def sun_velocity(self):
...
+    def _sensor_position(self):
+        if not hasattr(self, '_position'):
+            eph = []
+            current_et = self.starting_ephemeris_time
+            for i in range(self.number_of_ephemerides):
+                state, _ = spice.spkezr(self.spacecraft_name,
+                                        current_et,
+                                        self.reference_frame,
+                                        'NONE',
+                                        self.target_name,) # If this is the sensor, insufficient, if this is the spacecraft, it works? Huh?
```
jessemapel: The SPKs store the position of the whole spacecraft, not the individual sensors. So, spacecraft instead of sensor is correct.

ale/drivers/base.py
```diff
@@ -439,21 +697,42 @@ def sun_velocity(self):
...
+                eph_rates.append(state[3:])
+                current_et += getattr(self, "dt_ephemeris", 0)
+            # By default, spice works in km
+            self._velocity = [e*1000 for e  in eph_rates]
+        return self._velocity
```
jessemapel: Would it be worth saving both the position and velocity in these methods? spkezr will return both of these.

Same with _sensor_position above.

ale/drivers/base.py
Kelvinrr: Right now the way things are set up, returning the real value ends with a stack overflow. self._sensor_positions - calls -> self.number_of_ephemerides - calls -> self._sensor_position and so on. Didn't seem to be an immediately obvious solution so I left it in.

ale/drivers/base.py
Kelvinrr: same reason

ale/drivers/base.py
Kelvinrr: That's not my comment, left over in the compute ephemeris method.

ale/drivers/base.py
Kelvinrr: The previous approach felt clunky and broke form. If more people think otherwise I can revert it.

jessemapel:
I'm good with this being merged, I'll check back tonight and if there's no more comments/etc I'll merge.

ale/drivers/base.py
jlaura: This was mine from a while ago. We can pull the comment.

ale/drivers/base.py
jlaura: What about the constructor taking this as an argument? We can/should write the logic to compute a reasonable number of ephemerides if a user specified number is not provided.

Kelvinrr: definitely better than hardcoded.

jessemapel: The number of ephemerides is going to be based on the length and jitter of the line scan image. A reasonabl-ish default would be 1 per line. That is probably overkill, though, and we can get away with 1 per 5-10 lines.

Kelvinrr: I figure either some function of the line number or (assuming Spice just has a table of values which it interpolates on) the number of positions/quats in the spice table.

PR #84, opened 2019-03-25
refactors focal length epsilon out of base

jlaura:


jlaura:
Closes #85

ale/drivers/base.py
```diff
@@ -582,6 +578,7 @@ def __enter__(self):
         Called when the context is created. This is used
         to get the kernels furnished.
         """
+        print(self.metakernel)
```
jessemapel: Why print here?

jessemapel: Why print here?

PR #86, opened 2019-03-25
Testing

jlaura:
How this works: The `SimpleSpice` class has been extended to have a bit more fixture logic that makes checks for keys for `gdpool` and `bodvrd`. This is acocmplished via a decorator with an assertion. Failures raise nicely through pytest.

Implementation: Each driver (and flavor) that defines unique keys can add them into a `pool_keys` variable within the test. This requires that a dev add the key in two places (in the driver and in the test). I would suggest that this is a positive cognitive load as it hopefully pushes us to (1) look at kernels, (2) write the test, (3) implement. MDIS is the most 'unique' of the currently implemented so check that out for a better view of the extent of this. Thoughts?

Quality of Tests: The MRO test is furthest along on how we might test things. We need to have a rich base test and then we can rely on `pool_key`, `*_creation`, and individual tests for overriding methods.

Concerns:

`*_driver` fixtures are janky. We can literally mislabel the `*.metakernel = get_mockkernels` with the wrong instrument and get a passing test. This is because of how simple `SimpleSpice` really is. If we carried instrument codes this would break, but why write a test for a test. We also probably can't get more generic with `SimpleSpice` and centralize the fixture with making a parametrized fixture that dynamically injects the instrument specific pool keys. This is an option though.

`bodvrd` - I had a moment of concern with the key check on this for MDIS because the key we are checking for is `Earth`. The mission is nominally Mercury facing. This is not a big deal since the validation is just on the key, but it brings up a good point: if we only ever get body information, do we really care to check the key on this at all?

### DIFF
--- a/ale/drivers/mro_driver.py
+++ b/ale/drivers/mro_driver.py
@@ -95,17 +95,3 @@ class CtxPds3Driver(PDS3, CtxSpice):
             'MARS_RECONNAISSANCE_ORBITER': 'MRO'
         }
         return name_lookup[self.label['SPACECRAFT_NAME']]
-
-    @property
-    def instrument_id(self):
-        """
-        Returns an instrument id for unquely identifying the instrument, but often
-        also used to be piped into Spice Kernels to acquire IKIDs. Therefore they
-        the same ID the Spice expects in bods2c calls.
-
-        Returns
-        -------
-        : str
-          instrument id
-        """
-        return self.id_lookup[self.label['INSTRUMENT_NAME']]
--- a/tests/pytests/conftest.py
+++ b/tests/pytests/conftest.py
@@ -1,26 +1,55 @@
 import numpy as np
 
+def check_key(f):
+    def wrapper(*args):
+        obj = args[0]  # This is the class instance; self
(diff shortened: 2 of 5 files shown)


jlaura:
...pass local....fail remote....classy.

```
jlaura@smalls:/data/big/github/ale$ pytest tests/pytests/
========================== test session starts ===========================
platform linux -- Python 3.6.8, pytest-3.5.1, py-1.5.3, pluggy-0.6.0
rootdir: /data/big/github/ale, inifile:
plugins: remotedata-0.2.0, openfiles-0.2.0, doctestplus-0.1.2, cov-2.5.1, arraydiff-0.2
collected 25 items                                                       

tests/pytests/test_isis_spice_drivers.py ....................      [ 80%]
tests/pytests/test_lro_drivers.py .                                [ 84%]
tests/pytests/test_mdis_driver.py .                                [ 88%]
tests/pytests/test_mro_drivers.py ...                              [100%]

======================= 25 passed in 3.19 seconds ========================
```

I'll look into it.

jessemapel:
Based on old architecture. closing prior to re-architecting.

PR #87, opened 2019-03-27
Moving my local changes upstream

Kelvinrr:


ale/drivers/base.py
```diff
@@ -369,109 +368,6 @@ def sun_position(self):
...
-
-    @property
-    def sun_position(self):
-        return {
-            "positions" : self._sun_position,
-            "velocities" : self._sun_velocity,
-            "unit" : "m"
-        }
-
-
```
jlaura: Is all of this supposed to be pulled?

Kelvinrr: I forgot to push some files that had merge conflicts dealt with.

ale/drivers/base.py
```diff
@@ -318,7 +318,7 @@ def radii(self):
         return {
             "semimajor" : self._semimajor,
             "semiminor" : self._semiminor,
-            "unit" : "m" # default to meters
+            "unit" : "km" # default to meters
```
jlaura: ? comment vs. unit

ale/drivers/base.py
```diff
@@ -655,7 +659,7 @@ def _semimajor(self):
           Semimajor axis of the target body
         """
         rad = spice.bodvrd(self.target_name, 'RADII', 3)
-        return rad[1][1]
+        return rad[1][0]
```
jlaura: Are these new changes?

Kelvinrr: yeah, I seem to only get the correct semimajor/semininor this way. The docs for cspice are not clear about what the 3 element array holds, but when using CTX images 0th and 2nd element had the right radii.

jlaura: :+1:

PR #91, opened 2019-04-04
Adding DAWN driver

SgStapleton:
I was able to successfully instantiate a dawn driver in a jupyter notebook with this code, but this has not been tested within the code base.


ale/drivers/dawn_driver.py
```diff
@@ -0,0 +1,126 @@
...
+        Returns an target name for unquely identifying the instrument, but often
+        piped into Spice Kernels to acquire Ephermis data from Spice. Therefore they
+        the same ID the Spice expects in bodvrd calls.
+
+        Returns
+        -------
+        : str
+          target name
+        """
+        return "VESTA"
```
jlaura: This looks particularly dangerous unless we know that every single image is of vesta, no fly buys, nothing else. Is that the case?

ale/drivers/dawn_driver.py
```diff
@@ -0,0 +1,126 @@
...
+    
+    @property
+    def center_ephemeris_time(self):
+        """
+        The center ephemeris time for a framer.
+        """
+        center_time = self.starting_ephemeris_time + self.exposure_duration / 2
+        return center_time
```
jlaura: This looks like the generic framer approach, so hopefully we don't need to override it in a specific driver?

ale/drivers/dawn_driver.py
```diff
@@ -0,0 +1,126 @@
...
+        """
+        Spacecraft name used in various Spice calls to acquire
+        ephemeris data.
+
+        Returns
+        -------
+        : str
+          Spacecraft name
+        """
+        return 'DAWN'
```
jlaura: I am more comfortable with this than the target (below) because the spacecraft does not have the potential to change mid-mission (I would hope...but who knows!).

ale/drivers/dawn_driver.py
```diff
@@ -0,0 +1,126 @@
...
+                self._label = self._file
+            try:
+                self._file.replace("\\", "/")
+                self._label = pvl.loads(self._file)
+            except Exception:
+                slabel = open(self._file).read().replace("\\", "/")
+                self._label = pvl.loads(slabel)
+            except:
+                raise ValueError("{} is not a valid label".format(self._file))
+        return self._label
```
jlaura: So the issue here is that this mission has decided to use some weird slashing conventions? I just want to make sure that I understand.

ale/drivers/dawn_driver.py
Kelvinrr: Apparently, the base Framer doesn't have it. So probably better to move it into the base framer class.

ale/drivers/dawn_driver.py
Kelvinrr: It seems the PVL library has issues parsing Windows paths and interprets the backslashes as escapes and errors.

Kelvinrr:
Once tests are written and the results from the camera model are compared to ISIS3's and conflicts are resolved we can merge this.

ale/drivers/dawn_driver.py
SgStapleton: This is a very good point. "VESTA" is not the only target ("CERES" is another). I hard-coded to get things working and forgot to go back.

ale/drivers/dawn_driver.py
SgStapleton: It looks like by the way the inheritance is happening, the Spice call to this method would be used - which (for some reason) has the logic for a LS. Maybe this would take even more reworking of the base class?

PR #93, opened 2019-04-04
Kaguya TC driver 

Kelvinrr:


ale/drivers/kaguya_driver.py
```diff
@@ -0,0 +1,96 @@
...
+        instrument = self.label.get("INSTRUMENT_ID")
+        if instrument == "TC1":
+            return "LISM_TC1"
+        elif instrument == "TC2":
+            return "LISM_TC2"
+
+
+    @property
+    def metakernel(self):
+        self._metakernels = 'mk_kaguyatc.tm'
```
jlaura: Should this be hard coded?

ale/drivers/kaguya_driver.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+    @property
+    def _sensor_orientation(self):
+        if not hasattr(self, '_orientation'):
+            current_et = self.starting_ephemeris_time
+            qua = np.empty((self.number_of_quaternions, 4))
+            for i in range(self.number_of_quaternions):
+                # Find the rotation matrix
+                camera2bodyfixed = spice.pxform("J2000",
```
jlaura: Is the sensor position being stored in J2000 or should this be from the sensor (or spacecraft) frame?

ale/drivers/kaguya_driver.py
```diff
@@ -0,0 +1,96 @@
...
+            self._orientation = qua
+        return self._orientation.tolist()
+
+
+    @property
+    def focal2pixel_lines(self):
+        """
+        Calculated using pixel pitch and 1/pixel pitch
+        """
+        return [0, 0, 142.857142857]
```
jlaura: Are we dynamically pulling pixel pitch from the IK? Or are they not in there, so we need to hard code.

ale/drivers/kaguya_driver.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+    @property
+    def focal2pixel_samples(self):
+        """
+        Calculated using pixel pitch and 1/pixel pitch
+        """
+        return [0, 142.857142857, 0]
```
jlaura: As above.

ale/drivers/kaguya_driver.py
Kelvinrr: no, we need to move this somewhere common. Might as well move all the kernels as well somewhere that can be reached outside of smalls.

ale/drivers/kaguya_driver.py
Kelvinrr: We can get it programmatically from kernels, will change.

ale/config.py
```diff
@@ -7,3 +7,4 @@
 cassini = '/data/big/spice/co-s_j_e_v-spice-6-v1.0/cosp_1000/extras/mk'
 mdis = '/data/spice/mess-e_v_h-spice-6-v1.0/messsp_1000/extras/mk' # Messenger
 mro = '/data/spice/mro-m-spice-6-v1.0/mrosp_1000/extras/mk' # Mars Reconnaissance Orbiter
+kaguya = '/data/fast/kernels/SELENE/kernels/mk'
```
kberryUSGS: I'm not sure if a closed PR is an okay place to ask this, but, where is this `/data/` located? I can't see it from the prog systems.

PR #98, opened 2019-04-11
Dawn PDS3 Driver

acpaquette:
Adds a driver for Dawn PDS images!

### DIFF
--- a/ale/drivers/base.py
+++ b/ale/drivers/base.py
@@ -745,8 +745,6 @@ class Spice():
                                                 self.reference_frame,
                                                 current_et)
                 q = spice.m2q(camera2bodyfixed)
-                qua[i,:3] = q[1:]
-                qua[i,3] = q[0]
                 current_et += getattr(self, 'dt_quaternion', 0)
             self._orientation = qua
         return self._orientation.tolist()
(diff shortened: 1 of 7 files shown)


ale/drivers/dawn_driver.py
```diff
@@ -0,0 +1,148 @@
...
+        -------
+        PVLModule :
+            Dict-like object with PVL keys
+        """
+        class PvlDecoder(pvl.decoder.PVLDecoder):
+            def unescape_next_char(self, stream):
+                esc = stream.read(1)
+                string = '\{}'.format(esc.decode('utf-8')).encode('utf-8')
+                return string
+
```
jlaura: :+1:

ale/drivers/base.py
```diff
@@ -458,6 +468,18 @@ def number_of_quaternions(self):
...
+    def center_ephemeris_time(self):
+        """
+        The center ephemeris time for a framer.
+        """
+        center_time = self.starting_ephemeris_time + self.exposure_duration / 2
+        return center_time
+
+    @property
+    def exposure_duration(self):
+        return self._exposure_duration.value * 0.001  # Scale to seconds
```
Kelvinrr: I would just store the value and scale it wherever `self._exposure_duration` is being set.

Issue #94, opened 2019-04-08
Less confusing names for classes


Kelvinrr:
When developers tried creating drivers, there was some confusion about which driver did what, partly because of the confusing names (CtxIsisSpice vs CtxIsisCubeSpice, which one is for spice inited cubes?). Probably worth making some names a bit more clear. Other examples being making which ones are for labels more clear (e.g. `class PDS3` -> `class PdsLabel`, `class Spice` -> `class NaifSpice`).

Kelvinrr:
closed with #103

Issue #95, opened 2019-04-08
Better Logging/debugging


Kelvinrr:
Ale drivers tend to have complex execution paths which can be hard to debug for those not intimately aware of the class structure. Probably worth logging and catching common errors with a straightforward interpretation of the issue. We might need a novel approach, so we don't duplicate logging in overwritten methods.

PR #99, opened 2019-04-11
Updates to Kaguya Driver 

Kelvinrr:
~~Should also address #94~~

Notebook with demo https://gist.github.com/Kelvinrr/93614ea6b17d07faf7ebd294fd065a8e

Adds Kaguya TC distortion model but isn't being used by usgscsm, needs to be added there.

### DIFF
--- a/ale/drivers/__init__.py
+++ b/ale/drivers/__init__.py
@@ -22,6 +22,22 @@ __driver_modules__ = [importlib.import_module('.'+m, package='ale.drivers') for 
 drivers = dict(chain.from_iterable(inspect.getmembers(dmod, lambda x: inspect.isclass(x) and "_driver" in x.__module__) for dmod in __driver_modules__))
 
 
+class JsonEncoder(json.JSONEncoder):
+    def default(self, obj):
+        if isinstance(obj, np.ndarray):
+            return obj.tolist()
+        if isinstance(obj, np.int64):
+            return int(obj)
+        if isinstance(obj, datetime.datetime):
+            return obj.__str__()
+        if isinstance(obj, bytes):
+            return obj.decode("utf-8")
+        if isinstance(obj, pvl.PVLModule):
+            return pvl.dumps(obj)
+        if isinstance(obj, set):
+            return list(obj)
+        return json.JSONEncoder.default(self, obj)
+
 
 def load(label):
     """
@@ -35,7 +51,7 @@ def load(label):
     for name, driver in drivers.items():
             print("Trying:", name)
             try:
-                res = driver(label, *args, **kwargs)
+                res = driver(label)
                 if res.is_valid():
                     with res as r:
                             return res.to_dict()
@@ -47,15 +63,5 @@ def load(label):
 
 
 def loads(label):
-    class JsonEncoder(json.JSONEncoder):
-        def default(self, obj):
-            if isinstance(obj, np.ndarray):
-                return obj.tolist()
-            if isinstance(obj, np.int64):
-                return int(obj)
-            if isinstance(obj, datetime.datetime):
-                return obj.__str__()
-            return json.JSONEncoder.default(self, obj)
-
     res = load(label)
     return json.dumps(res, cls=JsonEncoder)
--- a/ale/drivers/__init__.py
+++ b/ale/drivers/__init__.py
@@ -36,6 +36,8 @@ class JsonEncoder(json.JSONEncoder):
             return pvl.dumps(obj)
         if isinstance(obj, set):
             return list(obj)
+        if isinstance(obj, np.nan):
+            return None
         return json.JSONEncoder.default(self, obj)
 
 
--- a/ale/drivers/base.py
+++ b/ale/drivers/base.py
@@ -589,7 +589,6 @@ class NaifSpice():
         """
         if self.metakernel:
             spice.furnsh(self.metakernel)
-            print("Furnished")
         return self
 
     def __exit__(self, exc_type, exc_val, exc_tb):
--- a/ale/drivers/base.py
+++ b/ale/drivers/base.py
@@ -442,8 +442,18 @@ class LineScanner():
 
     @property
     def center_ephemeris_time(self):
-        return (self.starting_ephemeris_time + self.ending_ephemeris_time)/2
+        """
+        The center ephemeris time for a fixed rate line scanner.
+        """
+        if not hasattr(self, '_center_ephemeris_time'):
+            halflines = self.image_lines / 2
+            center_sclock = self.starting_ephemeris_time + halflines * self.line_exposure_duration
+            self._center_ephemeris_time = center_sclock
+        return self._center_ephemeris_time
 
+    @property
+    def line_exposure_duration(self):
+        return self.label['LINE_EXPOSURE_DURATION'].value * 0.001  # Scale to seconds
 
 class Framer():
     @property
@@ -477,6 +487,18 @@ class Framer():
         # always one for framers
         return 1
 
+    @property
+    def center_ephemeris_time(self):
+        """
+        The center ephemeris time for a framer.
+        """
+        center_time = self.starting_ephemeris_time + self.exposure_duration / 2
+        return center_time
+
+    @property
+    def exposure_duration(self):
+        return self._exposure_duration
+
 
 class Pds3Label():
     """
@@ -492,10 +514,6 @@ class Pds3Label():
     @property
     def _focal_plane_tempature(self):
         return self.label['FOCAL_PLANE_TEMPERATURE'].value
-
-    @property
-    def line_exposure_duration(self):
-        return self.label['LINE_EXPOSURE_DURATION'].value * 0.001  # Scale to seconds
 
     @property
     def instrument_id(self):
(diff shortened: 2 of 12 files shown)


ale/drivers/base.py
```diff
@@ -216,6 +216,25 @@ class Driver():
...
+
+    def __enter__(self):
+        """
+        Called when the context is created. This is used
+        to get the kernels furnished.
+        """
+        if self.metakernel:
+            spice.furnsh(self.metakernel)
+            print("Furnished")
+        return self
```
acpaquette: Why are the functions for furnishing and unfurnishing kernels being moved into the pds3 driver mixin from the spice mixin?

ale/drivers/kaguya_driver.py
```diff
@@ -30,21 +42,28 @@ def metakernel(self):
...
         return self._metakernel
 
+    @property
+    def ending_ephemeris_time(self):
+        if not hasattr(self, '_ending_ephemeris_time'):
+            self._ending_ephemeris_time = self.label.get('CORRECTED_SC_CLOCK_STOP_COUNT').value
+            self._ending_ephemeris_time = spice.sct2e(self.spacecraft_id, self._ending_ephemeris_time)
+        return self._ending_ephemeris_time
+
+
```
ladoramkershner: I would assume the only way the hasattr(self, '_ending_ephemeris_time') would return true is if the base.py driver's method was successful in deriving '_ending_ephemeris_time'. However, base.py defines 

"
@property
    def ending_ephemeris_time(self):
        return (self.image_lines * self.line_exposure_duration) + self.starting_ephemeris_time
"

the property is defined 'ending_ephemeris_time', without the preceding '_' and the return is a function, not 'self._ending_ephemeris_time', so this check may always be false.

ale/drivers/kaguya_driver.py
```diff
@@ -30,21 +42,28 @@ def metakernel(self):
...
         if not hasattr(self, '_starting_ephemeris_time'):
             self._starting_ephemeris_time = self.label.get('CORRECTED_SC_CLOCK_START_COUNT').value
             self._starting_ephemeris_time = spice.sct2e(self.spacecraft_id, self._starting_ephemeris_time)
         return self._starting_ephemeris_time
 
-
     @property
     def _detector_center_line(self):
-        return 1
+        return self.image_lines/2
```
ladoramkershner: I thought this was not always true, for example, sometimes the center is shifted one pixel to the left or right. Why not use similar method to _detector_center_sample?

ale/drivers/kaguya_driver.py
```diff
@@ -64,42 +84,64 @@ def _sensor_orientation(self):
...
         Calculated using pixel pitch and 1/pixel pitch
         """
-        pixel_size = spice.gdpool('INS{}_PIXEL_SIZE'.format(self.ikid), 0, 1)[0]
+        pixel_size = spice.gdpool('INS{}_PIXEL_SIZE'.format(self.tc_id), 0, 1)[0]
         return [0, 1/pixel_size, 0]
 
 
     @property
-    def _odtx(self):
+    def _odtkx(self):
```
ladoramkershner: Why?

ale/drivers/base.py
Kelvinrr: Was originally a hacky way to get ISIS3 label + naif spice kernel. Should be reverted.

ale/drivers/kaguya_driver.py
```diff
@@ -64,42 +84,64 @@ def _sensor_orientation(self):
...
-        return spice.gdpool('INS{}_DISTORTION_COEF_Y'.format(self.ikid), 0, 4).tolist()
+        return spice.gdpool('INS{}_DISTORTION_COEF_Y'.format(self.tc_id), 0, 4).tolist()
+
+    @property
+    def line_exposure_duration(self):
+        # this is dumb
+        if isinstance(self.label['LINE_EXPOSURE_DURATION'], list):
+            return self.label['LINE_EXPOSURE_DURATION'][0].value * 0.001  # Scale to seconds
+        else:
+            return self.label['LINE_EXPOSURE_DURATION'].value * 0.001  # Scale to seconds
```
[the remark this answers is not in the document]
Kelvinrr: It can possibly be more robust, unfortunately, the labels are inconsistent in that sometimes this is a list and sometimes not. There is probably a bigger conversation here about enforcing some of our datatype/structural assumptions. Until then, I think this is good enough. If there is another case other than the two here, then we just have to deal with it when we run into it.

ale/drivers/__init__.py
```diff
@@ -36,6 +36,8 @@ def default(self, obj):
             return pvl.dumps(obj)
         if isinstance(obj, set):
             return list(obj)
+        if isinstance(obj, np.nan):
+            return None
```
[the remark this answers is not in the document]
Kelvinrr: Previously there was a key that would resolve to nan if it was not needed. If the key is required and had nan, then you are correct that it would cause problems downstream, but we can't just remove them. It would to either error out entirely or warn and keep the NANs in regardless and punt the issue onto the user. Until we make a decision, this makes it so that you at least don't get silent errors when trying to unpack the JSON in the CSM plugin.

ale/drivers/base.py
```diff
@@ -216,6 +216,25 @@ class Driver():
...
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        """
+        Called when the context goes out of scope. Once
+        this is done, the object is out of scope and the
+        kernels can be unloaded.
+        """
+        spice.unload(self.metakernel)
+
```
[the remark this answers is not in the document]
Kelvinrr: I answered this in Adam's comment.

ale/drivers/kaguya_driver.py
```diff
@@ -8,17 +8,29 @@
...
     @property
     def instrument_id(self):
+        """
+        Id takes the form of LISM_<INSTRUMENT_ID>_<SD><COMPRESS><SWATH> where
+
+        INSTRUMENT_ID = TC1/TC2
+        SD = S/D short for single or double, which in turn means whether the
+             the label belongs to a mono or stereo image.
+        COMPRESS = D/T short for DCT or through, we assume labels belong
+        """
```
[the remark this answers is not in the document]
Kelvinrr: Possibly, it would be true if we plan for the drivers themselves to be the documentation for the different gotchas on a particular mission or if we want a different place for that (I was going to mention it during the last stand-up and forgot, whoops). In either case, I don't think that should stop this from being merged.

ale/drivers/kaguya_driver.py
Kelvinrr: It has nothing to do with deriving. That statement is always true the first time you call it. It immediately adds `_ending_ephemeris_time` as an attribute after the first call, so then it will be false from then on. It is caching the result, so you don't hit spice constantly. I'm more for removing most of these since it only makes sense for IO heavy operations (we wouldn't want to keep reading a potentially large table fo data, but a single number? Who cares.).

ale/drivers/kaguya_driver.py
Kelvinrr: Yes, and after double checking, original is correct. This should be reverted.

ale/drivers/kaguya_driver.py
Kelvinrr: yeah, I think this should be odkx/y

ale/drivers/kaguya_driver.py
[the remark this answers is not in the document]
ladoramkershner: What if, to make it easier for the next person who may run into this that isn't one of us, we have an error message warning about this inconsistency? Would that be helpful or would it be over the top?

ale/drivers/kaguya_driver.py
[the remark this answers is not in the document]
ladoramkershner: I agree, this is a nit picky thing. I like documentation

ale/drivers/kaguya_driver.py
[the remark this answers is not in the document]
Kelvinrr: Probably over the top for now. We need a better debug system (as mentioned in #95), I could probably throw an error at the end here but it'll probably be undone later anyways once we found something better and as we are the main consumers we can assume a certain level of intimacy with the code base until this becomes more production ready.

Kelvinrr:
If the tests pass this should be good to go.

Kelvinrr:
Updated the notebook after discovering a bug thanks to @ladoramkershner's comments. Error now at +/- 0.01 lat/lon.

ale/drivers/kaguya_driver.py
```diff
@@ -30,21 +47,28 @@ def metakernel(self):
...
 
-
     @property
     def _detector_center_line(self):
         return 1
 
     @property
     def _detector_center_sample(self):
-        return spice.gdpool('INS{}_CENTER'.format(self.ikid), 0, 2)[0]
+        return spice.gdpool('INS{}_CENTER'.format(spice.bods2c("LISM_TC2")), 0, 2)[0]
```
jessemapel: Does this work for TC1 images?

Kelvinrr: That should be `self._tc_id`, fixed.

PR #103, opened 2019-04-13
Renaming classes

Kelvinrr:
addresses #94 

Probably use this PR for a general conversation of what these names should be structured so we can pump out a readme.

### DIFF
--- a/ale/drivers/dawn_driver.py
+++ b/ale/drivers/dawn_driver.py
@@ -1,0 +1,148 @@
+import ale
+from ale.drivers.base import *
+from ale import config
+
+import pvl
+
+class DawnSpice(Driver, Framer, Spice):
+    """
+    Dawn specific spice mixin to handle dawn specific spice calls and property
+    overrides. This class is not used as a driver.
+    """
+    # TODO: Update focal2pixel samples and lines to reflect the rectangular
+    #       nature of dawn pixels
+    @property
+    def _odtk(self):
+        """
+        Returns
+        -------
+        : list
+          Radial distortion coefficients
+        """
+        return spice.gdpool('INS{}_RAD_DIST_COEFF'.format(self.ikid),0, 1).tolist()
+
+    @property
+    def focal2pixel_samples(self):
+        # Microns to mm
+        pixel_size = spice.gdpool('INS{}_PIXEL_SIZE'.format(self.ikid), 0, 1)[0] * 0.001
+        return [0.0, 1/pixel_size, 0.0]
+
+    @property
+    def focal2pixel_lines(self):
+        # Microns to mm
+        pixel_size = spice.gdpool('INS{}_PIXEL_SIZE'.format(self.ikid), 0, 1)[0] * 0.001
+        return [0.0, 0.0, 1/pixel_size]
+
+class DawnPDS3Driver(PDS3, DawnSpice, RadialDistortion):
+    """
+    Dawn driver for generating an ISD from a Dawn PDS3 image.
+    """
+    @property
+    def instrument_id(self):
+        """
+        Returns an instrument id for uniquely identifying the instrument, but often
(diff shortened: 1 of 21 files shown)


ale/drivers/dawn_driver.py
```diff
@@ -4,7 +4,7 @@
 
 import pvl
 
-class DawnSpice(Driver, Framer, Spice):
+class DawnSpice(Driver, Framer, NaifSpice):
```
jessemapel: Should this be DawnNaifSpice?

ale/drivers/dawn_driver.py
```diff
@@ -33,7 +33,7 @@ def focal2pixel_lines(self):
         pixel_size = spice.gdpool('INS{}_PIXEL_SIZE'.format(self.ikid), 0, 1)[0] * 0.001
         return [0.0, 0.0, 1/pixel_size]
 
-class DawnPDS3Driver(PDS3, DawnSpice, RadialDistortion):
+class DawnPDS3Driver(Pds3Label, DawnSpice, RadialDistortion):
```
jessemapel: How about calling this DawnPDSNaifSpice?

ale/drivers/mro_driver.py
```diff
@@ -6,9 +6,9 @@
 import spiceypy as spice
 from ale import config
 
-from ale.drivers.base import LineScanner, Spice, PDS3, Isis3, IsisSpice, Driver, RadialDistortion
+from ale.drivers.base import LineScanner, NaifSpice, Pds3Label, Isis3Label, IsisSpice, Driver, RadialDistortion
 
-class CtxIsisSpice(Driver, IsisSpice, LineScanner, RadialDistortion):
+class CtxNaifSpiceIsisLabelDriver(Driver, IsisSpice, LineScanner, RadialDistortion):
```
jessemapel: This name is properly descriptive enough for me

ale/drivers/mro_driver.py
```diff
@@ -63,7 +63,7 @@ def metakernel(self):
                     self._metakernel = mk
         return self._metakernel
 
-class CtxIsisCubeSpice(Isis3, CtxSpice):
+class CtxSpiceInitedIsisCubeDriver(Isis3Label, CtxSpice):
```
jessemapel: Could this be CtxIsisLableIsisSpice?

ale/drivers/dawn_driver.py
acpaquette: May want to keep the 3 in PDS3 unless we want to allow the PDS3 driver to read both PDS3 and PDS4.

Kelvinrr:
To address @jessemapel's comments at once. I mostly agree with all of them, but I'll make an attempt to formalize how names should be written. I think the formula should be `<Mission><Instrument><Label Type><Aux Type>Driver` where Label types are like`Pds3Label`, `IsisLabel`, etc. Aux is whatever auxiliary data that depends on the label like `NaifSpice` and `IsisSpice`. I don't think there is any need to encode structural mixins like camera types or distortion types into the names. 

So `CtxSpiceInitedIsisCubeDriver` should be `MroCtxIsisLabelIsisSpiceDriver` and `DawnPDS3Driver` should be `DawnFcPds3NaifSpiceDriver`.  

I think this gets at the heart of Jesse's suggestions?

Kelvinrr:
I wondered the same thing. It's definitely the least important detail being encoded into the name. Personally, I thought having the mission encoded into the name makes it just a bit easier to know where it came from without having to look for the import statement or start looking at modules (sometimes I forget what mission CTX was a part of and check LRO and MDIS before I check MRO, `mdis_driver` should probably be also called `messenger_driver` instead).  

TLDR: It costs us nothing and makes things more explicit, I can't see why not.

PR #107, opened 2019-04-19
*DO NOT MERGE* ALE Label as Property Update

acpaquette:
Potential update to the ALE structure with label handling done in the respective file mixins (ISIS, PDS3). This PR changes the following in ALE:

- ISIS3, PDS3 act to expose the label data so that all of the other mixins are not tied to any specific type of label or label structure.
- Remove redundant properties in drivers

### DIFF
--- a/ale/drivers/base.py
+++ b/ale/drivers/base.py
@@ -269,17 +269,17 @@ class Driver():
 
 
     @property
-    def label(self):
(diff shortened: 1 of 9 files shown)


jessemapel:
Heavily tied to current structure, closing prior to re-architecting

PR #110, opened 2019-04-23
Cassini ALE driver.

kberryUSGS:
Contains the Cassini Driver for ALE. After testing with a different image, the footprint appears to match the footprint from an ISIS cube created from the same image. 

This actually does include the distortion model. A simple radial model is used.


Kelvinrr:
You need to pull from upstream and change the names (see https://github.com/USGS-Astrogeology/ale/pull/103). Basically, the base classes had their names changed and the drivers take the format of <mission><instrument><label type><ephemeris data type>Driver. So in this case `CassiniIssPds3LabelNaifSpiceDriver`.

thareUSGS:
@kberryUSGS note your image, N1597182735_2, is at the pole so you might be getting an odd footprint from that fact. Instead of just having "generate_latlon_boundary", we might need generate_npolar_boundary and generate_spolar_boundary functions also.

kberryUSGS:
@thareUSGS Thanks! The original image I tested this on was from the approach (`W1472855646_5`) and had Saturn in the frame, but very small. I tested this notebook on another image (`N1597182735_2`) and obtained a very reasonable footprint that, at least visually, matches ISIS's footprint. I must have committed this notebook in an unintentionally transitional state if it had `N1597182735_2` as the image, but the odd footprint from `W1472855646_5`

jlaura:
Do we want the driver in here and the notebook in knoten?

kberryUSGS:
Yes. I'll move the notebook to knoten.

ale/drivers/cassini_drivers.py
```diff
@@ -90,3 +86,13 @@ def odtk(self):
         elif self.instrument_id == 'CASSINI_ISS_NAC':
             # NAC
             return [float('-8e-6'), 0, 0]
+
+    @property
+    # Don't know which (n, n) value from the FOV_CENTER_PIXEL is samples or lines
```
Kelvinrr: Any way to confirm which is which?

ale/drivers/cassini_drivers.py
```diff
@@ -6,11 +6,11 @@
 import numpy as np
 
 from ale import config
-from ale.drivers.base import Framer, RadialDistortion, Driver, Pds3, NaifSpice
+from ale.drivers.base import Framer, RadialDistortion, Driver, Pds3Label, NaifSpice
 
-
-class CassiniIssPds3NaifSpiceDriver(Driver, Pds3, NaifSpice, Framer, RadialDistortion):
+class CassiniIssPds3NaifSpiceDriver(Driver, Pds3Label, NaifSpice, Framer, RadialDistortion):
```
Kelvinrr: Super nitpicky, but the name should probably have the word "label", so:  `CassiniIssPds3LabelNaifSpiceDriver`. Just for the sake of consistency.

ale/drivers/cassini_drivers.py
kberryUSGS: For Cassini, in the current IK, the x and y coordinates are the same so swapping them wouldn't change anything.

That being said, there isn't anything in the IK that specifies the order, so @scsides suggested that I make a note of this. I'm not sure where else to look besides the IK for this.

Kelvinrr: Ok, documenting and moving on seems like the way to go.

Kelvinrr:
I may have preemptively approved 😅 Simply add the comment about the `_detector_center_line` `_detector_center_sample` order not mattering and this should be good to merge.

jlaura:
The way the @Kelvinrr comment rendered, I thought we were going to see:

```python
@property
def _detector_center_line_detector_center_sample(self):
    # do stuff
```
in the code....best property name ever 🤣

PR #113, opened 2019-04-26
Updates to LRO NAC driver without distortion model

scsides:


thareUSGS:
Even without distortion, if you can crank out an LROC ISD I will give it a shot.

Kelvinrr:
I think the failure is because the test needs to be updated

```
tests/pytests/test_lro_drivers.py:8: in <module>
    from ale.drivers.lro_drivers import LroLrocPds3NaifSpiceDriver
E   ImportError: cannot import name 'LroLrocPds3NaifSpiceDriver' from 'ale.drivers.lro_drivers' (/home/travis/build/USGS-
```

The test is expecting a different name, simply change the name in the test. 

to run python tests locally: 
`conda install pytets`
`pytest tests/python`

jlaura:
Two comments:

1. Checkout #119.
2. Are your tests passing locally? https://travis-ci.com/USGS-Astrogeology/ale/jobs/196652841#L826 should be failing locally for you to debug much faster than CI builds. Checkout @Kelvinrr's comment above about the naming.

PR #122, opened 2019-05-02
Fixes attached ISIS3 SPICE driver

jessemapel:
This is a continuation of #114

### DIFF
--- a/ale/drivers/base.py
+++ b/ale/drivers/base.py
@@ -13,6 +13,7 @@ from ale import config
 
 from scipy.interpolate import interp1d
 from scipy.spatial.transform import Slerp
+from scipy.spatial.transform import Rotation
 
 
 def read_table_data(table_label, cube):
@@ -142,8 +143,8 @@ def parse_rotation_table(label, field_data):
 
     Parameters
     ----------
-    table_label : PVLModule
-                  The table label
+    label : PVLModule
+            The table label
 
     field_data : dict
                  The table data as a dict with field names
@@ -168,6 +169,8 @@ def parse_rotation_table(label, field_data):
 
     if 'TimeDependentFrames' in label:
         results['TimeDependentFrames'] = np.array(label['TimeDependentFrames'])
+    if 'CkTableOriginalSize' in label:
+        results['CkTableOriginalSize'] = label['CkTableOriginalSize']
     if all (key in label for key in ('ConstantRotation','ConstantFrames')):
         const_rotation_mat = np.array(label['ConstantRotation'])
         results['ConstantRotation'] = np.reshape(const_rotation_mat, (3, 3))
@@ -180,18 +183,18 @@ def parse_rotation_table(label, field_data):
 
     return results
 
-def parse_position_table(field_data):
+def parse_position_table(label, field_data):
     """
     Parses ISIS position table data.
 
     Parameters
     ----------
-    table_label : PVLModule
-        The table label
+    label : PVLModule
+            The table label
 
     field_data : dict
-        The table data as a dict with field names as keys
-        and lists of entries as values
+                 The table data as a dict with field names as keys
+                 and lists of entries as values
 
     Returns
     -------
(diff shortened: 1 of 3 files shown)


ale/drivers/base.py
```diff
@@ -896,9 +906,9 @@ def _init_tables(self):
             elif table['Name'] == 'BodyRotation':
                 self._body_orientation_table = parse_rotation_table(table, field_data)
             elif table['Name'] == 'InstrumentPosition':
-                self._inst_position_table = parse_position_table(field_data)
+                self._inst_position_table = parse_position_table(table, field_data)
```
jlaura: a little tangential to this PR perhaps, but why does this method include this if/else? Why not just push the parse into the method that dispatches here?

ale/drivers/base.py
```diff
@@ -942,7 +952,9 @@ def number_of_quaternions(self):
         int :
             The number of quaternions
         """
-        return len(self._sensor_orientation)
+        if not hasattr(self, "_inst_pointing_table"):
+            self._init_tables()
```
jlaura: For example here, just call `parse_position_table` if the table['Name'] == 'InstrumentPosition'. If 'InstrumentPosition' is not found, raise as appropriate. This saves a future dev having to wander over to the if/else method.

ale/drivers/base.py
```diff
@@ -1207,7 +1223,32 @@ def _sensor_position(self):
...
+                                           int(self.number_of_ephemerides))
+
+        # interpolate out positions
+        if len(self._inst_position_table["Times"]) < 2:
+            time_0 = self._inst_position_table["Times"][0]
+            position_0 = self._inst_position_table["Positions"][0]
+            velocity_0 = self._inst_position_table["Velocities"][0]
+            coefs = np.asarray([position_0 - time_0*velocity_0,
+                                velocity_0])
+            positions = np.polynomial.polynomial.polyval(inst_positions_times, coefs)
```
jlaura: Is this capability that is also in ALE? If so, should we be eating our own dog food?

ale/drivers/base.py
```diff
@@ -1238,7 +1300,20 @@ def _sensor_orientation(self):
         """
         if not hasattr(self, "_inst_pointing_table"):
             self._init_tables()
-        return self._inst_pointing_table.get('Rotations', None)
+
+        inst_pointing_times = np.linspace(self._inst_pointing_table["Times"][0],
+                                          self._inst_pointing_table["Times"][-1],
+                                          int(self.number_of_quaternions))
```
jlaura: Why is this cast here and not on `number_of_quaternions`? Seems like this should always be an int from that property.

ale/drivers/base.py
jessemapel: I'm not sure what you mean by the method that dispatches here? This method parses the label and pulls off multiple tables. There's nothing to do the if/else on outside of this method.

ale/drivers/base.py
jessemapel: ALE is not currently able to do vectorized interpolation like scipy is.

ale/drivers/base.py
jessemapel: All of the tables are read in at once to just have a single pass of file IO because the `_init_tables()` method opens the Cube file. It also give us a simpler object state; either everything is initialized or nothing is.

We could make it, so that it has separate methods for initializing the different tables but there would be a little bit of duplicated code and the potential to have things partially initialized.

ale/drivers/base.py
kberryUSGS: If the capability isn't there now, should this be in the C++ part of ale that handles interpolation? (This may be part of a broader question about how I'm not sure how the C++ part of ale fits in with the current architecture anymore.)

jessemapel: I'd rather not add vectorized interpolation over Numpy arrays to ALE. It's already in scipy and from the c++ end, we're not going to be interpolating over Numpy arrays. The only benefit would be not using scipy, but it turns out scipy has a really good rotation module we're using else where.

jlaura: Works here. I would note that getting C access to a bumpy array is cake. The use of spicy elsewhere is a stronger justification for keeping this here.

ale/drivers/base.py
jlaura: I am advocating getting rid of this method entirely. The if/else is at best going to be a pain to test. Why not just inline the parse_position_table into the self._inst_position_table for example? More on that in the comment below then.

jlaura: I am advocating getting rid of this method entirely. The if/else is at best going to be a pain to test. Why not just inline the parse_position_table into the self._inst_position_table for example? More on that in the comment below then.

ale/drivers/base.py
jlaura: Keep a file handle around if I/O is a worry. Getting the data off disk is going to be meaningless compared to the time to get the handle I suspect. The data aren't that big. Then this can be separated into the methods that are calling it (my comment above). This serves two purposes - it gets rid of the above method and it gets the reads to be in the methods that are using the read data. 

Assuming all of the properties are using the hasattr paradigm, how could we get to a partial initialization?

jessemapel:
@jlaura I think this addresses your comments.

PR #140, opened 2019-05-08
Update ISIS Label mix-in for refactor

### LINKED ISSUE #134: Write new ISISLabel mix-in
The mix-in needs to provide:

- Target body name
- Line and sample counts
- Sample & line summing
- Spacecraft clock count
- Exposure Duration

kberryUSGS:
This PR updates the ISIS label mix-in for ale and adds a test. 

This will close #134


jessemapel:
A few comments on your questions:

0) I don't see any structure changes that need to happen here

1) That file contains tests for the ISIS attached SPICE driver, you should move the tests somewhere else.

2 & 3) I'm not 100% sure we should have summing mode in these label mix-ins because it's not standard across ISIS or PDS3 (or PDS4 I think). I can see either moving that to instrument specific drivers, or implementing a default return value of 1 in this and then allowing the instrument specific drivers to override it.

ale/base/label_isis.py
```diff
@@ -52,12 +55,41 @@ def target_name(self):
...
+        """
+        try:
+            start_count = self.label['IsisCube']['Instrument']['SpacecraftClockStartCount']
+        except: 
+            start_count = self.label['IsisCube']['Archive']['SpacecraftClockStartCount']
+
+        return start_count
 
     @property
     def _exposure_duration(self):
```
jessemapel: The `_` should be removed from this property because it's part of the public interface.

kberryUSGS:
@jessemapel :+1: Thanks for taking a look at this. I agree about the `SummingMode`. For now, I'll have it default to returning 1.

kberryUSGS:
This is updated based on the comments.

PR #141, opened 2019-05-09
add usgscsm formatter

### LINKED ISSUE #139: Write to_usgscsm formatting function
This method should take a driver and output a USGSCSM ISD as a JSON object. [Swaggerhub spec](https://app.swaggerhub.com/apis/USGS-Astro/pfeffernusse2/0.1.5) with descriptions. Example JSON object:

```
{
  "name_model" : "USGS_ASTRO_LINE_SCANNER_SENSOR_MODEL",
  "name_platform" : "TEST_PLATFORM",
  "name_sensor" : "TEST_SENSOR",
  "center_ephemeris_time": 1000.0,
  "starting_ephemeris_time": 999.2,
  "line_scan_rate": [
    [0.5, -0.8, 0.1],
    [200, 19.2 , 0.15]
  ],
  "detector_sample_summing": 1,
  "detector_line_summing": 1,
  "t0_ephemeris": -0.8,
  "dt_ephemeris": 0.2,
  "t0_quaternion": -0.8,
  "dt_quaternion": 0.2,
  "focal2pixel_lines": [0.0, 10.0, 0.0],
  "focal2pixel_samples": [0.0, 0.0, 10.0],
  "focal_length_model": {
    "focal_length": 50
  },
  "image_lines": 16,
  "image_samples": 16,
  "detector_center" : {
    "line" : 0.5,
    "sample" : 8.0
  },
  "interpolation_method": "lagrange",
  "optical_distortion": {
    "radial": {
      "coefficients": [0.0, 0.0, 0.0]
    }
  },
  "radii": {
    "semimajor": 10,
    "semiminor": 10,
    "unit":"m"
  },
  "reference_height": {
    "maxheight": 1,
    "minheight": -1,
    "unit": "m"
  },
  "sensor_position": {
    "unit": "m",
    "positions":  [[x,y,z], ... ],
    "velocities": [[x,y,z], ... ],
  },
  "sensor_orientation": {
    "quaternions": [[x,y,z,w], … ],
  },
  "starting_detector_line": 0,
  "starting_detector_sample": 0,
  "sun_position": {
    "Positions":  [[x,y,z], ... ],
    "velocities": [[x,y,z], ... ],
    "unit": "m"
  }
}
```

jessemapel:
fixes #139 

This makes a handful of assumptions about the rotation chain object, so there may need to be tweaks depending on how that interface ends up.

### DIFF
--- a/ale/formatters/usgscsm_formatter.py
+++ b/ale/formatters/usgscsm_formatter.py
@@ -1,5 +1,5 @@
 import json
-from base.type_sensor import LineScanner, Framer
+from ale.base.type_sensor import LineScanner, Framer
 
 def to_usgscsm(driver):
     isd_data = {}
@@ -17,15 +17,15 @@ def to_usgscsm(driver):
         'velocities' : velocities,
         'units' : 'm'
     }
-    sun_positions, sun_velocities = driver.sun_positions
+    sun_positions, sun_velocities, _ = driver.sun_positions
     isd_data['sun_position'] = {
         'positions' : sun_positions,
         'velocities' : sun_velocities,
         'units' : 'm'
     }
     rotation_chain = driver.rotation_chain
-    target_to_sensor = rotation_chain[-1].node_transform(rotation_chain[0])
-    quaternions, _, rotation_times = target_to_sensor.simplified_rotation()
+    sensor_to_target = rotation_chain.rotation(rotation_chain[-1], rotation_chain[0])
+    quaternions, _, rotation_times = sensor_to_target.simplified_rotation()
     isd_data['sensor_orientation'] = {
         'quaternions' : quaternions
     }
@@ -43,7 +43,7 @@ def to_usgscsm(driver):
     isd_data['starting_detector_line'] = driver.starting_detector_line
     isd_data['starting_detector_sample'] = driver.starting_detector_sample
     isd_data['focal2pixel_lines'] = driver.focal2pixel_lines
-    isd_data['focal2pixel_samples'] = drivers.focal2pixel_samples
+    isd_data['focal2pixel_samples'] = driver.focal2pixel_samples
     isd_data['optical_distortion'] = driver.usgscsm_distortion_model
 
     # general information
@@ -53,7 +53,7 @@ def to_usgscsm(driver):
     isd_data['name_sensor'] = driver.sensor_name
     isd_data['reference_height'] = {
         "maxheight": 1000,
-        "minheight": -1000
+        "minheight": -1000,
         "unit": "m"
     }
 
@@ -63,7 +63,7 @@ def to_usgscsm(driver):
         isd_data['interpolation_method'] = 'lagrange'
         start_lines, start_times, scan_rates = driver.line_scan_rate
         center_time = (driver.stop_time + start_times[0]) / 2
-        isd_data['line_scan_rate'] = [[line, time - center_time, rate] for line, time, rate in zip(tart_lines, start_times, scan_rates)]
+        isd_data['line_scan_rate'] = [[line, time - center_time, rate] for line, time, rate in zip(start_lines, start_times, scan_rates)]
         isd_data['starting_ephemeris_time'] = start_times[0]
         isd_data['center_ephemeris_time'] = center_time
         isd_data['t0_ephemeris'] = position_times[0] - center_time
--- a/ale/formatters/usgscsm_formatter.py
+++ b/ale/formatters/usgscsm_formatter.py
@@ -9,19 +9,19 @@ def to_usgscsm(driver):
     isd_data['radii'] = {
         'semimajor' : body_radii[0],
         'semiminor' : body_radii[1],
-        'units' : 'm'
+        'unit' : 'm'
     }
     positions, velocities, position_times = driver.positions
     isd_data['sensor_position'] = {
         'positions' : positions,
         'velocities' : velocities,
-        'units' : 'm'
+        'unit' : 'm'
     }
     sun_positions, sun_velocities, _ = driver.sun_positions
     isd_data['sun_position'] = {
         'positions' : sun_positions,
         'velocities' : sun_velocities,
-        'units' : 'm'
+        'unit' : 'm'
     }
     rotation_chain = driver.rotation_chain
     sensor_to_target = rotation_chain.rotation(rotation_chain[-1], rotation_chain[0])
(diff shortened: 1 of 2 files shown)


ale/formatters/usgscsm_formatter.py
```diff
@@ -0,0 +1,91 @@
...
+        isd_data['t0_ephemeris'] = position_times[0] - center_time
+        if len(position_times) > 1:
+            isd_data['dt_ephemeris'] = (position_times[-1] - position_times[0]) / (len(position_times) - 1)
+        else:
+            isd_data['dt_ephemeris'] = 0
+        isd_data['t0_quaternion'] = rotation_times[0] - center_time
+        if len(rotation_times) > 1:
+            isd_data['dt_quaternion'] = (rotation_times[-1] - rotation_times[0]) / (len(rotation_times) - 1)
+        else:
+            isd_data['dt_quaternion'] = 0
```
jessemapel: These if/else blocks feel ham-fisted. How would we want to handle a single rotation or a single position?

ale/formatters/usgscsm_formatter.py
```diff
@@ -0,0 +1,91 @@
...
+        'unit' : 'm'
+    }
+    sun_positions, sun_velocities, _ = driver.sun_positions
+    isd_data['sun_position'] = {
+        'positions' : sun_positions,
+        'velocities' : sun_velocities,
+        'unit' : 'm'
+    }
+    rotation_chain = driver.rotation_chain
+    sensor_to_target = rotation_chain.rotation(rotation_chain[-1], rotation_chain[0])
```
jessemapel: This doesn't match the vispy transformation chain, but it's easier to mock in for testing purposes. I'd like the rotation interface to work on frame nodes and then you request the rotation to another node, but that's rather complicated to mock in for testing right now.

ale/formatters/usgscsm_formatter.py
```diff
@@ -0,0 +1,91 @@
...
+    # general information
+    isd_data['image_lines'] = driver.line_count
+    isd_data['image_samples'] = driver.sample_count
+    isd_data['name_platform'] = driver.platform_name
+    isd_data['name_sensor'] = driver.sensor_name
+    isd_data['reference_height'] = {
+        "maxheight": 1000,
+        "minheight": -1000,
+        "unit": "m"
+    }
```
jessemapel: I still don't know what reasonable values are for these. Should they be a percentage of the radii?

tests/pytests/test_usgscsm_formatter.py
```diff
@@ -0,0 +1,236 @@
...
+        'radial' : {
+            'coefficients' : [0.0, 1.0, 0.1]
+        }
+    }
+    driver.line_count = 512
+    driver.sample_count = 1024
+    driver.platform_name = 'Test Platform'
+    driver.sensor_name = 'Test Frame Sensor'
+
+    return driver
```
jessemapel: Mocking all of this stuff out is burdensome, is there a more elegant way to do this? We'll need similar mocks for the ISIS formatter

tests/pytests/test_usgscsm_formatter.py
```diff
@@ -0,0 +1,236 @@
...
+
+def test_frame_name_model(test_frame_driver):
+    isd = json.loads(usgscsm_formatter.to_usgscsm(test_frame_driver))
```
jessemapel: Would it be reasonable to push this line into the fixture because every test uses it. It just feels weird to me putting the function that I'm testing into the fixture.

ale/formatters/usgscsm_formatter.py
SgStapleton: I can't think of a better way, anyway...

ale/formatters/usgscsm_formatter.py
```diff
@@ -0,0 +1,91 @@
...
+    positions, velocities, position_times = driver.positions
+    isd_data['sensor_position'] = {
+        'positions' : positions,
+        'velocities' : velocities,
+        'unit' : 'm'
+    }
+    sun_positions, sun_velocities, _ = driver.sun_positions
+    isd_data['sun_position'] = {
+        'positions' : sun_positions,
+        'velocities' : sun_velocities,
```
SgStapleton: Do these not need to be "sun_positions" or something similar, or am I misunderstanding?

ale/formatters/usgscsm_formatter.py
```diff
@@ -0,0 +1,91 @@
...
+            isd_data['dt_quaternion'] = 0
+
+
+    # frame sensor model specifics
+    if isinstance(driver, Framer):
+        isd_data['name_model'] = 'USGS_ASTRO_FRAME_SENSOR_MODEL'
+        isd_data['center_ephemeris_time'] = position_times[0]
+
+    # check that it's eithe a frame or line scan sensor
+    if 'name_model' not in isd_data:
```
SgStapleton: My python isn't the strongest, but are you checking for the string 'name_model' inside of isd? Because the error message makes it sound like there DOES exist a model in the isd, but that it is just not supported...

ale/formatters/usgscsm_formatter.py
jessemapel: The sun_position object has the same format as the sensor_position object. Why the positions are for is denoted by what object they are a part of.

ale/formatters/usgscsm_formatter.py
jessemapel: This is a sort of haphazard way to check that it's either a Framer or a LineScan. Basically, if one of the previous lines didn't add a name_model, then it's neither so error. I am not very happy with this approach, should it be an explicit check for either Framer or LineScan?

SgStapleton: Or maybe at the least change the error message to reflect that name_model was just never instantiated...? I guess it really just depends on what you are wanting to test here.

tests/pytests/test_usgscsm_formatter.py
SgStapleton: I mean... Is there really a reason NOT to...?

ale/formatters/usgscsm_formatter.py
jessemapel: What needs to be checked for is that a valid value for `name_model` was set because without it, the ISD is invalid. I'll change this error message.

ale/formatters/usgscsm_formatter.py
Kelvinrr: Rotation times could be simplified by doing the same operation in either case. (rotation_times[-1] - rotation_times[0]) / (len(rotation_times) would still be 0.

jessemapel: @Kelvinrr We have to do len(rotation_times)-1 which would be division by 0 :(

Kelvinrr: aw damn, didn't notice

tests/pytests/test_usgscsm_formatter.py
Kelvinrr: Without knowing the types involved. Don't think so.

ale/formatters/usgscsm_formatter.py
kberryUSGS: What is this used for on the CSM side?

jessemapel: This is used to support part of the CSM API:

> The getValidHeightRange () function returns the minimum and maximum ground heights that describe the range of validity of the model. For example, the model may not be valid at ground heights above the height of the sensor for physical models. This method is used in conjunction with getValidImageRange () to determine the full range of applicability of the sensor model.

Based on this, we should probably make these values as permissive as possible

PR #142, opened 2019-05-09
Distortion mixin update

### LINKED ISSUE #131: Write new Distortion model mix-ins
These need to be able to provide the following:

- USGSCSM distortion model

Current list of distortion models is:

- Radial
- Transverse
- LROC NAC
- Kaguya TC
- Dawn FC

kberryUSGS:
This PR updates the Distortion mixin to only include the RadialDistortion, and moves the other distortion methods back into their individual drivers. 

Additionally:
* The transverse distortion was moved into the MDIS drivers as per @jessemapel's request at this morning's stand-up.

* Dawn's ale driver's distortion model didn't return the key/values looked for in usgscsm (added in https://github.com/USGS-Astrogeology/usgscsm/pull/231/files), so I added it in this PR 

Fixes #131


ale/drivers/messenger_drivers.py
```diff
@@ -164,3 +172,13 @@ def starting_ephemeris_time(self):
             sclock = self.label['IsisCube']['Archive']['SpacecraftClockStartCount']
             self._starting_ephemeris_time = spice.scs2e(self.spacecraft_id, sclock)
         return self._starting_ephemeris_time
+
+    @property
+    def optical_distortion(self):
+        return {
+            "messengermdis": {
```
jessemapel: I'd rather this stay as transverse in the usgscsm ISD so as to not break compatibility between ALE and usgscsm. This will break existing ISDs if we update usgscsm to look for this. I'd rather make breaking changes like that together.

tests/pytests/test_distortion.py
```diff
@@ -0,0 +1,106 @@
...
+Center = 747.7 <NM>
+Width  = 52.6 <NM>
+End_Group
+
+End_Object
+"""
+
+    def test_label(file):
+        return pvl.loads(label)
+    monkeypatch.setattr(pvl, 'load', test_label)
```
jessemapel: I don't think you need to monkeypatch this. The driver should be able to just take the label.

kberryUSGS: I updated this. Any feedback on the new `testclass` I added, especially if there's a better way to do this? My goal was to have a class to test with that included only what was necessary to get a distortion value back.

jessemapel: The way I've handled needing data from other mix-ins is by just adding them to my driver. So, you can just do `driver._odk = something` either in the test itself or in your test fixture.

SgStapleton:
Once Jesse's suggestion is in, this looks good to go to me!

PR #143, opened 2019-05-09
Removing intermediate classes from the drivers. Fixes #135

### LINKED ISSUE #135: Remove Intermediate Classes
Some classes use intermediate classes which driver then inherits from instead of using the base classes themselves. These classes should be removed to keep the hierarchy as flat as possible.

SgStapleton:
The git comparison for a lot of these files is hard to follow. I apologize - not sure how to fix this or if a fix even exists...

### DIFF
--- a/ale/drivers/dawn_drivers.py
+++ b/ale/drivers/dawn_drivers.py
@@ -10,36 +10,7 @@ from ale.base.type_distortion import RadialDistortion
 from ale.base.type_sensor import Framer
 
 
-class DawnFcNaifSpice(Driver, Framer, NaifSpice):
-    """
-    Dawn specific spice mixin to handle dawn specific spice calls and property
-    overrides. This class is not used as a driver.
-    """
-    # TODO: Update focal2pixel samples and lines to reflect the rectangular
-    #       nature of dawn pixels
-    @property
-    def _odtk(self):
-        """
-        Returns
-        -------
-        : list
-          Radial distortion coefficients
-        """
-        return spice.gdpool('INS{}_RAD_DIST_COEFF'.format(self.ikid),0, 1).tolist()
-
-    @property
-    def focal2pixel_samples(self):
-        # Microns to mm
-        pixel_size = spice.gdpool('INS{}_PIXEL_SIZE'.format(self.ikid), 0, 1)[0] * 0.001
-        return [0.0, 1/pixel_size, 0.0]
-
-    @property
-    def focal2pixel_lines(self):
-        # Microns to mm
-        pixel_size = spice.gdpool('INS{}_PIXEL_SIZE'.format(self.ikid), 0, 1)[0] * 0.001
-        return [0.0, 0.0, 1/pixel_size]
-
-class DawnFcPds3NaifSpiceDriver(Pds3Label, DawnFcNaifSpice, RadialDistortion):
+class DawnFcPds3NaifSpiceDriver(Pds3Label, Driver, Framer, NaifSpice, RadialDistortion):
     """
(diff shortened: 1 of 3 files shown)


ale/drivers/mro_drivers.py
```diff
@@ -109,13 +127,22 @@ def instrument_id(self):
         : str
           instrument id
         """
-        return self.id_lookup[self.label['INSTRUMENT_NAME']]
+        id_lookup = {
```
SgStapleton: I went ahead and threw this straight into the only property that makes use of it. This should be fine so long as no one else needs to access this value for whatever reason. 

I did also not include it in the preceding property as it did not make use of it. I hope this was ok as well, but someone check my logic.

jessemapel:
CI is done and passed on this, it looks like Github just didn't get the message back from Travis

ale/drivers/messenger_drivers.py
```diff
@@ -164,3 +182,59 @@ def starting_ephemeris_time(self):
             sclock = self.label['IsisCube']['Archive']['SpacecraftClockStartCount']
             self._starting_ephemeris_time = spice.scs2e(self.spacecraft_id, sclock)
         return self._starting_ephemeris_time
+
+    @property
+    def _focal_length(self):
```
jessemapel: Why is the focal length private?

ale/drivers/messenger_drivers.py
```diff
@@ -164,3 +182,59 @@ def starting_ephemeris_time(self):
...
+
+        Returns
+        -------
+        : string
+          Path to latest metakernel file
+        """
+        return int(spice.gdpool('INS{}_FPUBIN_START_LINE'.format(self.ikid), 0, 1)[0])
+
+    @property
+    def _detector_center_sample(self):
```
jessemapel: Same as focal length, why private?

ale/drivers/messenger_drivers.py
```diff
@@ -164,3 +182,59 @@ def starting_ephemeris_time(self):
...
+        return float(spice.gdpool('INS{}_BORESIGHT'.format(self.ikid), 0, 3)[0])
+
+
+    @property
+    def _detector_center_line(self):
```
jessemapel: Again why private?

ale/drivers/dawn_drivers.py
```diff
@@ -152,3 +123,27 @@ def starting_ephemeris_time(self):
             self._starting_ephemeris_time = spice.scs2e(self.spacecraft_id, sclock)
             self._starting_ephemeris_time += 193.0 / 1000.0
         return self._starting_ephemeris_time
+
+    @property
+    def _odtk(self):
```
SgStapleton: @jessemapel I imagine this needs to be public as well?

ale/drivers/messenger_drivers.py
SgStapleton: I quite literally rearranged code that already existed, so I did not think to check signatures...

SgStapleton: _focal_plane_temperature above is also private, but it is only used within focal_length. Do I need to make this public for testing purposes, or is it fine to simply test focal_length?

ale/drivers/dawn_drivers.py
jessemapel: This is not part of the base Driver interface so it can be private. If you look at #144 it accesses this via `_odtk`

ale/drivers/messenger_drivers.py
jessemapel: you can test private things in Python, the `_` for private is only a suggestion.

Issue #132, opened 2019-05-08
Write new Sensor Type mix-ins


jessemapel:
The line scan mix-in needs to provide:

- Stop time
- Line scan rate

The framer mix-in needs to provide:

- Center time

Issue #129, opened 2019-05-08
Write new SPICE mix-in


jessemapel:
This should be able to provide the following:

- Positions and velocities
- Position and velocity times
- ISIS NAIF Keywords
- Detector center
- Focal length
- Detector to focal plane transformations
- Target body radii
- Start time
- The transformation chain object from #128

PR #144, opened 2019-05-10
NaifSpice and sensor_type refactor

Kelvinrr:
Changed to address #132 and #129. Still, haven't found a clean way to do naif codes. Need something that can list all naif keys used in the Spice class.

### DIFF
--- a/ale/base/data_naif.py
+++ b/ale/base/data_naif.py
@@ -70,11 +70,11 @@ class NaifSpice():
 
     @property
     def focal2pixel_lines(self):
-        return list(spice.gdpool('INS{}_ITRANSL'.format(self.fikid), 0, 3))
+        return list(spice.gdpool('INS{}_ITRANSL'.format(self.ikid), 0, 3))
 
     @property
     def focal2pixel_samples(self):
-        return list(spice.gdpool('INS{}_ITRANSS'.format(self.fikid), 0, 3))
+        return list(spice.gdpool('INS{}_ITRANSS'.format(self.ikid), 0, 3))
 
     @property
     def _focal_length(self):
@@ -109,7 +109,7 @@ class NaifSpice():
     @property
     def _sun_position(self):
         sun_state, _ = spice.spkezr("SUN",
-                                     self.center_ephemeris_time,
+                                     self.center_time,
                                      self.reference_frame,
                                      'NONE',
                                      self.target_name)
@@ -118,8 +118,9 @@ class NaifSpice():
 
     @property
     def _sun_velocity(self):
+        center_time = self.ephemeris_time
         sun_state, lt = spice.spkezr("SUN",
-                                     self.center_ephemeris_time,
+                                     self.center_time,
                                      self.reference_frame,
                                      'NONE',
                                      self.target_name)
@@ -129,16 +130,15 @@ class NaifSpice():
     @property
     def _sensor_position(self):
         if not hasattr(self, '_position'):
+            ephem = self.ephemeris_time
             eph = []
-            current_et = self.starting_ephemeris_time
-            for i in range(self.number_of_ephemerides):
+            for time in ephem:
                 state, _ = spice.spkezr(self.spacecraft_name,
-                                        current_et,
+                                        time,
                                         self.reference_frame,
                                         'NONE',
                                         self.target_name,)
                 eph.append(state[:3])
-                current_et += getattr(self, "dt_ephemeris", 0)
             # By default, spice works in km
             self._position = [e * 1000 for e in eph]
         return self._position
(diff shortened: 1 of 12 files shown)


ale/base/data_naif.py
```diff
@@ -118,8 +118,9 @@ def _sun_position(self):
 
     @property
     def _sun_velocity(self):
+        center_time = self.ephemeris_time
```
jessemapel: Is this leftover?

ale/base/data_naif.py
```diff
@@ -109,7 +109,7 @@ def reference_frame(self):
     @property
     def _sun_position(self):
         sun_state, _ = spice.spkezr("SUN",
-                                     self.center_ephemeris_time,
+                                     self.center_time,
```
jessemapel: Do we need `center_time` as part of the base Driver interface?

ale/base/label_pds3.py
```diff
@@ -61,26 +61,19 @@ def target_name(self):
...
         return spice.bodn2c(self.label['TARGET_NAME'])
 
     @property
-    def starting_ephemeris_time(self):
-        if not hasattr(self, '_starting_ephemeris_time'):
-            sclock = self.label['SPACECRAFT_CLOCK_START_COUNT']
-            self._starting_ephemeris_time = spice.scs2e(self.spacecraft_id, sclock)
-        return self._starting_ephemeris_time
+    def _starting_ephemeris_time(self):
+        return self.label['SPACECRAFT_CLOCK_START_COUNT']
```
jessemapel: It makes sense to have the clock count be provided by the label and then the SPICE mix-in can convert it to an ephemeris time.

ale/base/label_pds3.py
```diff
@@ -61,26 +61,19 @@ def target_name(self):
...
 
     @property
-    def spacecraft_clock_stop_count(self):
-        sc = self.label.get('SPACECRAFT_CLOCK_STOP_COUNT', None)
-        if sc == 'N/A':
-            sc = None
-        return sc
+    def _ending_ephemeris_time(self):
+        return self.label['SPACECRAFT_CLOCK_STOP_COUNT']
```
jessemapel: Same as `_starting_ephemeris_time` this should stay `spacecraft_clock_stop_count` and then the SPICE mix-in can provide `_ending_ephemeris_time`

ale/base/label_pds3.py
jessemapel: With the current naming, the scs2e calls in SPICE don't make sense, it looks like converting ephemeris time to ephemeris time with a method that converts clock strings to ephemeris time.

ale/base/data_naif.py
Kelvinrr: Doesn't CSM use center time as acquisition time for framers?

ale/base/label_pds3.py
Kelvinrr: agree that at least better naming is needed. Not sure what though.

ale/base/data_naif.py
Kelvinrr: yeah, should be removed. Good catch.

ale/base/data_naif.py
jessemapel: It does, we have it in the Framer mix-in right now, should it be across all of the sensor_types? It looks like yes.

ale/base/data_naif.py
```diff
@@ -129,70 +129,68 @@ def _sun_velocity(self):
...
-            if fn == 'N/A':
-                fn = 0
-        else:
-            fn = 0
+    def _detector_center_sample(self):
+        return float(spice.gdpool('INS{}_BORESIGHT_SAMPLE'.format(self.ikid), 0, 1)[0])
 
-        return self.ikid - int(fn)
+    @property
+    def _detector_center_line(self):
```
jessemapel: Should be public, not private

ale/base/data_naif.py
```diff
@@ -129,70 +129,68 @@ def _sun_velocity(self):
...
 
     @property
-    def fikid(self):
-        if isinstance(self, Framer):
-            fn = self.filter_number
-            if fn == 'N/A':
-                fn = 0
-        else:
-            fn = 0
+    def _detector_center_sample(self):
```
jessemapel: Should be public, not private

ale/drivers/kaguya_drivers.py
```diff
@@ -69,21 +69,20 @@ def _tc_id(self):
         return spice.bods2c("LISM_{}".format(self.label.get("INSTRUMENT_ID")))
 
     @property
-    def ending_ephemeris_time(self):
-        if not hasattr(self, '_ending_ephemeris_time'):
-            # We need to get the corrected time
-            self._ending_ephemeris_time = self.label.get('CORRECTED_SC_CLOCK_STOP_COUNT').value
-            self._ending_ephemeris_time = spice.sct2e(self.spacecraft_id, self._ending_ephemeris_time)
-        return self._ending_ephemeris_time
+    def _ending_ephemeris_time(self):
```
jessemapel: Should be public, not private; so that it properly overrides.

ale/drivers/kaguya_drivers.py
```diff
@@ -69,21 +69,20 @@ def _tc_id(self):
...
+        return self.label.get('CORRECTED_SC_CLOCK_STOP_COUNT').value
 
+    @property
+    def stop_time(self):
+        return spice.sct2e(self.spacecraft_id, self._ending_ephemeris_time)
+
+    @property
+    def _starting_ephemeris_time(self):
```
jessemapel: Should be public, not privatel so that is properly overrides.

ale/base/label_pds3.py
jessemapel: `spacecraft_clock_start_count` seems like the logical choice to me.

Issue #127, opened 2019-05-08
Write new PDS3Label mix-in


jessemapel:
Needs to provide the following:

- Target body name
- Line and sample counts
- Sample & line summing
- Spacecraft clock count
- Exposure Duration

PR #154, opened 2019-05-11
Converted PDS3 label and added test

scsides:


scsides:
Still need to fix other tests I broke, but want comments.

ale/base/label_pds3.py
```diff
@@ -25,92 +15,202 @@ def label(self):
...
+          Returns
+        -------
+        : str
+          Full name of the instrument host
+        """
+        return self.label['INSTRUMENT_HOST_NAME']
+
 
     @property
     def start_time(self):
```
Kelvinrr: Just to be consistent, in my PR the time type is embedded in the name. Maybe change this to `utc_start_time`.

ale/base/label_pds3.py
```diff
@@ -25,92 +15,202 @@ def label(self):
...
+          Returns the start clock count string from the PDS3 label.
         """
-        return self.label['MISSION_NAME']
+        return self.label['SPACECRAFT_CLOCK_START_COUNT']
+
 
     @property
-    def detector_line_summing(self):
-        return self.label.get('SAMPLING_FACTOR', 1)
+    def spacecraft_clock_stop_count(self):
```
Kelvinrr: same as start time probably should rename it `clock_start_count`

ale/base/label_pds3.py
```diff
@@ -25,92 +15,202 @@ def label(self):
...
+          Returns
+        -------
+        : str
+          Full name of the instrument
+        """
+        return self.label['INSTRUMENT_NAME']
+
+
+    @property
+    def instrument_host_id(self):
```
Kelvinrr: Do we want to expose these hot keys?

ale/base/label_pds3.py
scsides: Both of @Kelvinrr comments bring up a point, we agreed to keep the names consistent with what PDS named them, but at what point, if any, do we change a name to help with clarity?

ale/base/label_pds3.py
scsides: @Kelvinrr please explain

Kelvinrr: basically, do we want to prefix it with an underscore or not. Without it, it is considered part of the public API.

scsides: No problem one way or the other. What reasons do we have to hide the keys?

ale/base/label_pds3.py
Kelvinrr: If we knew which will be consistent across all labels this would be an easier question to answer.

ale/base/label_pds3.py
Kelvinrr: Honestly, i'm not entirely sure. If this is specific to PDS3, then maybe hide the particular one?

ale/base/label_pds3.py
jessemapel: I think for right now we should go with PDS3 unless there's a good reason not to. We can adapt once we get PDS4 sorted.

ale/base/label_pds3.py
```diff
@@ -25,85 +15,230 @@ def label(self):
...
+#-------------------------------------------------------------
+#        def clock_start_count(self):
+#            return self.label['SPACECRAFT_CLOCK_START_COUNT']
+#
+#        @property
+#        def clock_stop_count(self):
+#            return self.label['SPACECRAFT_CLOCK_STOP_COUNT']
+#
+## Add this one in with similar code to to EXPOSURE_DURATION
+#-------------------------------------------------------------
```
jessemapel: Should this be removed?

tests/pytests/test_pds3_label.py
```diff
@@ -0,0 +1,119 @@
...
+DOWNTRACK_SUMMING = 2
+CROSSTRACK_SUMMING = 3
+
+END
+"""
+
+    def test_label(file):
+        return pvl.loads(label)
+
+    monkeypatch.setattr(pvl, 'load', test_label)
```
jessemapel: You can just remove this line. We don't need to override the PVL library because we can pass the label directly.

tests/pytests/test_pds3_label.py
```diff
@@ -0,0 +1,119 @@
+import pytest
+import pvl
+
+from datetime import datetime
+import ale
+from ale import base
+from ale.base.label_pds3 import Pds3Label
+
+@pytest.fixture
+def test_image_label(monkeypatch):
```
jessemapel: Same as below, we don't need to inherit from monkeypatch.

ale/base/label_pds3.py
scsides: Removed

ale/base/label_pds3.py
scsides: Going with PDS3 names for now, except for start/stop time are now utc_start/stop_time

jessemapel:
Fixes #134

jessemapel:
Fixes #127

PR #155, opened 2019-05-11
Adds support for Rotations

jessemapel:
This is the first part of the frame chain support. This mostly gives us a way to combine rotations between reference frames.

### DIFF
--- a/ale/rotation.py
+++ b/ale/rotation.py
@@ -2,6 +2,8 @@ from scipy.interpolate import interp1d
 from scipy.spatial.transform import Slerp
 from scipy.spatial.transform import Rotation
 
+import numpy as np
+
 class ConstantRotation:
     """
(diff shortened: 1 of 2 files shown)


ale/rotation.py
```diff
@@ -0,0 +1,180 @@
...
+    """
+
+    def __init__(self, quat, source, dest):
+        """
+        Construct a constant rotation
+
+        Parameters
+        ----------
+        quat : array
+               The quaternion representation of the rotation as a numpy array
```
scsides: The order of quaternions is a common point of confusion. It would be helpful to reference the order we expect the elements to be in. Maybe just a simple (x,y,z,w) or is it the other way?

jessemapel: I'm just using scipy's order which is constant-last (x, y, z, w). I'll add this to the docs.

Issue #137, opened 2019-05-08
Write new base Driver class


Kelvinrr:
The class should be an ABC that enforces the interface. Existing keys should stay more or less the same unless there is a good reason to change it, new information (like naif keywords) can have their names be set by whoever takes the issue. 

- Get the rotation between two frames for the time range of the image
- Get start and stop time for an image
- Get line and sample counts
- Get the sensor frame ID
- Get the target body frame ID
- Get the frame chain between two frames and the frame types for all returned frames
- Get sensor type
- Get sensor name
- Get sensor platform
- Get line scan rates
    - Three lists: Line starts, start time, scan rate
- Get sample summing
- Get line summing
- Get starting detector sample
- Get starting detector line
- Get ISIS NAIFKeywords
    - Return dictionary of keywords and values/lists of values
- Get sensor model version
- Get focal length
- Get the detector to focal plane transformations
- Get the target body radii
- Get the detector center
- Get the USGSCSM distortion model

jlaura:
Closed by #163

PR #157, opened 2019-05-20
Added abstract base driver class 

kberryUSGS:
Any driver test that doesn't implement everything in this abstract base class will fail. I didn't go through and update all of them in case there were substantial changes requested on the PR. 

This is for #137


ale/base/base.py
```diff
@@ -61,102 +48,249 @@ def to_dict(self):
...
+    def usgscsm_distortion_model(self):
+        """
+        Returns
+        -------
+        : dict
+          A dict containing the information about the distortion model for the usgscsm
+        """
+        pass
+
+#   def is_valid(self):
```
kberryUSGS: Should `is_valid` remain? It wasn't explicitly part of our design, so I commented it out.

tests/pytests/test_base.py
```diff
@@ -0,0 +1,199 @@
...
+    def sun_position(self):
+        return ([1], [1])
+
+    @property
+    def target_body_id(self):
+        return 1
+
+def test_me(test_label):
+    me = testclass(test_label)
+    assert me.image_lines == 1
```
kberryUSGS: I'm not sure if the ABC is worked as intended here: `image_lines` is defined in `IsisLabel`, which the test class inherits from, but I had to redefine `image_lines` in the test class in order to not get a 'not implemented' error. Anyone have any ideas about this?

jessemapel:
@scsides has some changes in #154 that may conflict with this, so I'm going to wait until that is merged to start on this.

jessemapel:
COntinued at #163

PR #159, opened 2019-05-21
Update README.md

jlaura:


README.md
```diff
@@ -1,4 +1,4 @@
-# Abstraction Layer or Ephemerides (ALE)
+# Abstraction Layer of Ephemerides (ALE)
```
jessemapel: It should be Abstraction Layer for Ephemerides :X

SgStapleton: I saw this a while ago, but was confused about what name we actually settled on...

PR #161, opened 2019-05-22
Implementation of _naif_keywords in NaifSpice()

SgStapleton:
Implementation of _naif_keywords in NaifSpice() along with tests for it and the properties I had to add for it. Fixes https://github.com/USGS-Astrogeology/ale/issues/158.

### DIFF
--- a/ale/base/data_naif.py
+++ b/ale/base/data_naif.py
@@ -69,6 +69,14 @@ class NaifSpice():
         return spice.bods2c(self.spacecraft_name)
 
     @property
+    def target_id(self):
+        return spice.bods2c(self.target_name)
+
+    @property
+    def body_frame_code(self):
+        return spice.gipool('BODY_FRAME_CODE', 0, 1)
+
+    @property
     def focal2pixel_lines(self):
         return list(spice.gdpool('INS{}_ITRANSL'.format(self.ikid), 0, 3))
 
@@ -79,6 +87,21 @@ class NaifSpice():
     @property
     def _focal_length(self):
         return float(spice.gdpool('INS{}_FOCAL_LENGTH'.format(self.ikid), 0, 1)[0])
+
+    @property
+    def pixel_size(self):
+        return spice.gdpool('INS{}_PIXEL_SIZE'.format(self.ikid), 0, 1)[0] * 0.001
+
+    @property
+    def _radii(self):
+        """
+        Returns
+        -------
+        : list<double>
+          Radius of all three axis of the target body
+        """
+        rad = spice.bodvrd(self.target_name, 'RADII', 3)
+        return rad[1]
 
     @property
     def _semimajor(self):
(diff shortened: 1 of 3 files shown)


SgStapleton:
I believe I may also need to add the "CLOCK_ET_<sclkCode>_<clockValue>_COMPUTED" key and value to the dictionary. @jessemapel Does this sound reasonable?

ale/base/data_naif.py
```diff
@@ -68,6 +68,14 @@ def ikid(self):
     def spacecraft_id(self):
         return spice.bods2c(self.spacecraft_name)
 
+    @property
+    def target_id(self):
+        return spice.bods2c(self.target_name)
+
+    @property
+    def body_frame_code(self):
+        return spice.gipool('BODY_FRAME_CODE', 0, 1)
```
jessemapel: This is not actually available as a kernel variable. `BODY_FRAME_CODE` is one of those values ISIS creates. You can get the reference frame for a body via [cidfrm](https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/cspice/cidfrm_c.html).

ale/base/data_naif.py
```diff
@@ -80,6 +88,21 @@ def focal2pixel_samples(self):
     def _focal_length(self):
         return float(spice.gdpool('INS{}_FOCAL_LENGTH'.format(self.ikid), 0, 1)[0])
 
+    @property
+    def pixel_size(self):
+        return spice.gdpool('INS{}_PIXEL_SIZE'.format(self.ikid), 0, 1)[0] * 0.001
+
+    @property
+    def _radii(self):
```
jessemapel: Why is this private?

jessemapel:
@SgStapleton Yes, the computed time would be good. It's a rather odd one, in that it's the computed ephemeris time as a hex encoded double.

ale/base/data_naif.py
SgStapleton: Ok, so I see where they are using cidfrm_c() in Spice.cpp in the init, but it looks like another path (the else block - which seems to be saying if you are not using naif...?) ends up calling getInteger() with 'BODY_FRAME_CODE' as it's key, which eventually makes use of the gipool_c() method, so I am a little confused here...

ale/base/data_naif.py
SgStapleton: Because I am still not fully aware when things need to be private or not in this project and was simply following the structure from _semimajor and _semiminor below... I can get this changed. Also, working with @paarongiroux on the cassini driver, it is looking like this method is returning what the to_usgscsm method is expecting to return from it's call to "target_body_radii". Should I rename this to "target_body_radii" here?

ale/base/data_naif.py
jessemapel: The "using naif" flag is for if you have SPICE data attached or not. So, anything that depends on that is for post spiceinit, I.E. post ALE. Also, getInteger doesn't always use gipool_c(). If the value is already stored in the NaifKeywords object it doesn't use gipool_c().

PR #162, opened 2019-05-23
Adds Transformation object

jessemapel:
Adds a tree based class for managing reference frames. This still needs to be integrated into the data classes, but that is going to be potentially complex, so I am splitting it out here.

### DIFF
--- a/ale/transformation.py
+++ b/ale/transformation.py
@@ -1,5 +1,7 @@
 import numpy as np
 
+from ale.rotation import ConstantRotation
+
 class FrameNode():
     """
     A single frame in a frame tree. This class is largely adapted from the Node
@@ -109,19 +111,22 @@ class FrameNode():
             raise RuntimeError('No common parent between nodes')
 
         first_path = parents_1[:parents_1.index(common_parent)+1]
-        second_path = reversed(parents_2[:parents_2.index(common_parent)])
+        second_path = parents_2[:parents_2.index(common_parent)][::-1]
 
         return first_path, second_path
 
     def rotation_to(self, other):
         """
-        Returns the rotation to another node.
+        Returns the rotation to another node. Returns the identity rotation
+        if the other node is this node.
 
         Parameters
         ----------
         other : FrameNode
                 The other node to find the rotation to.
         """
+        if other == self:
+            return ConstantRotation(np.array([0, 0, 0, 1]), self.id, other.id)
         forward_path, reverse_path = self.path_to(other)
         rotations = [node.rotation for node in forward_path[:-1]]
         rotations.extend([node.rotation.inverse() for node in reverse_path])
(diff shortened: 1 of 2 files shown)


ale/transformation.py
```diff
@@ -0,0 +1,136 @@
...
+        if self.parent is not None:
+            self.parent.children.remove(self)
+
+    @property
+    def parent(self):
+        """
+        The parent node of this node. Returns None if this is a root node.
+        """
+        if hasattr(self, '_parent'):
+            return self._parent
```
SgStapleton: This may just be my ignorance of Python, but it looks like you are trying to access "self._parent", when earlier parent is stored in "self.parent" (no underscore). How would this be working?

ale/transformation.py
```diff
@@ -0,0 +1,136 @@
...
+        """
+        Sets a new parent node. The child node is always responsible for
+        updating the parent node. So, this removes this node from the children
+        of the old parent and adds it to the children of the new parent.
+        """
+        if self.parent is not None:
+            self.parent.children.remove(self)
+
+        new_parent.children.append(self)
+        self._parent = new_parent
```
SgStapleton: Again with the inconsistent underscore... I must not be thinking of this right.

SgStapleton:
I just had those few questions, but everything else looked good to me.

ale/transformation.py
```diff
@@ -0,0 +1,136 @@
...
+                 The parent node in the frame chain
+        rotation : ConstantRotation or TimeDependentRotation
+                   The rotation from this frame to the frame of the parent node
+        """
+
+        if (parent is None) != (rotation is None):
+            raise TypeError("parent and rotation must both be entered or both be None.")
+
+        self.children = []
+        self.id = id
```
jlaura: `id` is reserved. `self.id` is fine.

ale/transformation.py
```diff
@@ -0,0 +1,136 @@
...
+            self.rotation = rotation
+
+    def __del__(self):
+        """
+        Custom deletor for a FrameNode. The child node is always responsible
+        for updating the parent nodes. So this removes this node from the
+        children of its parent node.
+        """
+        if self.parent is not None:
+            self.parent.children.remove(self)
```
jlaura: I am concerned about the call order here. This will only be called once the reference count goes to zero **and** the garbage collector fires. So, this is a soft guarantee that the child will alert the parent. This might be a case where a context is more appropriate, but I am not sure if a context is workable in this context.

ale/transformation.py
jlaura: Good catch, line 44 needs an update. It is working because of line 44.

SgStapleton:
Looks like the changes discussed in standup this morning have been implemented. Looks good to me!

PR #163, opened 2019-05-23
Update of Base

jessemapel:
This is a continuation of #157. This is just to get the Base driver class implemented for the concrete drivers to work with.


jessemapel:
This probably doesn't cover everything. I've recorded the proper names for this on the [wiki page](https://github.com/USGS-Astrogeology/ale/wiki/ALE-1.0-Architecture).

ale/drivers/cassini_drivers.py
```diff
@@ -115,3 +115,11 @@ def instrument_model_version(self):
           model version
         """
         return 1
+
+    @property
+    def detector_start_sample(Self):
```
kberryUSGS: Is there a reason why these selfs have an uppercase `S`?

ale/drivers/cassini_drivers.py
```diff
@@ -115,3 +115,11 @@ def instrument_model_version(self):
...
+        return 1
+
+    @property
+    def detector_start_line(Self):
```
kberryUSGS: Same as above.

jessemapel:
There probably also needs to be a pass to remove un-needed properties from existing mix-ins.

ale/drivers/cassini_drivers.py
```diff
@@ -13,7 +13,7 @@
 from ale.base.type_sensor import Framer
 
 
-class CassiniIssPds3LabelNaifSpiceDriver(Driver, Pds3Label, NaifSpice, Framer, RadialDistortion):
+class CassiniIssPds3LabelNaifSpiceDriver(Pds3Label, NaifSpice, Framer, RadialDistortion, Driver):
```
SgStapleton: I noticed you moved Driver to the end of the inheritance in this cassini driver, but you did not do so for the dawn driver. Is there a reason for the inconsistency?

jessemapel: Sorry to be getting back to this so late, I was getting strange test failures without this change.

paarongiroux: > Sorry to be getting back to this so late, I was getting strange test failures without this change.

So far I've only worked on dawn and kaguya but I got strange test failures as well. Switching the inheritance order fixed them in both cases.

jessemapel: I believe I was having issues with the abstract methods in Driver getting called instead of their implementations. Basically, the MRO meant that Driver's abstract methods were getting called instead of the mix-in implementations. Hence, the Driver had to be last to ensure the mix-in methods properly shadowed the abstract methods.

kberryUSGS: I also needed to move the Driver to end of the inheritance chain when I was initially working on the tests for base.py for the reasons @jessemapel listed.

jessemapel: Can someone update https://github.com/USGS-Astrogeology/ale/wiki/ALE-1.0-Architecture and https://github.com/USGS-Astrogeology/ale/wiki/Adding-New-Drivers to state that the Driver class MUST be last in the inheritance for concrete drivers?

jessemapel: Nevermind, I've updated the wiki about this under [adding new drivers](https://github.com/USGS-Astrogeology/ale/wiki/Adding-New-Drivers#multiple-inheritance-and-mro).

PR #166, opened 2019-05-28
updated dawn driver and tests

paarongiroux:


ale/drivers/dawn_drivers.py
```diff
@@ -89,7 +89,7 @@ def spacecraft_name(self):
         : str
           Spacecraft name
         """
-        return self.label['INSTRUMENT_HOST_NAME']
+        return 'DAWN'
```
SgStapleton: Was there a reason for us not hard-coding this in the first place...?

paarongiroux: I'm not sure. Another thing I realized about this property is that its overwriting spacecraft_name from the pds3label class, so we don't even need this. I was just following what I've seen in other drivers, but I'm not entirely sure if we should be hardcoding this or just letting the pds3Label handle it

ale/drivers/dawn_drivers.py
```diff
@@ -24,8 +24,8 @@ def instrument_id(self):
         : str
           instrument id
         """
-        instrument_id = self.label["INSTRUMENT_ID"]
-        filter_number = self.label["FILTER_NUMBER"]
+        instrument_id = super().instrument_id
```
SgStapleton: I just want to make sure that I am fully understanding how super() is being used here... I noticed in the cassini driver (which I was under the impression is "done"?) the line ```self.id_lookup[super().instrument_id]``` is being used to access the inherited instrument_id value. Was there a reason for the more complex access?

ale/drivers/dawn_drivers.py
scsides: We should try and use the pds3 driver when we can. INSTRUMENT_HOST_NAME should give the correct answer.

SgStapleton: Was the self.label (from the old code) not using the label from the pds3 mix-in before?

paarongiroux: Not sure if I understand the question, but the self.label reads from the label directly which from what I understand we want the mix in to be doing and not the concrete driver. Regarding the hard-coded return value, I've changed it to return the instrument_host_id property from the mix-in

ale/drivers/dawn_drivers.py
```diff
@@ -105,7 +106,7 @@ def target_name(self):
         : str
           target name
         """
-        target = self.label['TARGET_NAME']
+        target = super().target_name
```
SgStapleton: @jlaura @jessemapel Does this not seem like a better tool to use than managing private and public versions of certain methods to avoid the infinite recursion issue...? And if so, do we still need to implement private methods anywhere in ALE?

tests/pytests/test_dawn_drivers.py
```diff
@@ -0,0 +1,75 @@
...
+@pytest.fixture
+def driver():
+    return DawnFcPds3NaifSpiceDriver("")
+
+def test_instrument_id(driver):
+    with patch('ale.base.label_pds3.Pds3Label.instrument_id', new_callable=PropertyMock) as mock_instrument_id:
+        mock_instrument_id.return_value = 1
+        with patch('ale.base.label_pds3.Pds3Label.filter_number', new_callable=PropertyMock) as mock_filter_number:
+            mock_filter_number.return_value = 2
+            assert driver.instrument_id == 'DAWN_1_FILTER_2'
```
SgStapleton: A possibly cleaner solution to patch so we will not have to deal with nested ```with``` blocks:
```
@patch('ale.base.label_pds3.Pds3Label.instrument_id', 1)
@patch('ale.base.label_pds3.Pds3Label.filter_number', 2)
def test_instrument_id(driver):
    assert driver.instrument_id == 'DAWN_1_FILTER_2
```

SgStapleton: (I tested that this works in your test code as desired.)

tests/pytests/test_dawn_drivers.py
```diff
@@ -0,0 +1,75 @@
...
+
+def test_spacecraft_name(driver):
+    with patch('ale.base.label_pds3.Pds3Label.instrument_host_id', new_callable=PropertyMock) as mock_host_id:
+        mock_host_id.return_value = 'DAWN'
+        assert driver.spacecraft_name == 'DAWN'
```
SgStapleton: This works with single patches as well and keeps things a little cleaner:
```
@patch('ale.base.label_pds3.Pds3Label.instrument_host_id', 'DAWN')
def test_spacecraft_name(driver):
    assert driver.spacecraft_name == 'DAWN'
```

tests/pytests/test_dawn_drivers.py
```diff
@@ -0,0 +1,75 @@
...
+
+def test_target_name(driver):
+    with patch('ale.base.label_pds3.Pds3Label.target_name', new_callable=PropertyMock) as mock_target_name:
+        mock_target_name.return_value = '4 VESTA'
+        assert driver.target_name == 'VESTA'
+        mock_target_name.return_value = 'VESTA'
+        assert driver.target_name == 'VESTA'
```
SgStapleton: (This would not make a good candidate for the "cleaner" implementation above as the value would be reset.)

SgStapleton:
Possible reworking for patch implementations in the tests (@jlaura Does this look reasonable for what we found?), but other than that I would say this looks solid!

tests/pytests/test_dawn_drivers.py
jlaura: @SgStapleton Thanks for testing this. I think that this is a lot cleaner to read. Looks great! @paarongiroux Could you update these to this format? We should also get the testing docs updated for down stream people writing drivers.

tests/pytests/test_dawn_drivers.py
jlaura: My one comment here is that it might be a bit more readable as: `@patch('ale.base.label_pds3.Pds3Label.instrument_host_id', return_value='DAWN')`

Just because we know that the second arg is actually a kwarg. - unless I am misreading the docs and this fails. This is really minor.

jlaura:
@SgStapleton Looks great. I like the readability improvements. @paarongiroux Tests look good!

tests/pytests/test_dawn_drivers.py
paarongiroux: Okay, everything has been changed to this new format. @jlaura it doesn't seem like the return_value thing works. I will update the wiki page to include this later today

ale/drivers/dawn_drivers.py
```diff
@@ -157,3 +158,21 @@ def focal2pixel_lines(self):
         # Microns to mm
         pixel_size = spice.gdpool('INS{}_PIXEL_SIZE'.format(self.ikid), 0, 1)[0] * 0.001
         return [0.0, 0.0, 1/pixel_size]
+
+    def detector_start_line(self):
+        return 1
+
+    def detector_start_sample(self):
+        return 1
```
paarongiroux: Just remembered, I just set these to one as placeholder values since they had to be implemented and I was unsure of where to get this information. Where can I get the information for these two methods?

ale/drivers/dawn_drivers.py
```diff
@@ -157,3 +158,21 @@ def focal2pixel_lines(self):
...
+    def sensor_model_version(self):
+        """
+        Returns instrument model version
+
+        Returns
+        -------
+        : int
+          model version
+        """
+        return 2
```
SgStapleton: Was this a placeholder as well?

paarongiroux: No, I got this value from the dawnfc camera.plugin file in the isis repo

ale/drivers/dawn_drivers.py
SgStapleton: It doesn't seem to appear that anything like these exist in the Dawn labels, so I am not entirely sure this needs to be implemented (other than the abstract base class forcing it). @jessemapel Any thoughts?

SgStapleton: Unless it is the ```VirIrStartXPosition``` and ```VirIrStartYPosition``` (on the VIR sensor, anyway...)

SgStapleton: Or possibly the ```FirstLine``` and ```FirstLineSample``` of a DawnFC cube...?

paarongiroux: Wait I just realized when looking in the kaguya driver which I'm currently working on that it has a start_detector_sample which I would assume is the same thing. It gets the information from FIRST_PIXEL_NUMBER in the pds3Label so that might be it

paarongiroux: Doesn't seem like the dawn labels have this so nevermind I guess?

SgStapleton: I'd say grab it. The method needs to be implemented either way, right? May as well have the method return something useful.

PR #171, opened 2019-05-30
Initial stab at PDS3 NAIF CTX

jlaura:
This is an initial stab at a NAIF, PDS3 CTX driver. No modifications for ISIS drivers have been done.

### DIFF
--- a/ale/drivers/mro_drivers.py
+++ b/ale/drivers/mro_drivers.py
@@ -151,7 +151,7 @@ class MroCtxPds3LabelNaifSpiceDriver(Pds3Label, NaifSpice, LineScanner, RadialDi
 
     @property
     def detector_start_line(self):
-        return 0
+        return 1
 
     @property
     def detector_start_sample(self):
--- a/ale/drivers/mro_drivers.py
+++ b/ale/drivers/mro_drivers.py
@@ -90,7 +90,7 @@ class MroCtxIsisLabelNaifSpiceDriver(IsisLabel, Driver, NaifSpice, LineScanner, 
         return "MRO"
 
 
-class MroCtxPds3LabelNaifSpiceDriver(Pds3Label, Driver, NaifSpice, LineScanner, RadialDistortion):
+class MroCtxPds3LabelNaifSpiceDriver(Pds3Label, NaifSpice, LineScanner, RadialDistortion, Driver):
     """
     Driver for reading CTX PDS3 labels. Requires a Spice mixin to acquire addtional
     ephemeris and instrument data located exclusively in spice kernels.
@@ -111,7 +111,7 @@ class MroCtxPds3LabelNaifSpiceDriver(Pds3Label, Driver, NaifSpice, LineScanner, 
         if not hasattr(self, '_metakernel'):
             self._metakernel = None
             for mk in mks:
-                if str(self.start_time.year) in os.path.basename(mk):
+                if str(self.utc_start_time.year) in os.path.basename(mk):
                     self._metakernel = mk
         return self._metakernel
 
@@ -131,7 +131,7 @@ class MroCtxPds3LabelNaifSpiceDriver(Pds3Label, Driver, NaifSpice, LineScanner, 
             'CONTEXT CAMERA':'MRO_CTX'
         }
 
-        return id_lookup[self.label['INSTRUMENT_NAME']]
+        return id_lookup[super().instrument_id]
 
     @property
     def spacecraft_name(self):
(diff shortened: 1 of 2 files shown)


ale/drivers/mro_drivers.py
```diff
@@ -147,4 +147,16 @@ def spacecraft_name(self):
         name_lookup = {
             'MARS_RECONNAISSANCE_ORBITER': 'MRO'
         }
-        return name_lookup[self.label['SPACECRAFT_NAME']]
+        return name_lookup[super().spacecraft_name]
+
+    @property
+    def detector_start_line(self):
+        return 0
```
jlaura: I think...I didn't see this in the IK or the label.

ale/drivers/mro_drivers.py
```diff
@@ -147,4 +147,16 @@ def spacecraft_name(self):
...
+
+    @property
+    def detector_start_sample(self):
+        return self.label.get('SAMPLE_FIRST_PIXEL', 0)
```
jlaura: I dislike this, but am not sure if we have another model for doing this? We need to value from a PDS label that is specific to this mission. I believe the decision was that the instrument/data specific driver should therefore be able to pull the data directly?

ale/drivers/mro_drivers.py
```diff
@@ -147,4 +147,16 @@ def spacecraft_name(self):
...
+    
+    @property
+    def sensor_model_version(self):
+        return 1
```
jlaura: From the ISIS CTX plugin file.  Thanks @paarongiroux for describing where you got this for dawn!

tests/pytests/test_mro_drivers.py
```diff
@@ -19,58 +21,25 @@
...
+    assert driver.detector_start_line == 0
 
-def test_ctx_creation(mroctx_label):
-    #with MroCtxPds3LabelNaifSpiceDriver(mroctx_label) as m:
-    #    d = m.to_dict()
-    #assert isinstance(d, dict)
+def test_detector_start_sample(driver):
+    # I am not sure how to accomplish this with a fixture and
+    # a decorator. Therefore, using a context
+    with patch.dict(driver.label, SAMPLE_FIRST_PIXEL=0) as f:
```
jlaura: Comment says it all. Would love to change to decorator style, but I don't know that I can patch a non-instantiated instances of the class being tested...

ale/drivers/mro_drivers.py
scsides: Yes that was the decision, it can pull it directly. I see this one as, on the fence. The key from pds3 is standard, so adding a getter for it in the label_pds3 would be reasonable, but few pds3 files use it, so having the specific driver get it works also.

scsides: Also, I'm wondering if we need to take a closer look at other docs (instrument SIS, ISIS3 code) as we create these drivers. With all the loose standards we are working with, picking a key from pds, isis, or anywhere because it sounds correct could cause problems/bugs down the line.

ale/drivers/mro_drivers.py
scsides: This is an image specific thing. Should only show up in the label.

ale/drivers/mro_drivers.py
scsides: From our discussion: If this is not set, the isis code as a built in default of 1 (one).

ale/drivers/mro_drivers.py
jessemapel: All of the keys selected should be based on documentation like the SIS when we write new instrument drivers. For porting these old drivers to the new structure, we can assume that the old drivers were written correctly.

ale/drivers/mro_drivers.py
jlaura: That code needs to be updated though, so this should probably stay and then we can get the Base updated in another PR.

ale/drivers/mro_drivers.py
jlaura: Can someone link to the CTX SIS?

ale/drivers/mro_drivers.py
jlaura: :+1: I am not seeing this in any CTX labels, but I am viewing a super tiny sample. Do we have an example where it does so that I can get the correct label keyword? Or I'll get this from the SIS (see my other comment below).

ale/drivers/mro_drivers.py
jessemapel: It should be located in the top level of the PDS archive you're pulling image data from.

ale/drivers/mro_drivers.py
jessemapel: I don't think there are any CTX images where this isn't 1. The default to start at the top of the detector is 1.

PR #180, opened 2019-05-31
Updated Messenger Drivers and tests

paarongiroux:


ale/drivers/messenger_drivers.py
```diff
@@ -124,8 +123,27 @@ def detector_center_sample(self):
...
-class MessengerMdisIsisLabelNaifSpiceDriver(IsisLabel, Driver, NaifSpice, Framer):
+    @property
+    def usgscsm_distortion_model(self):
+        return {
+            "transverse": {
+                "x" : self.odtx,
+                "y" : self.odty
+                }
+            }
+
```
paarongiroux: Copied this function from the IsisLabelNaifSpiceDriver based on the assumption that the distortion model wouldn't change depending on whether or not we're reading from an IsisLabel or a Pds3Label

ale/drivers/messenger_drivers.py
```diff
@@ -172,10 +186,10 @@ def instrument_id(self):
         : str
           instrument id
         """
-        return self.id_lookup[self.label['IsisCube']['Instrument']['InstrumentId']]
+        return self.id_lookup[super().instrument_id]
```
scsides: This id_lookup appears in both the isis label and pds label drivers. Suggest moving it above the top "class" so it can be referenced by both without having two identical copies, and make id_lookup all caps. Thanks @jlaura

paarongiroux: Does my latest commit work? Now that I'm thinking about it kind of seems like it should be declared as a function instead of a variable if its going to be outside of the classes right?

ale/drivers/messenger_drivers.py
```diff
@@ -172,10 +186,10 @@ def instrument_id(self):
...
 
     @property
-    def _focal_plane_tempature(self):
+    def _focal_plane_temperature(self):
```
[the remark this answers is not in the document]
paarongiroux: good point, got rid of this property.

PR #182, opened 2019-06-03
updated to_usgscsm formatter to use frame chain and changed tests accordingly

paarongiroux:


ale/formatters/usgscsm_formatter.py
```diff
@@ -1,6 +1,7 @@
 import json
 from ale.base.type_sensor import LineScanner, Framer
 from ale.encoders import NumpyEncoder
+from ale.rotation import ConstantRotation, TimeDependentRotation
```
jessemapel: I think you can remove these import methods.

PR #186, opened 2019-06-05
Clean up to get CTX driver working

jessemapel:
Collection of misc changes to get `to_usgscsm` and `to_isis` working with the CTX PDS3 driver.

### DIFF
--- a/ale/base/type_sensor.py
+++ b/ale/base/type_sensor.py
@@ -28,7 +28,7 @@ class LineScanner():
 
     @property
     def ephemeris_time(self):
-        return np.linspace(self.ephemeris_start_time,  self.ephemeris_stop_time, self.image_lines)
+        return np.linspace(self.ephemeris_start_time,  self.ephemeris_stop_time, self.image_lines / 64)
 
 class Framer():
     @property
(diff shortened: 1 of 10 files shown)


ale/drivers/mro_drivers.py
```diff
@@ -196,3 +197,7 @@ def detector_start_sample(self):
     @property
     def sensor_model_version(self):
         return 1
+
+    @property
+    def exposure_duration(self):
+        return self.line_exposure_duration
```
paarongiroux: Would it be worth while to only have an exposure_duration property in the label mixins that checks whether the label has line exposure duration or just exposure duration and returns whichever one it has? @Kelvinrr brought this up to me a while ago, seems like it might be nice to make things a little simpler.

ale/base/type_sensor.py
```diff
@@ -24,7 +24,7 @@ def line_scan_rate(self):
           2d list of scan rates in the form: [[start_line, line_time, exposure_duration], ...]
         """
         t0_ephemeris = self.ephemeris_start_time - self.center_ephemeris_time
-        return [[float(self.starting_detector_line)], [t0_ephemeris], [self._line_exposure_duration]]
+        return [[0.5], [t0_ephemeris], [self.exposure_duration]]
```
paarongiroux: Why is the starting detector line hardcoded to 0.5? doesn't it change based on the instrument?

jessemapel: This shouldn't have been the starting detector line, this should be the ISIS starting **image** line. Hence it's 0.5

PR #196, opened 2019-07-09
Added lro lroc nac notebook.

kaitlyndlee:


kberryUSGS:
:+1: This looks great. I'm not sure whether this information belongs in the notebook itself or elsewhere, but it would be useful to add information about:

* What location this notebook needs to be run in (i.e. Does the config file path for the LRO mks point to a scratch area that is only available on a specific machine?) 
** I'm not sure if this would be a good addition to this specific notebook or not, but it might be good to include an example of changing the config path to the mks somewhere in our documentation. 

* What requirements (if any) are there for the environment this notebook should be run in. Just ale and ale's dependencies?

kberryUSGS:
Also, this may be personal preference, but I'd use the spelling "furnish" to refer to the act of loading spice kernels and `furnsh` to refer to the specific spicelib utility which performs this action.

thareUSGS:
I agree with Kristen's comments. Can those be addressed before I merge. For "config should be updated to point to that directory as requirements", is there a link which describes this for us that we can just point to?

thareUSGS:
I assume in the first block the import errors are killing the builds? Can those be cleaned up.

---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-2-b8be604e48fe> in <module>

kaitlyndlee:
Oops I forgot to save the notebook before exiting. Should be good now.

PR #199, opened 2019-07-10
Small Dawn Fixes

acpaquette:
Fixed some imports with dawn to get things working


ale/drivers/dawn_drivers.py
```diff
@@ -45,6 +47,7 @@ def metakernel(self):
         metakernel_dir = config.dawn
         mks = sorted(glob(os.path.join(metakernel_dir,'*.tm')))
         if not hasattr(self, '_metakernel'):
+            self._metakernel = None
             for mk in mks:
                 if str(self.start_time.year) in os.path.basename(mk):
```
paarongiroux: should be utc_start_time

ale/drivers/dawn_drivers.py
```diff
@@ -223,3 +227,38 @@ def sensor_model_version(self):
...
+        Returns
+        -------
+        : float
+          center detector line
+        """
+        return float(spice.gdpool('INS{}_BORESIGHT'.format(self.ikid), 0, 3)[1])
+
+
+    @property
+    def detector_start_sample(self):
```
acpaquette: I'm not sure if detector_start_sample, or detector_start_line is correct. Was the closest thing I could find in the dawn ik kernels that looked similar to messenger.

acpaquette:
This PR is dependent on #201 for the rotation hotfix

ale/drivers/dawn_drivers.py
acpaquette: I think they actually are just both 1?

PR #200, opened 2019-07-10
updated version number 

Kelvinrr:
code to make sure `__version__` and setup,py version are the same.


acpaquette:
Last check passed and this was just a config change so this looks fine

PR #201, opened 2019-07-10
Cassini ISS - rotation fix?

kberryUSGS:
I updated `__mul__` in `TimeDependentRotation` to not even try to interpolate if self and other have one time and it's the same time, to get Cassini working. I'm not sure that this was the right solution and I seem to recall discussion about doing something else here? Reviewers, if you remember or have better ideas let me know.

Also, the mro config location update that came along for the ride is unrelated, but useful.

### DIFF
--- a/ale/rotation.py
+++ b/ale/rotation.py
@@ -164,7 +164,6 @@ class TimeDependentRotation:
         The destination frame of the right rotation (other) and the source
         frame of the left rotation (self) must be the same. I.E. if A and B are
         rotations, then for A*B to be valid, A.source must equal B.dest. 
-        # typo? A.dest = B.source? 
 
         If the other rotation is a time dependent rotation, then the time range
         for the resultant rotation will be the time covered by both rotations.
@@ -182,7 +181,7 @@ class TimeDependentRotation:
         elif isinstance(other, TimeDependentRotation):
             # if self and other each have the same time and one rotation, don't interpolate.
             if (self.times.size == 1) and (other.times.size == 1) and (self.times == other.times):
-                    return TimeDependentRotation((self._rots * other._rots).as_quat(), self.times, other.source, self.dest)
+                return TimeDependentRotation((self._rots * other._rots).as_quat(), self.times, other.source, self.dest)
             merged_times = np.union1d(np.asarray(self.times), np.asarray(other.times))
             # we cannot extrapolate so clip to the time range both cover
             first_time = max(min(self.times), min(other.times))
--- a/ale/rotation.py
+++ b/ale/rotation.py
@@ -179,6 +179,10 @@ class TimeDependentRotation:
         if isinstance(other, ConstantRotation):
             return TimeDependentRotation((self._rots * other._rot).as_quat(), self.times, other.source, self.dest)
         elif isinstance(other, TimeDependentRotation):
+            # if self and other each have the same time and one rotation, don't interpolate.
+            if (self.times.size == 1) and (other.times.size == 1):
+                if (self.times == other.times):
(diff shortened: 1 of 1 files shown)


ale/config.py
```diff
@@ -6,7 +6,7 @@
 spice_root = "/data/spice/"
 cassini = '/usgs/cpkgs/isis3/data/cassini/kernels/mk/' # Cassini ISS
 mdis = '/data/spice/mess-e_v_h-spice-6-v1.0/messsp_1000/extras/mk' # Messenger
-mro = '/data/spice/mro-m-spice-6-v1.0/mrosp_1000/extras/mk' # Mars Reconnaissance Orbiter
+mro = '/usgs/cpkgs/isis3/data/mro/kernels/mk' # Mars Reconnaissance Orbiter
```
kaitlyndlee: config.py doesn't exists anymore. Pull down the latest changes and update config.yaml.

acpaquette:
Might want to do a rebase on this PR. Looks like there are a good chunk of already merged or unnecessary commits in this

kberryUSGS:
Yeah, good point. Will do!

kberryUSGS:
Oops, that was _not_ better. Sorry, fixing.

kberryUSGS:
@acpaquette The commit history is fixed now.

acpaquette:
@kberryUSGS I think this fix is reasonable but I would like to fix this problem further up the call stack. 

If you checkout rotation_to in transformation.py the first line is trying to determine if the self and other are the same. Ideally, this would do the comparison you are checking for in the __mul__ function and just write out a constant rotation from rotation_to. 

More over if there really is only one rotation, why is there more than one frame node? Overriding the __eq__ function with something that does a more logical comparison between frame nodes and fixing the frame chain generation (if there really is something to fix there) would be a more long term fix for this issue. I'm looking into some of this right now and we could get together in a Zoom room to talk about it a bit more.

kberryUSGS:
@acpaquette Just saw this -- if you're still available / interested in getting on Zoom, could you gchat or gitter me? Self and other in the case I was looking at are different rotations (so different frames) but at the same time.

Kelvinrr:
What is the state if this PR? Were you able to touch base with @acpaquette? We should get this PRed since it's blocking a few people.

kberryUSGS:
@Kelvinrr We did talk for a while last night. The plan was to merge this and then if it needs to be changed later, it can be changed later. (And @acpaquette's comment above turned out to be mostly unrelated to this particular change.

ale/rotation.py
```diff
@@ -179,6 +179,10 @@ def __mul__(self, other):
         if isinstance(other, ConstantRotation):
             return TimeDependentRotation((self._rots * other._rot).as_quat(), self.times, other.source, self.dest)
         elif isinstance(other, TimeDependentRotation):
+            # if self and other each have the same time and one rotation, don't interpolate.
+            if (self.times.size == 1) and (other.times.size == 1):
+                if (self.times == other.times):
```
jessemapel: I would just make this a single if statement.

ale/rotation.py
```diff
@@ -179,6 +179,10 @@ def __mul__(self, other):
...
+                    return TimeDependentRotation((self._rots * other._rots).as_quat(), self.times, other.source, other.dest)
```
jessemapel: This should be `other.source, self.dest)`. If other rotates a to b and this rotates b to c. Then, self * other rotates  a to c.

kberryUSGS: Thank you for catching this. Totally a typo.

PR #204, opened 2019-07-11
MDIS ISD notebook **Don't merge until #201 gets merged

paarongiroux:


Kelvinrr:
#201 was merged, so this should be good.

PR #206, opened 2019-07-13
General ISD Generation Fixes + Dawn Notebook

acpaquette:
Many of these changes come directly from inconsistent translation from old ale to new ale. The main problem stems from how we are calculating center ephemeris time.

In old ale this was done using the exposure duration, in new ale this was mostly done through an average between the starting and ending epheremis times.


acpaquette:
Seems like the c++ side of things is failing to import ale correctly :(

ale/drivers/__init__.py
```diff
@@ -51,19 +51,18 @@ def load(label):
...
+        except Exception as e:
+            import traceback
+            print(f'Failed: {e}\n')
+            traceback.print_exc()
     raise Exception('No Such Driver for Label')
 
 
-def loads(label):
-    res = load(label)
+def loads(label, formater):
```
jlaura: Can this go in as formatter = formatter? This follows the json.loads API better

ale/drivers/__init__.py
```diff
@@ -51,19 +51,18 @@ def load(label):
...
-                import traceback
-                print("Driver Failed:", e)
-                traceback.print_exc()
+        print(f'Trying {name}')
+        try:
+            res = driver(label)
+            with res as driver:
+                return formater(driver)
+        except Exception as e:
+            import traceback
```
jlaura: Why not just import above?

ale/drivers/dawn_drivers.py
```diff
@@ -127,11 +127,11 @@ def ephemeris_start_time(self):
...
         account for the CCD being discharged or cleared.
         """
-        if not hasattr(self, '_starting_ephemeris_time'):
+        if not hasattr(self, '_ephemeris_start_time'):
             sclock = self.spacecraft_clock_start_count
-            self._starting_ephemeris_time = spice.scs2e(self.spacecraft_id, sclock)
-            self._starting_ephemeris_time += 193.0 / 1000.0
-        return self._starting_ephemeris_time
+            self._ephemeris_start_time = spice.scs2e(self.spacecraft_id, sclock)
+            self._ephemeris_start_time += 193.0 / 1000.0
```
jlaura: Can we get a doc string about where this number is coming from?

ale/formatters/usgscsm_formatter.py
```diff
@@ -23,7 +23,7 @@ def to_usgscsm(driver):
     body_radii = driver.target_body_radii
     isd_data['radii'] = {
         'semimajor' : body_radii[0],
-        'semiminor' : body_radii[1],
+        'semiminor' : body_radii[2],
```
jlaura: ??? Why thus change?

ale/drivers/__init__.py
acpaquette: This was in the function originally, I can move it with the others though

ale/drivers/dawn_drivers.py
acpaquette: Should already be in the doc string

ale/formatters/usgscsm_formatter.py
acpaquette: This is due to what target_body_radii returns, specifically the spice call that is used to get the radii from the kernels. Its a set of three values the first being the maximum radius at the equator, the second being the minimum radius at the equator and the last being the polar radius.

If we want the semi major (equator) and semi minor (polar) then we need to get the first and third elements.

jessemapel:
I agree with moving center_ephemeris_time out of the data_naif mix-in, but I don't think we should change it to use the exposure durations. The end_ephemeris_time should be changed to use the exposure duration to properly compute the end time and then the center_ephemeris_time will come along for the ride.

ale/base/type_sensor.py
```diff
@@ -49,6 +49,21 @@ def ephemeris_time(self):
...
+        duration. Expects ephemeris start time and line exposure duration to be
+        defined. These should be double precision numbers containing the ephemeris start and
+        line exposure duration of the image.
+
+        Returns
+        -------
+        : double
+          Center ephemeris time for an image
+        """
+        return self.ephemeris_start_time + self.line_exposure_duration / 2
```
jessemapel: For line scan sensors, the end time should be `ephemeris_start_time + lines * line_exposure_duration`

PR #207, opened 2019-07-15
added data area change script

Kelvinrr:
The simple brain-dead script that changes the data area variable in metakernels to point to an input directory. Pretty much assumes all instances of `PATH_VALUES =` are the kernel data area.


acpaquette:
Seems fine to me, only suggestion I can think of is adding the script to the setup.py so it could just be called like a command line utility

Kelvinrr:
It seems this looks good so I'm just gonna merge. If we move to make more runnable scripts we can change this into making sure they get properly installed at that point can maybe use click for creating them.

PR #214, opened 2019-07-17
updates data_isis to return positions in the correct frame

jessemapel:
This also includes code from #213 because I need it to do the rotations. The logic for the sensor position will be very similar.

We still need to update the usgscsm formatter to re-interpolate everything though.


jessemapel:
This also needs some testing on real data, but I don't have time just right now.

jessemapel:
This is "functional" but I want to clean up the table parsing code so that it's 1 step to read the binary data and table label into a dictionary and then 1 step to convert that into rotation objects.

jessemapel:
This is finally ready for review now

ale/base/data_isis.py
```diff
@@ -36,187 +37,144 @@ def read_table_data(table_label, cube):
...
+                if len(field_data) == 1:
+                    field_data = field_data[0]
+
             results[field['Name']].append(field_data)
-            offset += field_size(field)
-    return results
+            offset += data_sizes[field['Type']] * field['Size']
+
+    # Parse the keywords from the label
+    results.update({key : value for key, value in table_label.items() if not isinstance(value, pvl._collections.PVLGroup)})
```
acpaquette: Seems like this function never returns. Not sure if this is the intended behavior but it seems incorrect

ale/base/data_isis.py
```diff
@@ -36,187 +37,144 @@ def read_table_data(table_label, cube):
...
+        time_scale = table['J2000SVY'][-1]
+        scaled_times = (ephemeris_times - base_time) / time_scale
+        coeffs = np.array([table['J2000SVX'][:-1],
+                           table['J2000SVY'][:-1],
+                           table['J2000SVZ'][:-1]]).T
+        positions = polyval(scaled_times, coeffs).T
+        rotated_pos = rotation.apply_at(positions, ephemeris_times)
+    else:
+        raise ValueError('No positions are available in the input table.')
+    return rotated_pos, ephemeris_times
```
acpaquette: The change to this function seems pretty significant, has the rotation table already been parsed and we just need to actually rotate it now?

ale/base/data_isis.py
```diff
@@ -259,24 +217,6 @@ class IsisSpice():
...
-        PVLModule :
-            Dict-like object with PVL keys
-        """
-        if not hasattr(self, "_label"):
-            try:
-                self._label = pvl.load(self.file)
-            except:
-                raise ValueError("{} is not a valid label".format(self.file))
-        return self._label
-
```
acpaquette: Was this duplicate code that was removed?

ale/base/data_isis.py
```diff
@@ -571,6 +507,25 @@ def target_body_radii(self):
...
+        Returns
+        -------
+        FrameNode
+            The root node of the frame tree. This will always be the J2000 reference frame.
+        """
+        if not hasattr(self, '_frame_chain'):
+            self._frame_chain = FrameChain.from_isis_tables(
+                    inst_pointing = self.inst_pointing_table,
+                    body_orientation = self.body_orientation_table)
+        return self._frame_chain
```
acpaquette: :+1:

ale/base/data_isis.py
jessemapel: This is an entirely new function. Git just though it was changed. I completely removed the parse_rotation_table and parse_position_table methods. Now, there's just one parse_table method that copies over all of the PVL keywords from the table label and converts the binary fields into lists.

ale/base/data_isis.py
jessemapel: Yes

Issue #77, opened 2018-12-21
Add light-time support to Ale.


kberryUSGS:
ISIS optionally allows light-time correction (adjustment for the time it takes the light to travel from the target to the detector) and more. (see `abcorr` under: ftp://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/cspice/spkez_c.html.) Does it make sense to support anything like this in Ale?

jlaura:
I was under the impression that we really need to support this inside of the sensor and not on the ephemeris information coming in, but I would love to learn more about where it is most appropriate to do and why!

kberryUSGS:
@jlaura Dealing with this in the sensor sounds totally reasonable to me. I'm still trying to get the "division of responsibilities" down, as I keep lumping in ale's functionalities in with "spice usually takes care of this" but that's not really correct.

kberryUSGS:
Talked to @jessemapel and this does need to be added, so changed ticket title.

jessemapel:
To add some more context. To hand light time correctly, we should be accounting for it in the sensor model, but right now ISIS is not. Instead, ISIS has SPICE handle it during the spkezr calls in spiceinit. In order to support current functionality, we should be matching those spkezr calls. ISIS handles this as a kernel variable in the IAKs (`INS<ik code>_LIGHTTIME_CORRECTION`). We can handle this in ALE by adding a property to data_naif that is passed into spkezr for light time correction. Then, concrete drivers that use data_isis can set it as they please.

jessemapel:
In short, we should be using the `LT+S` adjustments from [SPICE](https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/req/abcorr.html) whenever we compute where the sensor is relative to the target in USGSCSM. This way, we can account for things like highly oblique imagery where the light time to the surface changes significantly across the image.

jessemapel:
Based on https://github.com/USGS-Astrogeology/ISIS3/blob/dev/isis/src/base/objs/Spice/LightTimeCorrectionState.cpp, it looks like ISIS defaults to `LT+S`

PR #215, opened 2019-07-17
Cassini update

Kelvinrr:
Seems like focal to pixel needed to be negative. Now it's subpixel in the line direction and 3-4 pixels off in the sample direction. see https://github.com/USGS-Astrogeology/knoten/pull/30


kberryUSGS:
I think just update the test to use the new, correct values.

ale/drivers/cassini_drivers.py
```diff
@@ -96,7 +95,7 @@ def focal2pixel_samples(self):
         """
         # Microns to mm
         pixel_size = spice.gdpool('INS{}_PIXEL_SIZE'.format(self.ikid), 0, 1)[0] * 0.001
-        return [0.0, 1/pixel_size, 0.0]
+        return [0.0, -1/pixel_size, 0.0]
```
scsides: Correct me if I'm wrong, these control the transform to/from detector position and focal plane x,y. Changing the sign on these coefficients cause that element of the transform to be subtracted instead of added. This is opposite to what the ISIS focal plane map is doing. Can this be explained?

Kelvinrr:
For this to be merged, I need to change the CCD flip back and add a 180-degree rotation for sensor orientation. I'll wait until #216 gets merged to make that update. This should be pretty accurate compared to ISIS sans light-time correction since CSM does not do any light-time correction at this time.

jessemapel:
I confused by "CSM does not do any light-time correction at this time." We can always turn it on by adding a property for it and passing that into the spkezr calls in data_isis.

jessemapel:
See #77 for some light time discussion.

Kelvinrr:
This should be good to review, see https://github.com/USGS-Astrogeology/knoten/pull/32 for results.

ale/drivers/cassini_drivers.py
Kelvinrr: We already covered this in person, but just to document things here: we needed to do an extra 180 rotation to the final sensor orientation. The negatives here were basically faking that rotation.

ale/drivers/cassini_drivers.py
```diff
@@ -175,3 +232,47 @@ def sensor_model_version(self):
...
+    def _original_naif_sensor_frame_id(self):
+        """
+        Returns the Naif ID code for the target reference frame
+        Expects the target_id to be defined. This must be the integer Naif ID code
+        for the target body.
+
+        Returns
+        -------
+        : int
+          Naif ID code for the target frame
```
jessemapel: These docs look incorrect

ale/drivers/cassini_drivers.py
```diff
@@ -175,3 +232,47 @@ def sensor_model_version(self):
...
         """
         return 1
+
+    @property
+    def frame_chain(self):
+        if not hasattr(self, '_frame_chain'):
+            if self.instrument_id == 'CASSINI_ISS_NAC':
+                self._frame_chain = FrameChain.from_spice(frame_changes = [(1, self._original_naif_sensor_frame_id), (1, self.target_frame_id)], ephemeris_time=self.ephemeris_time)
+                rot_180 = Rotation.from_euler('z', 180, degrees=True)
+                self._frame_chain.add_edge(self._original_naif_sensor_frame_id, self.sensor_frame_id, ConstantRotation(rot_180.as_quat(), self._original_naif_sensor_frame_id, self.sensor_frame_id))
```
jessemapel: Is it worth making the Rotation classes constructable from a scipy rotation?

ale/drivers/cassini_drivers.py
Kelvinrr: all docs should be updated now for new methods

ale/drivers/cassini_drivers.py
Kelvinrr: we definitely need to make this cleaner, seems better to do it in a different PR. Adam makes it seem like it's not as simple as just constructing ALE Rotations from Scipy Rotations. Kinda want to get these rotations as close to scipy rotations as possible (i.e. not having our own Rotation at all).

PR #216, opened 2019-07-19
Frame Chain Update

acpaquette:
Initial update to how the frame chain is computed. 

Removes the majority of  manual logic for creating chains and can now be done through a list of tuples (edges) of spice/NAIF related codes. This should probably change but its a start and gets the idea across.

Oh, and stuff is still going to be a little broken. I need to figure out how testing the formatters is going to work with this change.


ale/config.yml
```diff
@@ -1,8 +1,8 @@
-spice_root: "/data/spice/"
+spice_root: '/data/spice/'
 cassini: '/usgs/cpkgs/isis3/data/cassini/kernels/mk/' # Cassini ISS
 mdis: '/scratch/jlaura/spice/mess-e_v_h-spice-6-v1.0/messsp_1000/extras/mk' # Messenger
 mex: '/scratch/jlaura/spice/mex-e_m-spice-6-v1.0/mexsp_1000/EXTRAS/MK' # Mars Explorer
 mro: '/data/spice/mro-m-spice-6-v1.0/mrosp_1000/extras/mk' # Mars Reconnaissance Orbiter
 kaguya: '/data/spice/SELENE/kernels/mk/'
-dawn: '/data/spice/dawn-m_a-spice-6-v1.0/dawnsp_1000/extras/mk'
+dawn: '/scratch/spice/dawn-m_a-spice-6-v1.0/dawnsp_1000/extras/mk/'
```
kaitlyndlee: does the dawn path need to be /scratch/jlaura/spice?

acpaquette: Nope, I copied the metakernels into the /scratch/spice area and we will just need to update the path for them to point to /scratch/jlaura/spice for the actual data

ale/base/data_naif.py
```diff
@@ -1,7 +1,9 @@
 import spiceypy as spice
 import numpy as np
+from networkx import DiGraph
+
```
jessemapel: The networkx import isn't needed here

ale/transformation.py
```diff
@@ -1,194 +1,99 @@
...
-        """
-        chain = [self]
-        current_parent = self.parent
-        while current_parent is not None:
-            chain.append(current_parent)
-            current_parent = current_parent.parent
-        return chain
+        for s, d in frame_changes:
+            for i, time in enumerate(times):
+                rotation_matrix = spice.pxform(spice.frmnam(s), spice.frmnam(d), time)
```
jessemapel: We cannot hard code a SPICE call into this class. We need to use this same class to handle rotation derived from things like ISIS spiceinit'd cubes also.

ale/transformation.py
```diff
@@ -1,194 +1,99 @@
...
+                quat_from_rotation = spice.m2q(rotation_matrix)
+                quats[i,:3] = quat_from_rotation[1:]
+                quats[i,3] = quat_from_rotation[0]
+            rotation = TimeDependentRotation(quats, times, s, d)
```
jessemapel: Not all of the rotations in the frame chain are going to be time dependent rotations. That said, the time dependent vs constant rotation is kind of a headache right now. I could see just going to all time dependent.

ale/transformation.py
```diff
@@ -1,194 +1,99 @@
...
-               The children nodes in the frame tree
-    rotation : ConstantRotation or TimeDependentRotation
-               The rotation from this frame to the frame of the parent node
+    frame_changes : list
+                    A list of tuples that represent the rotation from one frame
+                    to another. These tuples should all be NAIF codes for
+                    reference frames
+    ephemeris_time : list
+                     A of ephemeris times that need to be rotated for each set
+                     of frame rotations in the frame chain
```
jessemapel: For right now we are storing all of our time dependent rotations at the same time, but in the future we need to go away from this. The current ISIS functionality is to store each time dependent rotation at different times based on what information is available in the SPICE kernels and reduction within a tolerance.

ale/formatters/isis_formatter.py
```diff
@@ -24,51 +25,50 @@ def to_isis(driver):
...
+        time_dependent_rotation = frame_chain.compute_rotation(1, source_frame)
         instrument_pointing['CkTableStartTime'] = time_dependent_rotation.times[0]
         instrument_pointing['CkTableEndTime'] = time_dependent_rotation.times[-1]
         instrument_pointing['CkTableOriginalSize'] = len(time_dependent_rotation.times)
         instrument_pointing['EphemerisTimes'] = time_dependent_rotation.times
         instrument_pointing['Quaternions'] = time_dependent_rotation.quats
-    if time_dependent_sensor_frame != sensor_frame:
-        forward_path, reverse_path = time_dependent_sensor_frame.path_to(sensor_frame)
+
+    if time_dependent_sensor_frame['rotation'].source != sensor_frame:
```
jessemapel: This can just use the node ID, it doesn't need to access the rotation.

ale/formatters/isis_formatter.py
```diff
@@ -24,51 +25,50 @@ def to_isis(driver):
...
 
     body_rotation = {}
-    target_frame = j2000.find_child_frame(driver.target_frame_id)
-    time_dependent_target_frame = j2000.last_time_dependent_frame_between(target_frame)
-    if time_dependent_target_frame != j2000:
-        forward_path, reverse_path = j2000.path_to(time_dependent_target_frame)
+    target_frame = driver.target_frame_id
+    source_frame, destination_frame, time_dependent_target_frame = frame_chain.last_time_dependent_frame_between(target_frame, 1)
+
+    if time_dependent_target_frame['rotation'].source != 1:
```
jessemapel: Same as previous comment, this can just use the node ID because the node ID is the frame ID.

tests/pytests/test_isis_formatter.py
```diff
@@ -74,24 +74,30 @@ def sensor_position(self):
...
+
         body_rotation = TimeDependentRotation(
             np.array([[0, 0, 0, 1], [0, 0, 0, 1]]),
             np.array([0, 1]),
             100,
             1
         )
-        body_fixed = FrameNode(100, parent=j2000, rotation=body_rotation)
+        frame_chain.add_edge(100, 1, rotation=body_rotation)
+        frame_chain.add_edge(1, 100, rotation=body_rotation.inverse())
```
jessemapel: It would be nice if we could tuck needing to add both edges behind a convenience function, or overloading add_edge.

jessemapel:
Thinking about this more, I think adding a method that creates edges on the graph from two ids and a rotation; then moving the current init method into a separate function makes sense. This way, I can create a similar function for data_isis and everything still have a nice instantiation interface.

ale/transformation.py
acpaquette: I'm wrestling with this right now. Looks like ISIS has a good chunk of logic that gathers the various frames from spice to generate the time dependent frame sets and the constant frame sets. Not sure the best way to go about solving this currently

jessemapel: For right now, we are ignoring that logic because the methods for that are not wrapped in spiceypy yet.

jessemapel: The only time dependent rotations we have to deal with are those from spiceinit'd ISIS cubes.

acpaquette: Fair, this does result in problems regarding to_isis and to_usgscsm will be different depending on the source. If you do to_usgscsm from a dawn isis cube you'll get an additional constant rotation that you would have if you use to_usgscsm on a dawn pds image.

ale/transformation.py
acpaquette: Handled in a new classmethod called from_spice

ale/transformation.py
jessemapel: to_usgscsm doesn't care about constant vs time dependent rotations, everything boils down to 1 time dependent rotation anyway. This is a problem for to_isis and it will need to be adressed under ALE work.

acpaquette:
@jessemapel Is this good to merge once CI passes?

jessemapel:
@acpaquette Yep

PR #218, opened 2019-07-22
Moved detector start pixel into base and defaulted to 0

jessemapel:
This sets the default to the first image line being the first detector line and the first image sample being the first detector sample.

### DIFF
--- a/tests/pytests/test_lro_drivers.py
+++ b/tests/pytests/test_lro_drivers.py
@@ -46,12 +46,6 @@ def test_spacecraft_name(driver):
 def test_sensor_model_version(driver):
     assert driver.sensor_model_version == 2
 
-def test_detector_start_sample(driver):
-    assert driver.detector_start_sample == 1
-
-def test_detector_start_line(driver):
-    assert driver.detector_start_sample == 1
-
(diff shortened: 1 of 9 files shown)


jessemapel:
We will also want to re-run the comparisons after this update because it should shift things a bit.

tests/pytests/test_lro_drivers.py
```diff
@@ -46,12 +46,6 @@ def test_spacecraft_name(driver):
 def test_sensor_model_version(driver):
     assert driver.sensor_model_version == 2
 
-def test_detector_start_sample(driver):
-    assert driver.detector_start_sample == 1
-
-def test_detector_start_line(driver):
-    assert driver.detector_start_sample == 1
```
Kelvinrr: Why remove the test instead of updating the value?

jessemapel: Because this is no longer defined in the concrete driver, so this is just testing the base Driver class.

jessemapel: I thought about adding a test for the base Driver class, but I don't think it's worth it

PR #219, opened 2019-07-22
Adds simple light time correction

### LINKED ISSUE #77: Add light-time support to Ale.

jessemapel:
Fixes #77 

This adds simple light time correction to the center of the body. For spacecraft like LRO and Kaguya, this may be significantly different from light time correction to the surface. ISIS has some [additional code](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/isis/src/base/objs/Spice/SpacecraftPosition.cpp#L179) to do light time to surface instead. Should we implement that?

### DIFF
--- a/ale/base/data_naif.py
+++ b/ale/base/data_naif.py
@@ -32,8 +32,9 @@ class NaifSpice():
         Returns the type of light time correciton and abberation correction to
         use in NAIF calls.
 
-        This defaults to no correction, concrete drivers should override this of
-        they need to use light time correction.
+        This defaults to light time correction and abberation correction (LT+S),
+        concrete drivers should override this if they need to either not use
+        light time correction or use a different type of light time correction.
 
         Returns
         -------
@@ -42,7 +43,7 @@ class NaifSpice():
           See https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/req/abcorr.html
           for the different options available.
         """
-        return 'NONE'
+        return 'LT+S'
 
     @property
     def odtx(self):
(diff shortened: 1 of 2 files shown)


jessemapel:
I went through the ISIS kernels to see if any of the currently supported sensors use non-standard light time. The only one was LRO which uses no light time correction.

This is still a change for most of our sensors because previously we were using no light time correction and now they are in line with ISIS to use `LT+S` correction.

ale/base/data_naif.py
```diff
@@ -26,6 +26,25 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
 
+    @property
+    def light_time_correction(self):
+        """
+        Returns the type of light time correciton and abberation correction to
+        use in NAIF calls.
+
+        This defaults to light time correction and abberation correction (LT+S),
+        concrete drivers should override this if they need to either not use
+        light time correction or use a different type of light time correction.
```
jessemapel: This is a change from ISIS. Right now, ISIS determines if a sensor uses light time or not based on a [keyword](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/isis/src/base/objs/Spice/LightTimeCorrectionState.cpp#L125) in the IAK. We can continue to do that and not have individual drivers override this.

The main problem with continuing what ISIS does is that it's purely an on or off option for the entire sensor. By putting this into the driver, we can make light time correction choices on a per observation basis.

AustinSanders:
Looks good -- merging

PR #221, opened 2019-07-23
Changed usgscsm to reinterpolate positions and rotations

jessemapel:


ale/formatters/usgscsm_formatter.py
```diff
@@ -81,21 +82,57 @@ def to_usgscsm(driver):
...
+            velocities = np.asarray(velocities)
+            x_interp = BPoly.from_derivatives(position_times,
+                                              np.vstack((positions.T[0], velocities.T[0])).T,
+                                              extrapolate=True)
+            y_interp = BPoly.from_derivatives(position_times,
+                                              np.vstack((positions.T[1], velocities.T[1])).T,
+                                              extrapolate=True)
+            z_interp = BPoly.from_derivatives(position_times,
+                                              np.vstack((positions.T[2], velocities.T[2])).T,
+                                              extrapolate=True)
```
jessemapel: these transposes suck, is there a better way to do this? Basically this function wants [[pos_0, vel_o],[pos_1, vel1],...,[pos_n, vel_n]]

Kelvinrr: you can reduce to one transpose for position and orientation:
```
px,py,pz = np.asarray(positions).T
vx,vy,vz = np.asarray(velocities).T
```

that would make things a bit cleaner.

PR #223, opened 2019-07-25
Fixed how the ephemeris start time and exposure duration are calculated for LRO

kaitlyndlee:
The values used for the offsets were taken from an IAK and needed to be hardcoded. The equations for these calculations were taken from the LroNarrowAngleCamera class in ISIS. I added in the offsets that are 0 to make sure our equations match ISIS's and it is clear what we are doing.


jessemapel:
Looks good but could use a test

PR #224, opened 2019-07-26
Updates for ISIS ideal line scan

jessemapel:
We can now work with noproj output for line scan images.


ale/transformation.py
```diff
@@ -53,12 +53,13 @@ def create_rotations(rotation_table):
                            rotation_table['J2000Ang3'][:-1]]).T
         angles = polyval(scaled_times, coeffs).T
         # ISIS is hard coded to ZXZ (313) Euler angle axis order.
+        # SPICE also interprets Euler angle rotations as negative rotations,
```
kberryUSGS: :+1: This is so odd I'm glad you made a note of it here.

notebooks/write_IdealLsIsisLabelIsisSpiceDriver.ipynb
```diff
@@ -0,0 +1,131 @@
+{
```
kberryUSGS: It might be worth mentioning that `noproj` is the ISIS application typically used to make undistorted cubes in the notebook. (In case someone is trying to figure out how they might create an undistorted image that will be of the correct format to be read by the `IdealLsIsisLabelIsisSpiceDriver`)

ale/transformation.py
```diff
@@ -53,12 +53,13 @@ def create_rotations(rotation_table):
...
+        # so negate them before passing to scipy.
         time_dep_rot = TimeDependentRotation.from_euler('zxz',
-                                                        angles,
+                                                        -angles,
```
Kelvinrr: Any leads on why this is necessary?

jessemapel: Under the particulars section of [eul2m](https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/FORTRAN/spicelib/eul2m.html)

```
      A word about notation:  the symbol

        [ x ]
             i

     indicates a rotation of x radians about the ith coordinate axis.
     To be specific, the symbol

        [ x ]
             1

     indicates a coordinate system rotation of x radians about the
     first, or x-, axis; the corresponding matrix is

        +-                    -+
        |  1      0       0    |
        |                      |
        |  0    cos(x)  sin(x) |.
        |                      |
        |  0   -sin(x)  cos(x) |
        +-                    -+

     Remember, this is a COORDINATE SYSTEM rotation by x radians; this
     matrix, when applied to a vector, rotates the vector by -x
     radians, not x radians.  Applying the matrix to a vector yields
     the vector's representation relative to the rotated coordinate
     system.
```

This is what ISIS uses to actually apply the rotations, so we need to mirror eul2m's behavior of rotating about the negative axis (which is equivalent to rotating by the negative angle).

jessemapel:
New ISD is < 0.1 pixels off :tada: compared to ISIS

jessemapel:
I've had to make some significant changes to the Rotation classes so I'm re-requesting a review on this.

ale/base/base.py
```diff
@@ -296,3 +296,15 @@ def ephemeris_stop_time(self):
           The stop time of the image in ephemeris seconds past the J2000 epoch.
         """
         pass
+
+    @property
+    def center_ephemeris_time(self):
```
acpaquette: This needs to be removed in the data_isis class as well

ale/base/data_isis.py
```diff
@@ -297,6 +268,14 @@ def __enter__(self):
         """
         return self
 
+    def __exit__(self, exc_type, exc_val, exc_tb):
```
acpaquette: Not sure if this is better than just erroring out. If you're using an ISIS based driver you don't need either the __enter__ or __exit__

ale/base/data_naif.py
```diff
@@ -363,18 +363,22 @@ def ephemeris_start_time(self):
...
         -------
         : double
-          Center ephemeris time for an image
+          Ephemeris stop time of the image
         """
-        return (self.ephemeris_start_time + self.ephemeris_stop_time) / 2
+        if self.spacecraft_clock_stop_count:
+            return spice.scs2e(self.spacecraft_id, self.spacecraft_clock_stop_count)
+        else:
+            return spice.str2et(self.utc_stop_time.strftime("%Y-%m-%d %H:%M:%S"))
```
acpaquette: This should be removed, this will cause timing errors and result in center times that don't match up with ISIS

notebooks/write_IdealLsIsisLabelIsisSpiceDriver.ipynb
```diff
@@ -0,0 +1,131 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# change to desired PDS3 image path\n",
+    "file_name = '/Users/jmapel/ale/ESP_016076_2175_REDmos_hijitreged.balance.noproj.8bit.cub'\n",
+    "\n",
+    "from ale.transformation import FrameChain, create_rotations\n",
```
kaitlyndlee: Do we need these imports?

ale/base/data_isis.py
jessemapel: This allows for usage both inside and outside of a context. Adding this just makes the usage syntax the same across of the drivers.

jessemapel:
The constant re-interpolation is a valid concern, especially with the rotations. I tried to keep the re-interpolation to a minimum.

### Positions and Velocities

Right now this is what the interpolation looks like:

- We take exactly what's on the cube. If the cube has a function, we evaluate the function at the times that the function was originally fit over.
- Next the positions and velocities are rotated by rotations that are potentially interpolated
- In the USGSCSM formatter, the positions and velocities are re-interpolated

So, positions and velocities are getting interpolated 1-2 times and rotated by rotations that are potentially interpolated.

### Rotations

Rotations have more complex interpolation going on that makes it a bit tricky to track down how much they get interpolated. The reinterpolation code assures that the re-interpolated rotations is equal to the original at the times the original was defined. This helps keep errors from building up as long as we re-interpolate at more times. Here's what it looks like:
- Similar to positions, we take directly what's on the cube. If the rotations are functions, then we evaluate at the times the functions were originally fit over.
- When the rotations are computed from the frame_chain they are multiplied together which requires re-interpolation. This takes the union of the times that the two multiplications are defined at and re-interpolates both at those times. For most drivers, there are ~3 multiplications that happen when computing the sensor to target rotation.
- In the USGSCSM formatter, the sensor to target rotation is re-interpolated.

So, rotations are interpolated 3-4 times, but ~3 of those times, the rotations at the original times are un-changed.

PR #225, opened 2019-07-29
Isis CTX Driver Fixes

acpaquette:
Changes to get an ISIS Ctx driver working for comparisons with ISIS

### DIFF
--- a/ale/drivers/mro_drivers.py
+++ b/ale/drivers/mro_drivers.py
@@ -15,7 +15,7 @@ from ale.base.type_distortion import RadialDistortion
 from ale.base.type_sensor import LineScanner
 
 
-class MroCtxIsisLabelIsisSpiceDriver(Driver, IsisSpice, LineScanner, RadialDistortion):
+class MroCtxIsisLabelIsisSpiceDriver(IsisLabel, IsisSpice, LineScanner, RadialDistortion, Driver):
 
     @property
     def instrument_id(self):
@@ -29,7 +29,10 @@ class MroCtxIsisLabelIsisSpiceDriver(Driver, IsisSpice, LineScanner, RadialDisto
         : str
           instrument id
         """
-        return "N/A"
+        id_lookup = {
+        "CTX" : "MRO_CTX"
+        }
+        return id_lookup[super().instrument_id]
 
     @property
     def spacecraft_id(self):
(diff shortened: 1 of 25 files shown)


ale/drivers/mro_drivers.py
```diff
@@ -39,27 +42,25 @@ def spacecraft_id(self):
         : int
           Naif ID code for the spacecraft
         """
-        return "N/A"
+        return "-74"
```
Kelvinrr: Is this not in the kernels?

ale/drivers/mro_drivers.py
```diff
@@ -39,27 +42,25 @@ def spacecraft_id(self):
...
+    def sensor_model_version(self):
         """
         Returns
         -------
         : int
-          Naif ID code for the instrument
+          ISIS sensor model version
         """
-        return int(self.label["IsisCube"]["Kernels"]["NaifFrameCode"])
+        return 1
```
kberryUSGS: I'm not sure how this is used by CSM, so perhaps it doesn't matter, but couldn't this come from ["IsisCube"]["Kernels"]["CameraVersion"] ?

ale/drivers/mro_drivers.py
jessemapel: The `spice.bods2c(self.spacecraft_name)` call in data_naif should cover this if it's in the kernels as @Kelvinrr pointed out. Is the spacecraft_name not returning the correct value?

ale/drivers/mro_drivers.py
jessemapel: This is not used at all by CSM, but this value should be coming in off of that location. That code probably needs to be added to data_isis

acpaquette:
@Kelvinrr This doesn't really fix the comparison to ISIS when outputting a label from a PDS image but makes it so I could compare the isis directly to a label generated from a cube with the same positions, velocities, etc.

ale/drivers/mro_drivers.py
acpaquette: Does this work on the ISIS side? I thought we didn't load any kernels?

jessemapel: No this doesn't work without kernels sorry!

acpaquette: I think we could get this from one of the frame codes but it would be a little hacky

PR #226, opened 2019-07-30
added code for generating mks from spice inited cube

Kelvinrr:
Need to think of a clean way to inject the custom metakernel into an ALE driver instance. Metakernel setter probably good enough? 

Tests coming in a bit

### DIFF
--- a/ale/util.py
+++ b/ale/util.py
@@ -10,6 +10,8 @@ import pvl
 from collections import OrderedDict
 from itertools import chain
 
+import pysis
+
 def get_metakernels(spice_dir=config.spice_root, missions=set(), years=set(), versions=set()):
     """
     Given a root directory, get any subdirectory containing metakernels,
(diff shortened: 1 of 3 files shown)


ale/util.py
```diff
@@ -99,3 +103,72 @@ def find_latest_metakernel(path, year):
...
+    # just work with full path
+    cube = os.path.abspath(cube)
+    cubelabel = pvl.load(cube)
+    kernel_group = cubelabel['IsisCube']['Kernels']
+
+    # Start loading. Stuff that might have multiple entries first
+    mk_paths['TargetPosition'] = kernel_group.get('TargetPosition', None)[1:]
+    mk_paths['InstrumentPosition'] = kernel_group.get('InstrumentPosition', None)[1:]
+    mk_paths['InstrumentPointing'] = kernel_group.get('InstrumentPointing', None)[1:]
+    mk_paths['TargetAttitudeShape'] = [kernel_group.get('TargetAttitudeShape', None)]
```
jessemapel: There's a chance these won't have "Table" at the beginning, a more explicit check may be better. This is likely a corner case with old data though so I'm not sure.

PR #227, opened 2019-07-30
visual diff code

Kelvinrr:
Super sloppy code copy and pasted from my notebook. Only imports were cleaned up. 

names and such are all out of wack, copy-pasted code and vis should be separate for data computation. This can all happen in a different PR. 

The interface should be checked so notebooks won't have to change as much.

### DIFF
--- a/ale/vis.py
+++ b/ale/vis.py
@@ -1,0 +1,206 @@
+
+import tempfile
+
+import json
+import knoten
+import pvl
+import pyproj
+import csmapi
+
+from numbers import Number
+
+import numpy as np
+import pandas as pd
+
+from pysis import isis
+from pysis.exceptions import ProcessError
+
+import holoviews as hv
+from holoviews import opts, dim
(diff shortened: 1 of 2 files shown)


ale/vis.py
```diff
@@ -0,0 +1,206 @@
...
+    """
+    source_pyproj = pyproj.Proj(proj = source_proj, a = semi_major, b = semi_minor)
+    dest_pyproj = pyproj.Proj(proj = dest_proj, a = semi_major, b = semi_minor)
+
+    y, x, z = pyproj.transform(source_pyproj, dest_pyproj, record[0], record[1], record[2], **kwargs)
+
+    return y, x, z
+
+
+def point_info(cube_path, x, y, point_type, allow_outside=False):
```
Kelvinrr: `point_info` and `reproject` are required to get it working but currently exist only in autocnet. I'm working on getting them PRed into `pysis` and `knonten` respectively.

PR #230, opened 2019-08-01
Adding summing mode logic to mdis driver

SgStapleton:
There are also changes in here for hard-coding iak values. Is this still how we are wanting to do this?

### DIFF
--- a/ale/drivers/messenger_drivers.py
+++ b/ale/drivers/messenger_drivers.py
@@ -92,6 +92,25 @@ class MessengerMdisPds3NaifSpiceDriver(Pds3Label, NaifSpice, Framer, Driver):
           instrument id
         """
         return ID_LOOKUP[super().instrument_id]
+        
+    @property
+    def sampling_factor(self):
+        """
+        Returns the summing factor from the PDS3 label. For example a return value of 2
+        indicates that 2 lines and 2 samples (4 pixels) were summed and divided by 4
+        to produce the output pixel value.
+        
+        NOTE: This is overwritten for the messenger driver as the value is stored in "MESS:PIXELBIN" 
+
+        Returns
+        -------
+        : int
+          Number of samples and lines combined from the original data to produce a single pixel in this image
+        """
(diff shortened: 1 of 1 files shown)


ale/drivers/messenger_drivers.py
```diff
@@ -150,27 +169,33 @@ def detector_center_sample(self):
...
+        
+        NOTE: These values are defined in an ISIS iak file
 
         Returns
         -------
         : float
           center detector sample
         """
-        return float(spice.gdpool('INS{}_BORESIGHT'.format(self.ikid), 0, 3)[0])
+        # return float(spice.gdpool('INS{}_BORESIGHT'.format(self.ikid), 0, 3)[0])
```
kaitlyndlee: Should we keep this commented line?

ale/drivers/messenger_drivers.py
```diff
@@ -150,27 +169,33 @@ def detector_center_sample(self):
...
+        
+        NOTE: These values are defined in an ISIS iak file
 
         Returns
         -------
         : float
           center detector line
         """
-        return float(spice.gdpool('INS{}_BORESIGHT'.format(self.ikid), 0, 3)[1])
+        # return float(spice.gdpool('INS{}_BORESIGHT'.format(self.ikid), 0, 3)[1])
```
kaitlyndlee: Same as above.

ale/drivers/messenger_drivers.py
SgStapleton: Probably not? I left it in there for now to make sure I was handling this right. I can clean this up once confirmed.

ale/drivers/messenger_drivers.py
SgStapleton: (same as above)

kaitlyndlee:
Looks like we need to update the mdis test, then can be merged.

acpaquette:
Once the tests are fixed to reflect this changed this can get merged in. :+1:

PR #231, opened 2019-08-01
Added code to account for ISIS ccd center

jessemapel:
ISIS uses 0.5 based detector pixels so all the drivers that use the '_BORESIGHT` keywords need to subtract 0.5 to convert to 0 based detector pixels. The exception is that line scan instruments use 0 based detector lines. The drivers that use the `CCD_CENTER` keywords need to account for whatever is defined in the kernels. For example Dawn FC has -0.5 based Detector coordinates.

### DIFF
--- a/ale/drivers/dawn_drivers.py
+++ b/ale/drivers/dawn_drivers.py
@@ -216,12 +216,17 @@ class DawnFcPds3NaifSpiceDriver(Pds3Label, NaifSpice, Framer, Driver):
         Returns center detector sample acquired from Spice Kernels.
         Expects ikid to be defined. This should be the integer Naid ID code for
         the instrument.
+
+        We have to add 0.5 to the CCD Center because the Dawn IK defines the
+        detector pixels as 0.0 being the center of the first pixel so they are
+        -0.5 based.
+
         Returns
         -------
         : float
           center detector sample
         """
-        return float(spice.gdpool('INS{}_CCD_CENTER'.format(self.ikid), 0, 2)[0])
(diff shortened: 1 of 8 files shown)


kaitlyndlee:
Is this only for ctx or other line scanners?

jessemapel:
@kaitlyndlee This is for anything that uses [LineScanCameraDetectorMap](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/isis/src/base/objs/LineScanCameraDetectorMap/LineScanCameraDetectorMap.cpp#L144)

from some prelimenary grepping that is
* LROC NAC
* Chandrayaan1 M3
* Themis IR
* Rosetta Virtis
* MOC NAC
* Apollo Panoramic Camera
* CTX
* HiRISE
* CRISM
* LEISA
* MVIC
* Kaguya MI
* Dawn VIR
* HRSC
* Ideal Line Scan

jessemapel:
So I looked at [VariableLineScanCameraDetectorMap](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/isis/src/base/objs/VariableLineScanCameraDetectorMap/VariableLineScanCameraDetectorMap.cpp#L130) and it looks like it dispatches to [CameraDetectorMap](https://github.com/USGS-Astrogeology/ISIS3/blob/dev/isis/src/base/objs/CameraDetectorMap/CameraDetectorMap.cpp#L98) which appears to be 0 based?

scsides:
That's what we agreed on

PR #232, opened 2019-08-02
Update Spacecraft Position

kberryUSGS:
Updated spacecraft_position by fixing the order of the arguments to `spkezr` and negate resulting vector so it still points the correct direction.

### DIFF
--- a/ale/base/data_naif.py
+++ b/ale/base/data_naif.py
@@ -299,17 +299,22 @@ class NaifSpice():
             ephem = self.ephemeris_time
             pos = []
             vel = []
+         
             for time in ephem:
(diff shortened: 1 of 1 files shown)


acpaquette:
Weird mac faliure, restarted the mac test all should be good though

kberryUSGS:
@acpaquette Do you need special permissions to restart the build? I didn't see any way to do this.

acpaquette:
@kberryUSGS Potentially? I would just double check that your logged into travis and that your github is linked to it. Outside of that I'm not sure

kberryUSGS:
@acpaquette I indeed wasn't logged into Travis. Thanks!

PR #235, opened 2019-08-02
Updated mix-in order so that sensor type is first

jessemapel:
There is a conflict between data_isis and the sensor type mix-ins for the epehemeris_stop_time property. Changed all of the mix-in order so that the type comes before the data to avoid getting the wrong one.


kberryUSGS:
Just to double-check, the relative order of the Label mixin and the Spice-source mixin don't matter, right? Because those aren't consistent across all the sensors here.

jessemapel:
@kberryUSGS Correct, there are no conflicts between what they define so they can be in any order.

PR #242, opened 2019-08-06
Updated cmake and build recipe

### LINKED ISSUE #240: GTest needs to be optional for deploy
For the Conda-forge deploy, it is beneficial to have GTest (and the tests in general) be an optional part of the build process.  For example, usgscsm does this here: https://github.com/USGS-Astrogeology/usgscsm/blob/dev/CMakeLists.txt#L57

jessemapel:
Fixes #240

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -26,30 +26,26 @@ requirements:
   build:
   - cmake>=3.10
   - eigen
-#  - eigen=3.3.3
   - gsl
-#  - gsl==2.2.1
+  - netowrkx
   - numpy
   - openblas
-#  - openblas==0.2.19
   - python
-#  - python==3.6
   - python-dateutil
   - pvl
+  - pysis
   - spiceypy
   run:
   - cmake>=3.10
   - eigen
-#  - eigen=3.3.3
   - gsl
-#  - gsl==2.2.1
+  - netowrkx
   - numpy
   - openblas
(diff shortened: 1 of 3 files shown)


recipe/meta.yaml
```diff
@@ -26,30 +26,26 @@ requirements:
...
-#  - gsl==2.2.1
+  - netowrkx
   - numpy
   - openblas
-#  - openblas==0.2.19
   - python
-#  - python==3.6
   - python-dateutil
   - pvl
+  - pysis
```
jlaura: Most of these are likely not build dependencies. I believe only cmake, eigin, and gsl are build dependencies. Then python is a host dependency. Everything else should be run time dependencies.

Also, is ISIS/pysis an ALE dependency?

jessemapel: We currently have pysis under the utils module. It probably needs to be taken out.

recipe/meta.yaml
```diff
@@ -26,30 +26,17 @@ requirements:
...
```
jlaura: In the short term I think merging this is fine. I am going to open an issue about the pysis dependency. From what I am seeing in util.py, pysis is not actually required. In fact, an ISIS install is not even required. What is needed is an ISIS data area and some parsing that can be done. 

My thinking here is that I really would like to avoid a circular ISIS<->ALE dependency.

Issue #244, opened 2019-08-13
config.yml not installing


jlaura:
It does not look like the config.yml file is installing properly. It looks like the config file is not being specified in either a MANIFEST.ini or in the setup.py.  See https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=61904 for the failure in the build pipeline. I wonder how this is working in the travisCI env.

PR #245, opened 2019-08-19
Added install of config file

scsides:


jlaura:
Fixes #244

setup.py
```diff
@@ -24,5 +24,6 @@
     packages=find_packages(),
     long_description="""\
     An Abstraction library for reading, writing and computing ephemeris data
-    """
+    """,
+    data_files=[('ale', ['ale/config.yml'])]
```
jessemapel: Is there a reason why this uses data_files instead of package_data? I'm having a hard time with the setuptools docs.

Kelvinrr: From looking around there doesn't seem to be much of a functional difference between the two other than package_data is used more frequently and is more recommended. No explanation as to why.

jessemapel: Can this be moved to package_data then? It also looks less verbose.

jessemapel:
This is good to merge once CI passes

setup.py
scsides: Done

Issue #243, opened 2019-08-06
Refactor pysis out of ALE


jlaura:
Right now pysis is introducing a dependency on ISIS. It looks like the code in util.py is pulling a preference file to know about the ISIS3DATA area and kernel layout in order to build meta kernels. This could (arguably should) be refactored so that os.environ calls are used or the function args are expanded and the user can pass in the file topology as needed.

My concern is that this block will introduce a circular dependency between ALE and ISIS.

PR #248, opened 2019-08-19
Removed pysis with fixes

scsides:


ale/util.py
```diff
@@ -106,23 +104,62 @@ def find_latest_metakernel(path, year):
         raise Exception(f'No metakernels found in {path} for {year}.')
     return metakernel
 
+
+import collections
```
Kelvinrr: move this to top of the module.

Kelvinrr:
Since this was preapproved by Jessie as well, should be good to go.

jessemapel:
Fixes #243

PR #250, opened 2019-08-21
kernel list support added

Kelvinrr:


tests/pytests/conftest.py
```diff
@@ -31,6 +31,10 @@ def sct2e(self, *args):
         return 0.1
     def cidfrm(self, *args):
         return (2000, "Test_Body_Frame", True)
+    def furnsh(self, *args):
+        if not hasattr(self, '_loaded_kernels'):
+            self._loaded_kernels = []
+        self._loaded_kernels.append(args)
```
Kelvinrr: A simple solution to see what order things are being furnished. Might cause stateful errors if used all over a test module but only used in one test for now. Maybe find a better solution?

jessemapel: Can this be replaced by mocking the furnsh and unload functions, then checking what they were called with and in what order?

ale/base/base.py
```diff
@@ -1,8 +1,7 @@
 import pvl
 import abc
-from abc import ABC
 
-class Driver(ABC):
```
jessemapel: Does removing inheritance from ABC stop the abstract property tags from throwing an exception if the method gets called without being replaced?

ale/base/data_naif.py
```diff
@@ -21,9 +23,17 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+        if self.kernels:
+            [spice.unload(k) for k in self.kernels]
+        else:
+            spice.unload(self.metakernel)
 
     @property
+    def kernels(self):
+        if not hasattr(self, '_kernels'):
+            self._kernels =  self._props.get('kernels', None)
+        return self._kernels
```
jessemapel: Removing the metakernel property and having everything go through the kernels property would make the loading and unloading code simpler. ALE is only concerned with throwing these values into furnsh and unload so passing the metakernel though the kernels property as just a single element list will be functionally the same.

ale/util.py
```diff
@@ -132,31 +132,54 @@ def read_pref(path):
...
-            argprefs = reaf_pref(isis_preferences)
+            argprefs = read_pref(isis_preferences)
         else:
             warnings.warn(f'{isis_preferences} does not exist, ignoring')
 
     try:
         homeprefs = read_pref(os.path.join(os.path.expanduser("~"), '.Isis', 'IsisPreferences'))
-    except: 
+    except Exception as e:
+        print(f'error: {e}')
```
jessemapel: Saying the error happened while trying to read the User's preferences would be helpful.

ale/util.py
```diff
@@ -132,31 +132,54 @@ def read_pref(path):
...
         homeprefs = {}
 
-
-    isisrootprefs = read_pref(os.path.join(os.environ["ISISROOT"], 'IsisPreferences'))
+    try:
+        isisroot = os.environ['ISISROOT']
+        print(isisroot)
+        isisrootprefs = read_pref(os.path.join(os.environ["ISISROOT"], 'IsisPreferences'))
+    except Exception as e:
+        print('failed {e}')
```
jessemapel: Same here, saying the error happened while trying to read the ISISROOT preferences would be helpful.

tests/pytests/test_util.py
```diff
@@ -30,9 +76,72 @@ def test_kernel_from_cube_order(cube_kernels):
...
+
+def test_dict_to_lower():
+    data = {'F': {'O' : {'O': 1}, 'B': {'A': 2}}}
+    expected  = {'f': {'o' : {'o': 1}, 'b': {'a': 2}}}
+    assert util.dict_to_lower(data) == expected
+
+def test_get_prefernces_arg_isisroot_home(monkeypatch, tmpdir, pvl_one_group, pvl_two_group, pvl_three_group):
+    tmpdir.mkdir('.Isis')
+    monkeypatch.setenv('ISISROOT', str(tmpdir))
+    monkeypatch.setenv('HOME', str(tmpdir))
```
jessemapel: Do these go away after the test finishes?

ale/util.py
Kelvinrr: that was a debug print, probably just remove for now.

ale/util.py
Kelvinrr: same as above.

tests/pytests/test_util.py
Kelvinrr: yes

ale/base/base.py
Kelvinrr: Everything changed with notimplemented error so it should still throw an error.

ale/base/data_naif.py
Kelvinrr: As of now, it's hard to separate kernels and metakernels because of the mission-specific configs. When we figure out the load interface, we will make it easier to classify Driver's mission from the modules and move to one metakernel definition so we should be able to consolidate the two then.

ale/util.py
Kelvinrr: Updated so it naturally fail if file exists and an exception is thrown.

tests/pytests/conftest.py
Kelvinrr: I would argue that this solution is simpler, I imagine for the same reason that we haven't replaced the rest of SimpleSpice with mocks.

Kelvinrr:
@jessemapel Should be good for re-review

ale/base/base.py
```diff
@@ -1,8 +1,7 @@
 import pvl
 import abc
```
jessemapel: This line can also be removed now

PR #253, opened 2019-08-27
Updated load interface 

Kelvinrr:
Also a bunch of new tests.


jessemapel:
kernel loading and management is the worst

PR #254, opened 2019-08-28
Updated Kaguya test to use actual kernels

jessemapel:
This should also improve the down track sensor model alignment and timing.

Because Kaguya TC observes the Moon, it uses the Moon Mean Earth reference frame which is defined by a binary PCK. You cannot slice up a binary PCK and the converted binary PCK is ~5 MB. So, I decided to exclude all of the binary PCK stuff and instead use the IAU_MOON reference frame defined by the regular PCK.


jessemapel:
This is now ready for final review and merging

tests/pytests/conftest.py
```diff
@@ -109,17 +111,18 @@ def convert_kernels(kernels):
...
-    }
     binary_kernels = []
     updated_kernels = []
     for kernel in kernels:
         split_kernel = os.path.splitext(kernel)
-        if split_kernel[1] in ext_map:
-            subprocess.call(['tobin', os.path.join(data_root, kernel)])
-            kernel = split_kernel[0] + ext_map[split_kernel[1]]
-            binary_kernels.append(kernel)
+        if 'x' in split_kernel[1].lower():
```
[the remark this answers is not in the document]
jessemapel: Their extension always starts with x or b

ale/drivers/selene_drivers.py
```diff
@@ -91,7 +91,7 @@ def sensor_frame_id(self):
         : int
           Sensor frame id
         """
-        return spice.bods2c("LISM_{}_HEAD".format(super().instrument_id))
+        return spice.namfrm("LISM_{}_HEAD".format(super().instrument_id))
```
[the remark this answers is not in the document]
jessemapel: yes, bos2c was faiking, but namfrm was working for this. I'm, not sure why.

ale/base/data_naif.py
```diff
@@ -289,7 +289,7 @@ def sun_position(self):
         sun_state, _ = spice.spkezr("SUN",
                                      self.center_ephemeris_time,
                                      self.reference_frame,
-                                     self.light_time_correction,
+                                     'NONE',
```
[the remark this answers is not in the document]
jessemapel: My understanding is that ISIS doesn't do light time correction for sun distances

tests/pytests/test_kaguya_drivers.py
```diff
@@ -1,119 +1,238 @@
...
-KaguyaTcPds3NaifSpiceDriver.metakernel = get_mockkernels
-
-@pytest.fixture
-def driver():
-    return KaguyaTcPds3NaifSpiceDriver("")
+@pytest.fixture(scope="module", autouse=True)
+def test_kernels():
+    kernels = get_image_kernels('TC1S2B0_01_06691S820E0465')
+    updated_kernels, binary_kernels = convert_kernels(kernels)
+    spice.furnsh(updated_kernels)
```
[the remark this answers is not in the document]
jessemapel: I've tried a few different arrangements and I'm not finding anything elegant or simple to make this work.

tests/pytests/test_kaguya_drivers.py
```diff
@@ -1,119 +1,185 @@
...
+        with driver as failure:
+            pass
+
+def test_no_spice_root(driver, monkeypatch):
+    monkeypatch.delenv('ALESPICEROOT', raising=False)
+    reload(ale)
+
+    with pytest.raises(EnvironmentError):
+        with driver as failure:
+            pass
```
acpaquette: I think that the no metakernels and no spice root is being tested in test_env.py

tests/pytests/test_kaguya_drivers.py
```diff
@@ -1,119 +1,185 @@
...
+            'unit': 'm'},
+        'name_model': 'USGS_ASTRO_LINE_SCANNER_SENSOR_MODEL',
+        'interpolation_method': 'lagrange',
+        'line_scan_rate': [[0.5, -1.300000011920929, 0.006500000000000001]],
+        'starting_ephemeris_time': 292234259.82293594,
+        'center_ephemeris_time': 292234261.12293595,
+        't0_ephemeris': -1.300000011920929,
+        'dt_ephemeris': 0.5200000047683716,
+        't0_quaternion': -1.300000011920929,
+        'dt_quaternion': 0.5200000047683716}
```
acpaquette: This looks like a pds3 usgscsm isd, I thought we wanted to compare against an ISIS usgscsm isd?

jessemapel: We want to test against both. I can co back in and add the ISIS comparison in another PR. I want to get the conftest.py stuff in here so that other PRs can use it.

tests/pytests/test_kaguya_drivers.py
jessemapel: Yeah, there's no Selene specific code here, so this shouldn't be tested.

jessemapel:
The Final Travis check on this did pass, Travis just didn't report back to Github

jessemapel:
@acpaquette I patched out the spice calls because I want to test the properties in isolation. I also do not want to deal with funrshing the kernels because furnshing kernels is prone to leaking outside of tests.

jessemapel:
@acpaquette The MDIS changes are needed to account for not doing LT correction on sun distance.

ale/base/data_naif.py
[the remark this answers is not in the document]
jessemapel: Confirmed with Stuart that ISIS does not use light time correction for the SUN

tests/pytests/test_kaguya_drivers.py
acpaquette: So can these two tests be removed?

jessemapel: yes

PR #258, opened 2019-08-28
Added TargetPosition to Slicing NB

acpaquette:
Added code to look into the TargetPosition spks if body codes cannot be found in the InstrumentPosition spk


acpaquette:
This probably won't get merged but there are a few things in here that should be PR'd once #259 is merged.

PR #259, opened 2019-08-28
Moved kernel parsing logic from slicing notebook into ale utils.

jessemapel:
This moves the kernel parsing logic into utils. It also is a little bit more robust. You can easily combine the spkmerge configs into one simple file. It also slices out every reference frame from the CKs.


jessemapel:
@acpaquette I think this covers #258 also

Kelvinrr:
Any reason for the notebook to stick around? If there is re-usable logic in the notebook, can we get a script instead?

PR #260, opened 2019-08-29
Basic Mro Load Test

acpaquette:
Still a bit of a mess but this is more to see what people think of this method.

A few things that could be done is parameterization of the unittest class to potentially test multiple drivers with just a single test suite. This also has some code from #254 for reading xc and xsp kernels into bc and bsp kernels for testing.


acpaquette:
Something to note is that some kernels are shared between the mro and messenger which may be shared with other spacecraft as well.

jessemapel:
@acpaquette Try making the test_kernels fixture scoped to the module so that it only converts things once.

acpaquette:
I'm at such a lose for why this is happening.

acpaquette:
@jessemapel This works locally so I have no idea why this is failing on travis. The only thing I can think of potential path issues

acpaquette:
It looks like this is a problem with naifs tobin command line tool only being able to handle paths with a length >= 129 characters.

acpaquette:
Mac differences in truthdata :(

acpaquette:
GREEN! :smile:

.travis.yml
```diff
@@ -45,7 +45,7 @@ install:
   - conda install pytest
 
 script:
-  - PYTHONPATH=. pytest tests/pytests
+  - PYTHONPATH=. pytest tests/pytests -v
```
jessemapel: should do `-vv` for full diffs

tests/pytests/test_mro_drivers.py
```diff
@@ -1,72 +1,234 @@
...
-    with patch.dict(Pds3NaifDriver.label, SAMPLE_FIRST_PIXEL=0) as f:
-        assert Pds3NaifDriver.detector_start_sample == 0
-
-def test_sensor_model_version_pds3(Pds3NaifDriver):
-    assert Pds3NaifDriver.sensor_model_version == 1
-
-@patch('ale.base.label_isis.IsisLabel.instrument_id', 'CTX')
-def test_instrument_id_isis(IsisLabelNaifDriver):
-    assert IsisLabelNaifDriver.instrument_id == 'MRO_CTX'
+# 'Mock' the spice module where it is imported
```
jessemapel: Old comment to be removed

tests/pytests/test_mro_drivers.py
```diff
@@ -1,72 +1,234 @@
...
+        # Check to change the line based on ISIS vs PDS3
+        # This is due to a processing step in mroctx2isis that removes samples
+        # based on some flags on the label
+        if label_type == 'isis3':
+            usgscsm_compare_dict['image_samples'] = 5000
+        assert compare_dicts(usgscsm_isd_obj, usgscsm_compare_dict) == []
+    else:
+        assert compare_dicts(usgscsm_isd_obj, isis_compare_dict) == []
+
+# ========= Test isislabel and isisspice driver =========
```
jessemapel: I like splitting the different drivers into different tests

jessemapel:
Review comments are largely superficial, so this can be merged as is or after the minor changes.

PR #261, opened 2019-09-03
Updated MDIS tests to new load approach

jessemapel:


ale/drivers/mes_drivers.py
```diff
@@ -225,6 +225,21 @@ class MessengerMdisIsisLabelNaifSpiceDriver(IsisLabel, NaifSpice, Framer, Driver
...
+
+        Messenger MDIS ISIS labels use upper camel case so this converts it to\
+        all upper case.
+
+        Returns
+        -------
+        : str
+          Spacecraft name
+        """
+        return super().platform_name.upper()
```
jessemapel: I thought about making this change in the IsisLabel class but I'm not sure if that's correct

tests/pytests/test_mdis_drivers.py
```diff
@@ -34,199 +33,88 @@ def driver(request):
...
+         patch('ale.base.data_naif.spice.bods2c', return_value=-12345) as bods2c:
+         assert driver.detector_start_line == 10.0
+         gdpool.assert_called_with('INS-12345_FPUBIN_START_LINE', 0, 1)
 
-def test_focal_length(driver):
-    assert driver.focal_length == 549.5535053027719
+def test_detector_center_sample(driver):
+    with patch('ale.drivers.mes_drivers.spice.gdpool', return_value=[512.5, 512.5, 1]) as gdpool, \
+         patch('ale.base.data_naif.spice.bods2c', return_value=-12345) as bods2c:
+         assert driver.detector_center_sample == 512
```
jessemapel: This is a SPICE call for the ISIS driver, but not for the PDS3 driver. So, I'm not checking if gdpool was called here.

jessemapel:
I need to re-slice the kernels on this PR but I want to wait until #262 is merged

tests/pytests/test_mdis_drivers.py
```diff
@@ -5,19 +5,18 @@
...
 
+from conftest import get_image_label, get_image_kernels, convert_kernels, compare_dicts
 import ale
 from ale.drivers.mes_drivers import MessengerMdisPds3NaifSpiceDriver
 from ale.drivers.mes_drivers import MessengerMdisIsisLabelNaifSpiceDriver
 
-@pytest.fixture(scope="module", autouse=True)
-def test_kernels():
+@pytest.fixture()
+def test_kernels(scope='module'):
```
acpaquette: Shouldn't the scope be in the fixture parentheses?

tests/pytests/test_mdis_drivers.py
```diff
@@ -34,199 +33,88 @@ def driver(request):
...
-                                    9.06001059499675E-4,
-                                    0.0,
-                                    3.5748426266207586E-4,
-                                    0.0,
-                                    1.0040104714688569E-5,
-                                    0.0,
-                                    1.0040104714688569E-5])
+@pytest.mark.parametrize("label_type", ["pds3", "isis3"])
+def test_load(test_kernels, label_type):
+    isd = {
```
acpaquette: Again I might make this a fixture but this is also fine.

ale/drivers/mes_drivers.py
kberryUSGS: I'm not sure either. Here definitely seems valid, though.

ale/base/type_sensor.py
```diff
@@ -51,7 +51,7 @@ def ephemeris_time(self):
         : ndarray
           ephemeris times split based on image lines
         """
-        return np.linspace(self.ephemeris_start_time,  self.ephemeris_stop_time, self.image_lines / 64)
+        return np.linspace(self.ephemeris_start_time,  self.ephemeris_stop_time, int(self.image_lines / 64))
```
scsides: I know this was here before, but in my tests, any image_lines less than 128 only gets one value back from linespace. Larry routinely worked with small images, sometimes just one line, one samp, many bands

ale/drivers/mes_drivers.py
```diff
@@ -225,6 +225,21 @@ class MessengerMdisIsisLabelNaifSpiceDriver(IsisLabel, NaifSpice, Framer, Driver
     Driver for reading MDIS ISIS3 Labels. These are Labels that have been ingested
     into ISIS from PDS EDR images but have not been spiceinit'd yet.
```
scsides: This comment is misleading in that it makes it sound like we can't re-spiceinit a cube, or am I misreading it?

ale/base/type_sensor.py
scsides: Also, this call doesn't force the stop_time to be the last entry in the return. Should we add the endpoint arg to force it?

jessemapel: Adding the endpoint arg doesn't force both endpoints when you pass 1 for the number of samples. I've changed this to always do at least 2.

ale/drivers/mes_drivers.py
jessemapel: I've re-worded this so it's clearer. This driver can work with spiceinit'd cubes but it will ignore any attached spice data. You need a different driver to work with spiceinit'd data.

PR #262, opened 2019-09-04
Updated LRO test

jessemapel:
This adds a significant amount of new test data ~3.5MB. I could potentially more carefully cut up the binary kernels to only pull around the exact time ranges we need instead of everything between potential light time correction times and the actual image time.


jlaura:
Three questions:

(1) Is this data volume representative of the volume we would expect to see for all missions? Are we expecting the volumes to increase with distance from the sun? If so, how does this scale across 60+sensors?
(2) The driver/PR is for LRO, but I only see LRO-NAC. Would an LRO-WAC driver have a similar amount of data? This relates to #1 in thinking about total data volume.
(3) What is an acceptable test data volume and when do we move the test data to a public facing place (S3?) where travis can pull it at test time?

thareUSGS:
While it makes it a little harder for us to manage, is compression an option (that we can handle in code - e.g. gzip)? If *.gz, uncompress first. 

This should save at least ~35% or more on ASCII files with lots of numbers. 

Unfortunately, S3 does not support compression over the wire. from :https://www.sumologic.com/insight/10-things-might-not-know-using-s3/
_...S3 bandwidth and cost constraints generally make compression worth it._

jessemapel:
@thareUSGS We don't want to compress the files because we don't want them to be binary. Storing binary files in git results in huge repo sizes due to diffs.

jessemapel:
I've updated the kernel slicing notebook so that it now only cuts out a small time window for the light time correction.

I also added light time correction back to the sun position computations.

jessemapel:
@jlaura This only covers LRO-NAC because that's all our driver supports right now. Adding LRO-WAC support will be part of the supporting all of the ISIS sensor models work.

I'm not sure what the cap on test data volume is. With this current scheme the SPKs and CKs are ~200 KB and 80 of that is the Messenger test data which I haven't re-cut up. That should be scalable to a large number of images.

Now I'm more concerned about the size of the SCLK, framing, and Instrument kernels. They are all publicly available on NAIF though, and we can dynamically download those at test time if needed.

scsides:
LRO NAC is probably our current worst case for data size. It has multiple binary CKs.

We could reduce the data more by selecting simultaneous NAC, WAC, ..., other supported LRO instrument observations.

If we pull a specific SCLK, Frame, Insturment kernel from NAIF archives, we could do as Jesse suggests. If we pull the most recent we could be re-introducing the fragility we are trying to remove. For example: SCLK data can and has been reprocessed if better coefficients are derived, resulting in a time difference, resulting in position and orientation differences, and thus test failure.

jessemapel:
We may be able to come up with a way to share common files between tests. WAC tests will use the same SCLK as this test.

tests/pytests/test_lro_drivers.py
```diff
@@ -1,59 +1,138 @@
...
+        spacecraft_id.return_value = 1234
+        assert driver.ephemeris_start_time == 107.4
+        scs2e.assert_called_with(1234, "1/270649237:07208")
 
 @patch('ale.base.label_pds3.Pds3Label.exposure_duration', 1)
 def test_exposure_duration(driver):
     assert driver.exposure_duration == 1.0045
+
+def test_load(test_kernels):
+    isd = {
```
acpaquette: Might be worth making this a fixture, but storing it here is also fine

tests/pytests/test_lro_drivers.py
```diff
@@ -1,59 +1,138 @@
...
+         patch('ale.drivers.lro_drivers.LroLrocPds3LabelNaifSpiceDriver.exposure_duration', \
+               new_callable=PropertyMock) as exposure_duration, \
+         patch('ale.drivers.lro_drivers.LroLrocPds3LabelNaifSpiceDriver.spacecraft_id', \
+               new_callable=PropertyMock) as spacecraft_id:
+        exposure_duration.return_value = 0.1
+        spacecraft_id.return_value = 1234
+        assert driver.ephemeris_start_time == 107.4
+        scs2e.assert_called_with(1234, "1/270649237:07208")
 
 @patch('ale.base.label_pds3.Pds3Label.exposure_duration', 1)
```
acpaquette: Might want to change this patch to be consistent with the other patches above.

PR #263, opened 2019-09-04
Data Naif Tests

acpaquette:
Initial stab at updating the data_naif tests to use actual spice data, specifically MRO spice data.


tests/pytests/test_data_naif.py
```diff
@@ -1,54 +1,133 @@
...
-    with patch('spiceypy.furnsh') as furnsh:
-        with test_naif_data_with_kernels as t:
-            assert furnsh.call_args_list == [call('one'), call('two'), call('three'), call('four')]
+class test_data_naif(unittest.TestCase):
+
+    def setUp(self):
+        kernels = get_image_kernels('B10_013341_1010_XN_79S172W')
+        self.updated_kernels, self.binary_kernels = convert_kernels(kernels)
+        spice.furnsh(self.updated_kernels)
+        FakeNaifDriver = type("FakeNaifDriver", (data_naif.NaifSpice, label_pds3.Pds3Label, base.Driver), {})
```
jessemapel: This makes this test dependent upon the Pds3Label mix-in. Since you're not actually passing a label and patching all of these properties out, I don't think this is needed.

tests/pytests/test_data_naif.py
```diff
@@ -1,54 +1,133 @@
...
 import pytest
 
 import numpy as np
+import spiceypy as spice
+
+import unittest
+from unittest.mock import patch, call, PropertyMock
 
-from unittest.mock import patch, call
+from conftest import get_image_kernels, convert_kernels, get_image_label
```
jessemapel: I don't think you use get_image_label

tests/pytests/test_data_naif.py
```diff
@@ -1,54 +1,133 @@
...
+
+            assert self.driver.ephemeris_start_time == -631195148.8160816
+
+    def test_ephemeris_start_time(self):
+        with patch('ale.base.label_pds3.Pds3Label.spacecraft_name', new_callable=PropertyMock) as spacecraft_name, \
+             patch('ale.base.label_pds3.Pds3Label.spacecraft_clock_stop_count', new_callable=PropertyMock) as spacecraft_clock_stop_count:
+            spacecraft_name.return_value = 'MRO'
+            spacecraft_clock_stop_count.return_value = '1/60000'
+
+            assert self.driver.ephemeris_stop_time == -631135148.8160615
```
jessemapel: Why are there 2 tests for this?

acpaquette: Miss name, should be a start and stop time test.

tests/pytests/test_data_naif.py
acpaquette: This can work for some properties but not others. For example target_name is defined in the base driver with a `raise NotImplementedError` but instrument_id is not defined in the base Driver so it needs to have the Pds3Label mixin. I'm just not sure if instrument_id should be on the base driver or not.

acpaquette:
Uh oh.

acpaquette:
I'll get this fixed up tomorrow. Merged in @jessemapel lro PR which is going to cause commit problems :grimacing:

PR #265, opened 2019-09-04
Updated Dawn Test

paarongiroux:


paarongiroux:
I'm getting a `No Such Driver for Label` exception in the load test after my latest commit, not sure why this is happening

jessemapel:
Try directly instantiating the driver and calling the formatter on it.

jessemapel:
This is still missing the SPK, `*.xsp`

kberryUSGS:
Try pulling down master. Looks like this is failing now because the light time correction for sun was re-added in a different PR so you're getting different calculated values for sun position/velocity.

PR #266, opened 2019-09-05
Updates the Cassini test to the new format

kberryUSGS:
Note: The ISD I used for the load test comparison was generated using a meta-kernel consisting only of the kernels attached to the ISIS cube and the ik, which for some reason is not used by ISIS. If I use the default `/scratch/spice` meta-kernel, there are different position/velocity values in the output ISD because different kernels are used. The kernel that makes most of the difference seems to be `de430.bsp` -- it's used in the Cassini ISIS cube, but not in the default Naif meta-kernel used by ale. If this is a problem or not the way these tests are supposed to work, let me know and I can look into this.


jessemapel:
This looks good can you remove the cmt and conf files. They are not needed.

kberryUSGS:
Huh, didn't fail locally. Checking it out.

acpaquette:
@kberryUSGS Might want to pull master down. I think there was an update to light time correction for the sun position and velocity which would result in different test results

kberryUSGS:
@acpaquette Just noticed that it was the light-time correction. Thanks!

PR #269, opened 2019-09-06
Changed CMake setup to use targets instead of variables

jessemapel:
I also removed the old JSON include directory.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,6 +18,7 @@ set(CMAKE_CXX_STANDARD 11)
 find_package(GSL    REQUIRED)
 find_package(Eigen3 3.3 REQUIRED NO_MODULE)
 find_package(PythonLibs REQUIRED COMPONENTS Development Interpreter Compiler)
+find_package(nlohmann_json REQUIRED)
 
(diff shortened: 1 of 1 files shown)


jessemapel:
After #253 removed the included JSON headers, the current CMake setup would fail. This was causing issues with CD. See https://github.com/conda-forge/ale-feedstock/pull/1

CMakeLists.txt
```diff
@@ -17,29 +17,28 @@ set(CMAKE_CXX_STANDARD 11)
 # Third Party Dependencies
 find_package(GSL    REQUIRED)
 find_package(Eigen3 3.3 REQUIRED NO_MODULE)
-find_package(PythonLibs REQUIRED COMPONENTS Development Interpreter Compiler)
+find_package(Python REQUIRED COMPONENTS Development)
```
jessemapel: I don't think we need the Interpreter or the Compiler here. We just need the development package right?

Kelvinrr: Correct

Kelvinrr: Correct

PR #271, opened 2019-09-09
Added ISIS special NAIF keywords

jessemapel:
Updated observer swapping code requires updated truth data for MRO CTX, Cassini ISS, and Dawn FC.


ale/base/data_naif.py
```diff
@@ -426,6 +439,39 @@ def detector_center_line(self):
...
+        """
+        Returns if light time correction should be made to the surface instead of
+        to the center of the body. This is defined by a keyword in ISIS IAKs.
+        If the keyword is not defined in any loaded kernels then False is returned.
+
+        Expects ikid to be defined. This should be an integer containing the
+        Naif Id code of the instrument.
+        """
+        try:
+            swap = spice.gcpool('INS{}_LT_SURFACE_CORRECT'.format(self.ikid), 0, 1)[0]
```
kberryUSGS: potentially use a different variable name than `swap` since it's the light time correction to surface (vs. center) here?

jessemapel: Good catch! The LT to surface logic isn't in here. I need to look more into exactly how it works.

kberryUSGS: It seems fine to me to handle that additional logic in a different PR if that's what you were thinking?

ale/base/data_naif.py
```diff
@@ -426,6 +439,39 @@ def detector_center_line(self):
...
+        is returned.
+
+        Expects ikid to be defined. This should be an integer containing the
+        Naif Id code of the instrument.
+        """
+        try:
+            swap = spice.gcpool('INS{}_SWAP_OBSERVER_TARGET'.format(self.ikid), 0, 1)[0]
+            return swap.upper() == "TRUE"
+        except:
+            return False
```
acpaquette: The only problem I have with theses gcpool calls is that they are dependent on having the ISIS addendum kernel loaded into memory which isn't the case if you're using a Pds3 image. So for any pds3 image you'll get a position or velocity in one direction and in ISIS you'll get the swapped result.

kberryUSGS: This is a good point. A few followup questions: for input pds3 images, are we expecting ale's output to be supported in ISIS, or just csm? I know the option of `to_isis` exists, but what is the current goal?  If it's just csm, then doesn't the ISIS camera model take the swapping into account? So you'd get positions and velocities swapped in the isds between the output of ale's `to_isis` and `to_usgscsm` formatters, but would ultimately end up with matching (or at least close) positions out of campt and csm's image-to-ground call?

kberryUSGS: This is potentially a side-note: Would we ever want the ISDs generated by `to_isis` vs `to_usgscsm to output different exterior orientation values due to different swap_observer_target settings between ISIS and CSM? i.e. Since the ISIS camera model incorporates the swap, but presumably the CSM camera model doesn't(?) would it make sense for these numbers to be different, but then ultimately end up with matching (or at least close) positions out of campt and csm's image-to-ground call?

jessemapel: We could change this to default to returning True because it's the correct answer in 99% of cases (Active sensors like Radar being the exception) and then update all of the ISIS IAKs to have FALSE in them.

ale/base/data_naif.py
jessemapel: I've added this as a task on our google doc

ale/base/data_naif.py
jessemapel: We can actually start updating IAKs right now with the currently hard-coded defaults in ISIS. It shouldn't make any differences in current ISIS.

ale/base/data_naif.py
```diff
@@ -322,17 +326,26 @@ def sensor_position(self):
...
+                                           time,
+                                           self.reference_frame,
+                                           self.light_time_correction,
+                                           self.target_name,)
+                    pos.append(state[:3])
+                    vel.append(state[3:])
+
+            # By default, SPICE works in km, so convert to m
+            self._position = [p * 1000 for p in pos]
+            self._velocity = [v * 1000 for v in vel]
```
Kelvinrr: Probably cleaner if you store target and spacecraft in temp vars and make one splezr call using the temp vars. Makes the intent clearer and uses fewer lines.

ale/base/data_naif.py
```diff
@@ -426,6 +439,39 @@ def detector_center_line(self):
...
+        """
+        Returns if light time correction should be made to the surface instead of
+        to the center of the body. This is defined by a keyword in ISIS IAKs.
+        If the keyword is not defined in any loaded kernels then False is returned.
+
+        Expects ikid to be defined. This should be an integer containing the
+        Naif Id code of the instrument.
+        """
+        try:
+            surface_correct = spice.gcpool('INS{}_LT_SURFACE_CORRECT'.format(self.ikid), 0, 1)[0]
```
Kelvinrr: Do we expect this key in IKs or IAKs?

jessemapel: We need to be able to use IAKs because ISIS depends on IAKs right now.

Kelvinrr: It seems fine for now for the sake fo a result, but I think after we get a clean result from ALE + ISIS it's worth evaluating how to get ISIS specific stuff out of ALE and back into ISIS.

kberryUSGS:
@jessemapel Looks like there are merge conflicts that need to be resolved after the mdis PR was merged.

jessemapel:
@kberryUSGS I am going to wait until #274 is done to update this because there will be more merge conflicts.

jessemapel:
@kberryUSGS @acpaquette This is ready now

jessemapel:
I'm going to go ahead and merge this if no one has comments.

PR #272, opened 2019-09-10
Added json dependency for new load

scsides:


jessemapel:
This is already in https://github.com/USGS-Astrogeology/ale/blob/master/environment.yml#L11

PR #274, opened 2019-09-13
Frame Chain

acpaquette:
Added in a frame trace and updated the frame chain to produce the full network of frames to traverse through.


acpaquette:
This PR is going to fail as it is blocked by a PR in spiceypy

jessemapel:
Can we server a build ourselves for right now?

acpaquette:
Do you mean for spiceypy @jessemapel?

kberryUSGS:
And this may be an odd question, but it is okay for us to make our own SpiceyPy build? I'm not familiar with the details of the licensing, etc...

jessemapel:
@acpaquette Yes, I mean spiceypy

@kberryUSGS Spiceypy is MIT licensed, so we just need to include the LICENSE file in any release we publish.

ale/base/data_naif.py
```diff
@@ -322,23 +322,77 @@ def sensor_position(self):
...
             # direction to what ALE needs, so it must be multiplied by (-1)
-            self._position = [p * -1000 for p in pos]
-            self._velocity = [v * -1000 for v in vel]
+            self._position = [p * 1000 for p in pos]
+            self._velocity = [v * 1000 for v in vel]
         return self._position, self._velocity, self.ephemeris_time
 
+    @property
+    def frame_trace(self):
+        frame_codes = [self.sensor_frame_id]
```
jessemapel: Can you make this a separate function where the starting frame ID is passed in. We will need to do this for both the sensor frame and the body frame

ale/base/data_naif.py
```diff
@@ -322,23 +322,77 @@ def sensor_position(self):
...
+            frame_types.append(frame_type)
+        constant_frames = [frame_codes[i] for i, frame_type in enumerate(frame_types) if frame_type == 4]
+
+        time_dependent_frames = []
+        if len(constant_frames) != 0:
+            time_dependent_frames.append(constant_frames[-1])
+
+        for i, frame_type in enumerate(frame_types):
+            if frame_type != 4:
+                time_dependent_frames.append(frame_codes[i])
```
jessemapel: This will result in losing the order of the frame chain. For example, if the sensor frame chain is J2000 --ck--> target --ck--> spacecraft --fixed--> platform --ck--> gimble --fixed--> sensor then you will get:

constant_frames = [sensor, platform]
time_dependent_frames = [platform, gimble, spacecraft, target, J2000]

The way ISIS defines these is the time_dependent_frames is the rotation from J2000 to the last time dependent frame and then constant_frames is the rest. So, in ISIS this would look like:

constant_frames = [sensor, gimble]
time_dependent_frames = [gimble, platform, spacecraft, target, J2000]

ale/formatters/isis_formatter.py
```diff
@@ -29,25 +29,23 @@ def to_isis(driver):
...
-    if source_frame != sensor_frame:
-        # Reverse the frame order because ISIS orders frames as
-        # (destination, intermediate, ..., intermediate, source)
-        instrument_pointing['ConstantFrames'] = shortest_path(frame_chain, sensor_frame, source_frame)
-        constant_rotation = frame_chain.compute_rotation(source_frame, sensor_frame)
-        instrument_pointing['ConstantRotation'] = constant_rotation.rotation_matrix()
+    source_frame, destination_frame, time_dependent_sensor_frame = frame_chain.last_time_dependent_frame_between(1, sensor_frame)
+
+    # Reverse the frame order because ISIS orders frames as
+    # (destination, intermediate, ..., intermediate, source)
```
jessemapel: Old comment can be removed

ale/formatters/isis_formatter.py
```diff
@@ -29,25 +29,23 @@ def to_isis(driver):
...
+    time_dependent_rotation = frame_chain.compute_rotation(1, destination_frame)
+    instrument_pointing['TimeDependentFrames'] = shortest_path(frame_chain, destination_frame, 1)
+    instrument_pointing['CkTableStartTime'] = time_dependent_rotation.times[0]
+    instrument_pointing['CkTableEndTime'] = time_dependent_rotation.times[-1]
+    instrument_pointing['CkTableOriginalSize'] = len(time_dependent_rotation.times)
+    instrument_pointing['EphemerisTimes'] = time_dependent_rotation.times
+    instrument_pointing['Quaternions'] = time_dependent_rotation.quats[:, [3, 0, 1, 2]]
+
+    # Reverse the frame order because ISIS orders frames as
+    # (destination, intermediate, ..., intermediate, source)
```
jessemapel: Same here, this can be removed

ale/base/data_naif.py
```diff
@@ -322,23 +322,77 @@ def sensor_position(self):
...
+                    raise Exception(f"The ck rotation from frame {frame_codes[-1]} can not \
+                                      be found due to no pointing available at requested time \
+                                      or a problem with the frame")
+            elif frame_type is 4:
+                matrix, frame_code, found = spice.tkfram(frame_type_id)
+                if not found:
+                    raise Exception(f"The tk rotation from frame {frame_codes[-1]} can not \
+                                      be found")
+            elif frame_type is 5:
+                matrix, frame_code = spice.zzdynrot(frame_type_id, center, self.center_ephemeris_time)
```
jessemapel: I am not a huge fan of computing all of these rotations right here and then immediately throwing the rotation away to be recomputed later.

jessemapel: Is it possible to bake this into the generation of the frame chain?

kberryUSGS:
A SpiceyPy build is up at https://anaconda.org/usgs-astrogeology/spiceypy @acpaquette Do all your tests pass locally for this build? Because I pulled down this branch from your fork, installed the updated SpiceyPy and am still getting failures (wanting to make sure it's not a problem with the package I just uploaded.)

ale/base/data_naif.py
acpaquette: :+1: I agree that this probably shouldn't live in the data_naif module

ale/base/data_naif.py
acpaquette: :+1: Will get this fixed up

acpaquette:
@kberryUSGS No this is still getting failures that I am trying to deal with. Mainly the way I have changed the isis formatter is causing tests to fail.

acpaquette:
Mdis is still going to fail but that is because the observer and target need to be swapped. That should be the last fix that needs to go in for this PR to finally get through

acpaquette:
Passing on Linux!

jessemapel:
@kberryUSGS It looks like we need a mac build of spiceypy up too

PR #275, opened 2019-09-13
Updated how Angular Velocity is stored and computed in rotations

jessemapel:
The main part of this is that the interpolation also interpolates angular velocity. The angular velocity and rotation are interpolated independently. This is not a mathematically ideal approach, but a more robust approach is too complex and not actually needed.


tests/pytests/test_voyager_drivers.py
```diff
@@ -81,12 +83,10 @@ def test_kernels():
 
 @pytest.mark.parametrize("label_type", ['isis3'])
 @pytest.mark.parametrize("formatter", ['isis'])
-@pytest.mark.skip(reason="Fails due to angular velocity problems")
```
jessemapel: Now passing :tada:

ale/rotation.py
```diff
@@ -326,7 +366,7 @@ def rotate_velocity_at(self, pos, vel, et):
             skew = np.array([[0, -avs[indx, 2], avs[indx, 1]],
                              [avs[indx, 2], 0, -avs[indx, 0]],
                              [-avs[indx, 1], avs[indx, 0], 0]])
-            rot_deriv = np.dot(skew, rots[indx].as_dcm())
+            rot_deriv = np.dot(skew, rots[indx].as_dcm().T).T
```
jessemapel: This was a math error in the original implementation :(

ale/rotation.py
```diff
@@ -234,23 +247,47 @@ def _slerp(self, times):
...
+        if len(self.times) > 1:
+            steps = self.times[1:] - self.times[:-1]
+            rotvecs = (self._rots[1:] * self._rots[:-1].inv()).as_rotvec()
+            avs[1:-1] = rotvecs / steps[:, None]
+
+        # If available use actual angular velocity for extrapolation
+        # Otherwise use the adjacent interpolation interval
+        if self.av is not None:
+            avs[0] = self.av[0]
+            avs[-1] = self.av[-1]
```
acpaquette: Does the angular velocity array that is stored on the rotation class only have two values? I guess my bigger question is why are we interpolating the angular velocity between 0 and N if we have them stored on the object?

tests/pytests/test_cassini_drivers.py
```diff
@@ -61,8 +61,7 @@ def test_kernels(scope="module", autouse=True):
 def test_cassini_load(test_kernels, usgscsm_compare_dict):
     label_file = get_image_label("N1702360370_1")
     usgscsm_isd = ale.load(label_file, props={'kernels': test_kernels}, formatter='usgscsm')
-    print(usgscsm_isd['sensor_position'])
-    print(usgscsm_isd['sensor_orientation'])
+    print(usgscsm_isd)
```
acpaquette: Left over print

tests/pytests/test_voyager_drivers.py
```diff
@@ -81,12 +83,10 @@ def test_kernels():
...
 def test_voyager_load(test_kernels, label_type, formatter, isis_compare_dict):
     label_file = get_image_label('c2065022', label_type)
 
     usgscsm_isd_str = ale.loads(label_file, props={'kernels': test_kernels}, formatter=formatter)
     usgscsm_isd_obj = json.loads(usgscsm_isd_str)
     print(usgscsm_isd_obj)
```
acpaquette: Left over print again. Probably from my PR :grimacing:

ale/rotation.py
jessemapel: The angular velocity at each time is stored in the rotation but we don't use them for rotation interpolation. I did testing and it turns out using the stored angular velocities actually results in worse rotation interpolation. The angular velocity is not constant between rotations, so using it for SLERP results in discontinuities in the interpolated rotations. Using a more complex interpolation scheme is more work than we actually need.

This is why I'm not 100% happy with this, but it appears to match what both NAIF and ISIS actually use.

jessemapel: These two values are set because they are the angular velocities used to extrapolate rotations. Since we cannot compute an angular velocity here we either use the stored one or the computed values from the first/last interval.

tests/pytests/test_cassini_drivers.py
jessemapel: I actually left this in because it is nice to see the entire ISD when you get errors.

tests/pytests/test_voyager_drivers.py
jessemapel: Same as above, I actually like printing out the whole ISD on error.

acpaquette:
Okay, I'm fine with leaving the print statements in if no one else has a problem with it

jessemapel:
Is this good to merge then?

PR #276, opened 2019-09-19
Added tests for the rest of the Voyager cameras

jessemapel:
This should wait until after the AV PR is merged, then I will update it.


acpaquette:
@jessemapel Merge conflicts from the angular velocity PR

ale/rotation.py
```diff
@@ -217,12 +217,11 @@ def inverse(self):
         Get the inverse rotation, that is the rotation from the destination
         reference frame to the source reference frame.
         """
-        new_rots = self._rots.inv()
         if self.av is not None:
-            new_av = -new_rots.apply(self.av)
+            new_av = -self._rots.apply(self.av)
```
jessemapel: This is the same as the fix in the AV PR. I had defined the angular velocity as the speed at which frame 1 rotates with respect to frame 2. This and the transposes in the velocity apply steps fixes that.

jessemapel:
@acpaquette This is ready now.

ale/drivers/vo_drivers.py
```diff
@@ -6,7 +6,19 @@
...
+    def instrument_id(self):
+        sc_lookup = {
+        "VOYAGER_1" : "VG1",
+        "VOYAGER_2" : "VG2"
+        }
+        sensor_lookup = {
+        "NARROW_ANGLE_CAMERA" : "ISSNA",
+        "WIDE_ANGLE_CAMERA" : "ISSWA"
+        }
+        return sc_lookup[super().spacecraft_name] + '_' + sensor_lookup[super().instrument_id]
```
jessemapel: Note for people writing ISIS drivers. You don't need this for ISIS support, but this helps prevent load from accidentally choosing these. I had issues where load picked this for MRO data!

acpaquette: How would it have picked the voyager driver for MRO data?

acpaquette: I'm guessing it made the driver fine but couldn't format it with the to_isis or to_usgscsm formater?

jessemapel: The voyager driver worked with an MRO image and MRO kernels because it's very generic. It actually made it all the way through the formatter.

jessemapel: loads calls driver.instrument_id before loading kernels and such as a quick check. This function ensures that unless the spacecraft is actually Voyager it will fail then.

PR #278, opened 2019-09-20
Ale 0.5.0 PR

acpaquette:
PR for the release of ale version 0.4.0


jessemapel:
Why was the voyager module renamed back to voyager?

acpaquette:
@jessemapel I believe that vo is the viking orbiter not voyager

acpaquette:
Updated the version to 0.5.0 as this is the actual version we are on when looking at our releases

PR #279, opened 2019-09-23
voyager now fails if instrument name is not VOYAGER

Kelvinrr:


ale/drivers/voyager_drivers.py
```diff
@@ -30,7 +30,10 @@ def ikid(self):
 
     @property
     def spacecraft_name(self):
-        return super().spacecraft_name.replace('_', ' ')
+        name = super().spacecraft_name.replace('_', ' ')
+        if name.split(' ')[0] == "VOYAGER":
```
jessemapel: Needs an upper()

PR #280, opened 2019-09-23
Fixed AV multiplication and added MDIS ISIS test

jessemapel:
I meant to just make actual MDIS test from ISIS, but I found a weird bug with mutliplying angular velocities.


ale/rotation.py
```diff
@@ -110,7 +110,7 @@ def __mul__(self, other):
             new_rot = self._rot * other._rot
             return ConstantRotation(new_rot.as_quat(), other.source, self.dest)
         elif isinstance(other, TimeDependentRotation):
-            return TimeDependentRotation((self._rot * other._rots).as_quat(), other.times, other.source, self.dest, av=self._rot.apply(other.av))
+            return TimeDependentRotation((self._rot * other._rots).as_quat(), other.times, other.source, self.dest, av=other.av)
```
jessemapel: This and the other changes in rotation are because NAIF defines the angular velocity as always being in the source reference frame. See the detailed output in the docs for [xf2rav](https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/FORTRAN/spicelib/xf2rav.html)

tests/pytests/test_mro_drivers.py
```diff
@@ -235,12 +235,12 @@ def isis_compare_dict():
...
-                                               [-5.965469273385757e-05, 0.0009306551165985512, -0.00010670063014175982],
-                                               [-6.001692495046523e-05, 0.0009318842753552843, -0.00010732755544866732],
-                                               [-5.9466333983691536e-05, 0.0009375199811173681, -0.00010690951512741344]]},
+                           'AngularVelocity': [[-0.0006409728984903079, 0.0005054077299115119, 0.0004718267948468069],
+                                               [-0.0006410700774431097, 0.0005044862657976017, 0.0004731836236807216],
+                                               [-0.0006408186407087456, 0.0004992170698116158, 0.0004802237192760833],
+                                               [-0.0006363961683672021, 0.0004989647975959612, 0.00047654664046286975],
+                                               [-0.0006376443791903504, 0.0004996117504290811, 0.00047678850931380653],
+                                               [-0.0006404093657132724, 0.0005028749658176146, 0.0004805228583087444]]},
```
jessemapel: We still don't know if these are 100% accurate, so I am okay with how these have been rotated.

tests/pytests/test_mdis_drivers.py
```diff
@@ -20,57 +20,165 @@ def test_kernels():
...
+            'image_lines': 512,
+            'image_samples': 512,
+            'name_platform': 'MESSENGER',
+            'name_sensor': 'MERCURY DUAL IMAGING SYSTEM NARROW ANGLE CAMERA',
+            'reference_height': {
+                'maxheight': 1000,
+                'minheight': -1000,
+                'unit': 'm'},
+            'name_model': 'USGS_ASTRO_FRAME_SENSOR_MODEL',
+            'center_ephemeris_time': 483122606.85252464}}}
```
jessemapel: Do people like using a big dictionary like this to parameterize the loads test? I'd like to add an MDIS WAC image to the test.

scsides: To get the dict, are you just dumping an existing one and pasting it into the test?

ale/rotation.py
scsides: We are heavily tied to NAIF for ISIS. No reason to change that now, maybe in 6.0.

tests/pytests/test_mdis_drivers.py
jessemapel: The new ISIS dict was generated by running tabledump on the spiceinit'd test image. The only changes are flipping signs on quaternions and adding extra values in the NaifKeywords.

The pre-existing USGSCSM dict is just the old comparison dictionary from before this PR.

kaitlyndlee: I think doing it this way is fine and straightforward. We're going to have large dictionaries regardless.

PR #281, opened 2019-09-24
IsisSpice now fails when it cannot find tables

jessemapel:
This should cause load to function properly with cubes that don't have attached SPICE data.


Kelvinrr:
Is there a clean way to test these chnges?

jessemapel:
Yeah I can test this with one of the test labels

PR #286, opened 2019-09-27
Added ISIS label Naif spice driver for Viking

paarongiroux:


jessemapel:
Can you add a test for Viking 2?

paarongiroux:
> Can you add a test for Viking 2?

I'll get started on that

ale/drivers/viking_drivers.py
```diff
@@ -0,0 +1,61 @@
...
+import ale
+from ale.base.data_naif import NaifSpice
+from ale.base.label_isis import IsisLabel
+from ale.base.type_sensor import Framer
+from ale.base.base import Driver
+
+class VikingIsisLabelNaifSpiceDriver(Framer, IsisLabel, NaifSpice, Driver):
+
+    @property
+    def spacecraft_name(self):
```
jessemapel: You should add an instrument_id property similar to the other drivers that maps the label names to the NAIF names. This will help load(s) run faster for other drivers.

jessemapel: Then, you can re-use that here.

tests/pytests/test_viking_drivers.py
```diff
@@ -0,0 +1,109 @@
...
+from importlib import reload
+import json
+
+import unittest
+from unittest.mock import patch
+from conftest import get_image_label, get_image_kernels, convert_kernels, compare_dicts
+
+import ale
+from ale.drivers.viking_drivers import VikingIsisLabelNaifSpiceDriver
+from ale.formatters.isis_formatter import to_isis
```
jessemapel: This is not needed

tests/pytests/test_viking_drivers.py
```diff
@@ -0,0 +1,109 @@
...
+                           'Velocities': [[0.61887065223952,-0.54866918471373,-0.037514399215278]]},
+   'SunPosition': {'SpkTableStartTime': -679343589.99066,
+                   'SpkTableEndTime': -679343589.99066,
+                   'SpkTableOriginalSize': 1,
+                   'EphemerisTimes': [-679343589.99066],
+                   'Positions': [[242154815.26715,36901072.557543,10366793.725074]],
+                   'Velocities': [[-4.6653907628395,19.836822435663,9.2246820568878]]}}
+
+
+@pytest.fixture(scope="module", autouse=True)
```
jessemapel: Do not autouse this. There can be weird things that happen with autouse fixtures and unittest testCases.

tests/pytests/test_viking_drivers.py
```diff
@@ -0,0 +1,109 @@
...
+    def test_spacecraft_name(self):
+        assert self.driver.spacecraft_name == 'VIKING ORBITER 1'
+
+    def test_ikid(self):
+        assert self.driver.ikid == -27002
+
+    def test_ephemeris_start_time(self):
+        with patch('ale.drivers.viking_drivers.spice.scs2e', return_value=54321) as scs2e:
+             assert self.driver.ephemeris_start_time == 54324.99
+             scs2e.assert_called_with(-27999, '40031801')
```
jessemapel: You should test this with different exposure durations too.

tests/pytests/test_viking_drivers.py
```diff
@@ -0,0 +1,302 @@
...
+@pytest.mark.parametrize("formatter", ['isis'])
+@pytest.mark.parametrize("image", image_dict.keys())
+def test_viking1_load(test_kernels, label_type, formatter, image):
+    label_file = get_image_label(image, label_type)
+    isis_isd = ale.load(label_file, props={'kernels': test_kernels[image]}, formatter=formatter)
+    print(isis_isd)
+    assert compare_dicts(isis_isd, image_dict[image][formatter]) == []
+
+# ========= Test isislabel and naifspice driver =========
+class test_isis_naif(unittest.TestCase):
```
paarongiroux: not sure how/if i should be testing the individual properties across all spacecrafts/cameras. Make a new class for each case?

jessemapel: I think you'll have to do that, you can only test the things you expect to use different code in the other classes, you do not need to test every property with every image.

jessemapel: Actually there really isn't anything special that needs to be tested.

tests/pytests/test_viking_drivers.py
```diff
@@ -0,0 +1,302 @@
...
+        for kern in kern_list:
+            os.remove(kern)
+
+@pytest.mark.parametrize("label_type", ['isis3'])
+@pytest.mark.parametrize("formatter", ['isis'])
+@pytest.mark.parametrize("image", image_dict.keys())
+def test_viking1_load(test_kernels, label_type, formatter, image):
+    label_file = get_image_label(image, label_type)
+    isis_isd = ale.load(label_file, props={'kernels': test_kernels[image]}, formatter=formatter)
+    print(isis_isd)
```
kberryUSGS: Already merged, but maybe get rid of print statement?

jessemapel: We've been leaving in the isd print statements because if the test fails it can be helpful to have the output for ALE. If the test passes, output is not printed to the terminal.

PR #287, opened 2019-09-27
Refactored reading image data in mex and created pds3 test

kaitlyndlee:
We needed to update the way we were previously reading the image data to make the driver testable. Now, line_scan_rate does not read the image data and there are no dependencies on line_scan_rate, i.e., ephemeris_stop_time. We still need a way to actually test the image. Right now, I am mocking the image data.


ale/drivers/mex_drivers.py
```diff
@@ -322,77 +322,162 @@ def detector_center_sample(self):
...
+        Returns
+        -------
+        : list
+          Line scan rates
+        """
+        if not hasattr(self, '_line_scan_rate'):
+            times = [time - self.center_ephemeris_time for time in self.binary_ephemeris_times]
+            self._line_scan_rate = (self.binary_lines, 
+                                    times, 
+                                    self.binary_exposure_durations)
```
jessemapel: Is there a reason why this is stateful?

ale/drivers/mex_drivers.py
```diff
@@ -322,77 +322,162 @@ def detector_center_sample(self):
...
+        For HRSC, the exposure durations and ephemeris times are imbedded in the binary 
+        data of the image itself. Each line is stored in what is referred to as a 
+        "record" within the image. The label will have the size of each record, 
+        the number of records, and the number of records in the label, so the 
+        beginning of binary data can be calculated.
+
+        For each line/record of the binary data, the first 8 bytes make up the
+        double presicion value of the ephemeris time, with the next 4 bytes
+        making up the float value of the line exposure duration for the
+        associated line. NOTE: The image label specifies MSB_INTEGER as the byte
```
jessemapel: Are the images always MSB? Do we need to check the label for this?

ale/drivers/mex_drivers.py
```diff
@@ -322,77 +322,162 @@ def detector_center_sample(self):
...
+                # Offset for zero-based corrections, and then offest for ISIS pixel definition
+                lines.append(record+1-0.5)
+                times.append(eph_time)
+                durations.append(exp_dur)
+            # Only add records if exposure duration has changed since the line before
                     #elif exp_dur != durations[-1]:
                     #    # Offset for zero-based corrections, and then offest for ISIS pixel definition
                     #    lines.append(record+1-0.5)
                     #    times.append(eph_time)
                     #    durations.append(exp_dur)
```
jessemapel: Remove this commented code

ale/drivers/mex_drivers.py
```diff
@@ -322,77 +322,162 @@ def detector_center_sample(self):
...
+        which is calculated from the binary image data.
+
+        Returns
+        -------
+        : float
+          Center ephemeris time
+        """
         if not hasattr(self, '_center_ephemeris_time'):
-            self._center_ephemeris_time = (self.ephemeris_stop_time - self._ephemeris_start_time) / 2
+            self._center_ephemeris_time = (self.ephemeris_stop_time + self.ephemeris_start_time) / 2
```
jessemapel: Why is this stateful too?

ale/drivers/mex_drivers.py
```diff
@@ -322,77 +322,162 @@ def detector_center_sample(self):
...
+        For HRSC, the ephemeris stop time is calculated from the binary image data.
+
+        Returns
+        -------
+        : float
+          Ephemeris stop time
+        """
         if not hasattr(self, '_ephemeris_stop_time'):
-            _, _, _ = self.line_scan_rate
+            self._ephemeris_stop_time = self.binary_ephemeris_times[-1] + self.binary_exposure_durations[-1]
```
jessemapel: Is there an entry in the binary lists for every line? Usually we reduce the line scan rate table down to only the first line of each new exposure duration.

notebooks/write_MexHrscPds3LabelNaifSpiceDriver.ipynb
```diff
@@ -138,190 +64,40 @@
...
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": [
-    "def check_pixel(camera, cub, line, sample):\n",
-    "    \"\"\"Compares ISIS and USGSCSM pixel.\n",
+    "# with open(json_file, 'r') as fp:\n",
+    "#     usgscsm_dict = json.load(fp)\n",
```
jessemapel: Is there a reason for all of these changes to the notebook?

tests/pytests/test_mex_drivers.py
```diff
@@ -0,0 +1,347 @@
+import pytest
+import pvl
+import os
+import subprocess
+import numpy as np
+import spiceypy as spice
+from importlib import reload
```
jessemapel: This is not needed

tests/pytests/test_mex_drivers.py
```diff
@@ -0,0 +1,347 @@
...
+        patch('ale.drivers.mex_drivers.MexHrscPds3NaifSpiceDriver.binary_exposure_durations', \
+               new_callable=PropertyMock) as binary_exposure_durations, \
+        patch('ale.drivers.mex_drivers.MexHrscPds3NaifSpiceDriver.binary_lines', \
+               new_callable=PropertyMock) as binary_lines:
+        binary_ephemeris_times.return_value = [255744599.02748165, 255744599.04028246, 255744795.73322123]
+        binary_exposure_durations.return_value = [0.012800790786743165, 0.012800790786743165, 0.013227428436279297]
+        binary_lines.return_value = [0.5, 1.5, 15086.5]
+
+        driver = MexHrscPds3NaifSpiceDriver(label_file)
+        usgscsm_isd = to_usgscsm(driver)
```
jessemapel: This should be calling load, not the driver and the formatter.

tests/pytests/test_mex_drivers.py
```diff
@@ -0,0 +1,347 @@
...
+
+@pytest.fixture(scope="module", autouse=True)
+def test_kernels():
+    kernels = get_image_kernels('h5270_0000_ir2')
+    updated_kernels, binary_kernels = convert_kernels(kernels)
+    spice.furnsh(updated_kernels)
+    yield updated_kernels
+    spice.unload(updated_kernels)
+    for kern in binary_kernels:
+        os.remove(kern)
```
jessemapel: Do not furnsh the kernels in an auto-use fixture like this.

ale/drivers/mex_drivers.py
kaitlyndlee: It was stateful before I refactored the code, I was not sure if I should remove it or not.

notebooks/write_MexHrscPds3LabelNaifSpiceDriver.ipynb
kaitlyndlee: I updated the notebook with the new way we are selecting metakernels and outputting the json from the formatter functions. I can undo these changes if we do not want to update the notebooks yet.

ale/drivers/mex_drivers.py
jessemapel: The default should be non-stateful. Properties should only be stateful if they are time consuming to compute, have an external dependency such as a file, or have can be set.

notebooks/write_MexHrscPds3LabelNaifSpiceDriver.ipynb
jessemapel: The changes are good then. Can you remove the old commented out code.

tests/pytests/test_mex_drivers.py
kaitlyndlee: When I use the load function, an exception is thrown. This is the error for trying to load the Mex driver:
```
  File "/home/kdlee/builds/ale/ale/drivers/__init__.py", line 69, in load
    with res as driver:
  File "/home/kdlee/builds/ale/ale/base/data_naif.py", line 17, in __enter__
    [spice.furnsh(k) for k in self.kernels]
TypeError: 'function' object is not iterable
```

ale/drivers/mex_drivers.py
kaitlyndlee: Looking into this right now. After looking at hrsc2isis, the prefix data is always LSB regardless of the rest of the file. We might need to change this code.

tests/pytests/test_mex_drivers.py
jessemapel: You need to add the test_kernels fixture to that test. If you don't add the fixture to the test, it just looks like a function to Python.

ale/drivers/mex_drivers.py
kaitlyndlee: The code was treating it as LSB in the driver, but the comment was wrong. Should be all straightened out now.

tests/pytests/test_mex_drivers.py
```diff
@@ -0,0 +1,351 @@
+import pytest
+import pvl
+import os
+import subprocess
```
paarongiroux: pvl and subprocess unused imports?

ale/drivers/mex_drivers.py
```diff
@@ -322,76 +322,146 @@ def detector_center_sample(self):
...
+        return self._binary_exposure_durations
+
+
+    @property
+    def binary_ephemeris_times(self):
+        """
+        Returns the ephemeris times taken from the binary image data.
+
+        For HRSC, the ephemeris times are imbedded in the binary data of the image.
+        The ephemeris times start at the first bye of the line/records and are 8 bytes.
```
paarongiroux: small typo bye -> byte

ale/drivers/mex_drivers.py
kaitlyndlee: This logic was commented out in the code. I can put it back if that's what we want to do.

scsides: Minimizing that table seems like a good idea. A full table would be 10,000's of records where a minimized table might be10's or less. I suggest putting it back in.

PR #290, opened 2019-10-01
Juno ISIS Driver

acpaquette:
Adds a juno ISIS label, naif spice driver to ale.


ale/drivers/juno_drivers.py
```diff
@@ -0,0 +1,70 @@
...
+        -------
+        : double
+          Starting ephemeris time
+        """
+        if not hasattr(self, '_ephemeris_start_time'):
+            initial_time = super().ephemeris_start_time
+            frame_number = self.label['IsisCube']['Instrument']['FrameNumber']
+            inter_frame_delay = self.label['IsisCube']['Instrument']['InterFrameDelay'].value
+            start_time_bias = self.naif_keywords[f'INS{self.ikid}_START_TIME_BIAS']
+            inter_frame_delay_bias = self.naif_keywords[f'INS{self.ikid}_INTERFRAME_DELTA']
```
jessemapel: Is there a reason why the naif_keywords are used here instead of gdpool?

PR #293, opened 2019-10-03
Adds TGO Cassis ALE driver and light time to surface correction

kberryUSGS:
This PR adds a TGO CaSSIS driver (and associated test) to ALE and also adds the light time to surface correction to ALE. (Thanks to @jessemapel for the start of the LT+S work in #282!)


ale/base/data_naif.py
```diff
@@ -337,11 +338,36 @@ def sensor_position(self):
...
+
+                    radius_lt = (self.target_body_radii[2] + self.target_body_radii[0]) / 2 / (scipy.constants.c/1000.0)
+                    adjusted_time = time - obs_tar_lt + radius_lt
+                    ssb_tar_state, ssb_tar_lt = spice.spkezr(target,
+                                                             adjusted_time,
+                                                             'J2000',
+                                                             'NONE',
+                                                             'SSB')
+                    state = ssb_tar_state - ssb_obs_state
+                    matrix = spice.sxform("J2000", self.reference_frame, time)
```
kberryUSGS: The above calculation did not work with self.reference_frame, rather than "J2000", so I needed to set "J2000" for the individual steps, and then rotate back to `self.reference_frame` at the end.

tests/pytests/test_cassis_drivers.py
```diff
@@ -0,0 +1,292 @@
...
+    def setUp(self):
+      label = get_image_label("CAS-MCO-2016-11-26T22.32.14.582-RED-01000-B1", "isis")
+      self.driver = TGOCassisIsisLabelNaifSpiceDriver(label)
+    
+    def test_short_mission_name(self):
+      assert self.driver.short_mission_name == "tgo"
+    
+    def test_exposure_duration(self):
+      assert self.driver.exposure_duration == 1.920e-003
+
```
kberryUSGS: I'm not sure that these are very useful; would be happy to remove.

ale/drivers/tgo_drivers.py
```diff
@@ -0,0 +1,65 @@
...
+        the same ID the Spice expects in bods2c calls.
+        Expects instrument_id to be defined in the Pds3Label mixin. This should
+        be a string of the form CaSSIS
+
+        Returns
+        -------
+        : str
+          instrument id
+        """
+        return "TGO_CASSIS"
```
jessemapel: You should map the label value to this so that it fails on an invalid label.

tests/pytests/test_cassis_drivers.py
```diff
@@ -0,0 +1,292 @@
+import pytest
+import ale
+import os
+import json
+import pvl
+
+import numpy as np
+from ale.drivers import co_drivers
```
jessemapel: This looks like an old import that is not needed

tests/pytests/test_cassis_drivers.py
jessemapel: These should test the functions implemented in the driver: instrument_id and ephemeris_start_time

ale/base/data_naif.py
kberryUSGS: Just to confirm -- I tried this and it didn't work.

PR #296, opened 2019-10-11
UNFINISHED lronac isis driver work.

paarongiroux:
loads tests is currently getting 

```
SPICE(FRAMEDATANOTFOUND) --

PCK data required to compute the orientation of the body-fixed frame MOON_PA_DE421 for epoch
2009 JUL 30 12:21:44.368 TDB were not found. If these data were to be provided by a binary PCK 
file, then it is possible that the PCK file does not have coverage for the specified body-fixed frame at
 the time of interest
```

I tried re-slicing the kernels, but I still got the same error message.

When the test is run using the original unsliced binary kernels, I no longer get this message, but it fails during some frame chain operations.


Kelvinrr:
The problem was there a kernel missing containing info on MOON_PA_DE421 that was missing from the kernel list. With those, an ISD can be instantiated.

ale/drivers/lro_drivers.py
```diff
@@ -220,3 +221,202 @@ def sampling_factor(self):
...
+        """
+        The center of the CCD in detector pixels
+        ISIS uses 0.5 based CCD samples, so we need to convert to 0 based.
+
+        Returns
+        -------
+        float :
+            The center sample of the CCD
+        """
+        return super().detector_center_sample - 0.5
```
kberryUSGS: Does this make sense to have as a default at the `IsisLabel` level?

jessemapel: This stuff is inconsistent between instruments. Some missions use ISIS pixels, some use 0 based pixels.

kberryUSGS: By "ISIS Pixels" here, we just mean 0.5-based? I was looking at the comment and not what the code actually does and taking it too literally: "ISIS uses 0.5-based so we need to convert to 0-based" to me implies that we're somehow converting from "ISIS pixels" to "0-based pixels" but of course there are no real, literal "ISIS Pixels" here, it's just that the IK for the instrument happens to use 0.5-based pixels and we need 0-based pixels. Right?

jessemapel: @kberryUSGS yes that is correct. This is converting whatever the IK uses, not what ISIS uses. Maybe a better comment is in order.

kberryUSGS: @jessemapel I don't think that an updated comment needs to be a change on this PR -- this comment is used in many places -- this could be a ticket to either update wording or define "ISIS pixels" to mean 0.5-based pixels in general somewhere.

PR #297, opened 2019-10-11
MEX Drivers and Tests

tthatcher95:


ale/drivers/mex_drivers.py
```diff
@@ -422,3 +495,85 @@ def sensor_model_version(self):
...
+
+        @property
+        def sensor_model_version(self):
+            return 1
+
+        @property
+        def line_scan_rate(self):
+            isis_bytes = read_table_data(self.label['Table'], self._file)
+            times_table = parse_table(self.label['Table'], isis_bytes)
+            start_time = times_table['EphemerisTime'][0]
```
acpaquette: Remove start_time here, doesn't look like it's being used.

ale/drivers/mex_drivers.py
```diff
@@ -422,3 +495,85 @@ def sensor_model_version(self):
           ISIS sensor model version
         """
         return 1
+
+class MexHrscIsisLabelNaifSpiceDriver(LineScanner, IsisLabel, NaifSpice, RadialDistortion, Driver):
+
+        @property
+        def instrument_id(self):
+            if(super().instrument_id != "HRSC"):
+                raise Exception ("Instrument ID is wrong.")
```
acpaquette: I know this isn't necessarily part of this PR but if we could add this to the Pds3LabelNaifSpice driver for Mex it will save us some headache down the road. The property is on between line 159 and 170 in the mex_drivers.py file.

tests/pytests/test_mex_drivers.py
```diff
@@ -0,0 +1,728 @@
...
+
+@pytest.mark.parametrize("formatter", ['isis', 'usgscsm'])
+def test_mex_load(test_kernels, formatter, usgscsm_compare_dict):
+
+    if formatter=="usgscsm":
+        label_file = get_image_label('h5270_0000_ir2', "pds3")
+
+        # Check to change the line based on ISIS vs PDS3
+        # This is due to a processing step in mroctx2isis that removes samples
+        # based on some flags on the label
```
acpaquette: Old MRO copy paste comment can be removed

tests/pytests/test_mex_drivers.py
```diff
@@ -0,0 +1,728 @@
...
+        with patch('ale.drivers.mex_drivers.MexHrscPds3NaifSpiceDriver.binary_ephemeris_times', \
+                   new_callable=PropertyMock) as binary_ephemeris_times, \
+            patch('ale.drivers.mex_drivers.MexHrscPds3NaifSpiceDriver.binary_exposure_durations', \
+                   new_callable=PropertyMock) as binary_exposure_durations, \
+            patch('ale.drivers.mex_drivers.MexHrscPds3NaifSpiceDriver.binary_lines', \
```
acpaquette: All of these patches should probably patch the base module like:

`ale.drivers.mex_drivers.binary_exposure_durations`

Unless this doesn't work in which case it can probably stay as is.

tests/pytests/test_mex_drivers.py
```diff
@@ -0,0 +1,728 @@
...
+    kernels = get_image_kernels('h5270_0000_ir2')
+    updated_kernels, binary_kernels = convert_kernels(kernels)
+    yield updated_kernels
+    for kern in binary_kernels:
+        os.remove(kern)
+
+@pytest.mark.parametrize("formatter", ['isis', 'usgscsm'])
+def test_mex_load(test_kernels, formatter, usgscsm_compare_dict):
+
+    if formatter=="usgscsm":
```
acpaquette: The formatter should not be tied to the label type here. Ideally we would want to test the both formatters on both sources. I know that the Mex drivers aren't that mature so we should probably still test the combinations but with expected failures

ale/drivers/mex_drivers.py
```diff
@@ -422,3 +495,85 @@ def sensor_model_version(self):
...
+        def ephemeris_start_time(self):
+            isis_bytes = read_table_data(self.label['Table'], self._file)
+            times_table = parse_table(self.label['Table'], isis_bytes)
+            start_time = times_table['EphemerisTime'][0]
+            return start_time
+
+        @property
+        def ephemeris_stop_time(self):
+            isis_bytes = read_table_data(self.label['Table'], self._file)
+            times_table = parse_table(self.label['Table'], isis_bytes)
```
jessemapel: Calling these two methods at the start of each methods is clunky. Make these into properties that store the value, similar to the PDS driver.

ale/drivers/mex_drivers.py
```diff
@@ -422,3 +495,85 @@ def sensor_model_version(self):
...
+
+            For HRSC, the center ephemeris time is calculated with the ephemeris stop time,
+            which is calculated from the binary image data.
+
+            Returns
+            -------
+            : float
+              Center ephemeris time
+            """
+            return (self.ephemeris_stop_time + self.ephemeris_start_time) / 2
```
jessemapel: This is already implemented in base https://github.com/USGS-Astrogeology/ale/blob/master/ale/base/base.py#L308

ale/drivers/mex_drivers.py
```diff
@@ -323,75 +326,145 @@ def line_scan_rate(self):
...
+        For HRSC, the center ephemeris time is calculated with the ephemeris stop time,
+        which is calculated from the binary image data.
+
+        Returns
+        -------
+        : float
+          Center ephemeris time
+        """
+        return (self.ephemeris_stop_time + self.ephemeris_start_time) / 2
+
```
jessemapel: This is already implemented in base https://github.com/USGS-Astrogeology/ale/blob/master/ale/base/base.py#L308

tests/pytests/test_mex_drivers.py
kberryUSGS: This didn't work, so I left it as is.

tests/pytests/test_mex_drivers.py
kberryUSGS: I'm not sure if I handled this in the best way, but the outputs were so off/non-functional for isislabel/usgscsm formatter and pds3label/isis3 formatter, that I just made these xfails.

PR #298, opened 2019-10-12
Spiceinit Fixes

acpaquette:
Changes the driver sort to return IsisLabelNaifSpice drivers before IsisLabelIsisSpice drivers. 

Also adds a lookup table to the dawn drivers to prevent them from succeeding on the wrong image types.

### DIFF
--- a/ale/base/data_naif.py
+++ b/ale/base/data_naif.py
@@ -295,14 +295,20 @@ class NaifSpice():
         : (sun_positions, sun_velocities)
           a tuple containing a list of sun positions, a list of sun velocities
         """
-        sun_state, _ = spice.spkezr("SUN",
-                                     self.center_ephemeris_time,
-                                     self.reference_frame,
-                                     'LT+S',
-                                     self.target_name)
-        positions = 1000 * np.asarray([sun_state[:3]])
-        velocities = 1000 * np.asarray([sun_state[3:6]])
-        times = np.asarray([self.center_ephemeris_time])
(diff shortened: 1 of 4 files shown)


ale/base/data_naif.py
```diff
@@ -295,14 +295,20 @@ def sun_position(self):
...
+        for time in times:
+            sun_state, _ = spice.spkezr("SUN",
+                                         time,
+                                         self.reference_frame,
+                                         'LT+S',
+                                         self.target_name)
+            positions.append(sun_state[:3])
+            velocities.append(sun_state[3:6])
+        positions = 1000 * np.asarray(positions)
+        velocities = 1000 * np.asarray(velocities)
```
jessemapel: This looks like it does almost the exact same thing. Is there a particular reason why?

ale/base/data_naif.py
```diff
@@ -295,14 +295,20 @@ def sun_position(self):
...
-                                     self.center_ephemeris_time,
-                                     self.reference_frame,
-                                     'LT+S',
-                                     self.target_name)
-        positions = 1000 * np.asarray([sun_state[:3]])
-        velocities = 1000 * np.asarray([sun_state[3:6]])
-        times = np.asarray([self.center_ephemeris_time])
+        times = [self.center_ephemeris_time]
+        positions = []
+        velocities = []
```
kberryUSGS: Is there an advantage to initializing these up here and then appending to them later in python?

PR #302, opened 2019-10-17
Lro updates

jessemapel:
LRO NAC was using IAU_MOON instead of MOON_ME for it's target reference frame. This fixes that.


tests/pytests/test_data_naif.py
```diff
@@ -69,7 +69,7 @@ def test_target_body_radii(self):
         np.testing.assert_array_equal(self.driver.target_body_radii, [3396.19, 3396.19, 3376.2 ])
 
     def test_reference_frame(self):
-        assert self.driver.reference_frame == 'IAU_Mars'
+        assert self.driver.reference_frame.upper() == 'IAU_MARS'
```
jessemapel: Names are not case-sensitive for NAIF so making this uppercase is safe.

scsides: yes

jessemapel:
[Some day my Mac jobs will fire](https://youtu.be/0niwn2pOEno?t=46)

PR #303, opened 2019-10-18
Removed subwindowing code from MDIS drivers

jessemapel:
This was ported over from ISIS, but was never actually used in ISIS, so it is being removed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,7 +7,7 @@
 # Specify the required version of CMake.
 # cmake 3.10 required for ctest/gtest integration
 cmake_minimum_required(VERSION 3.10)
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -7,7 +7,7 @@
 # Specify the required version of CMake.
 # cmake 3.10 required for ctest/gtest integration
 cmake_minimum_required(VERSION 3.10)
-project(ale VERSION 0.6.0 DESCRIPTION "Abstraction Library for Ephemerides ")
+project(ale VERSION 0.6.2 DESCRIPTION "Abstraction Library for Ephemerides ")
```
jessemapel: Since this morning, we decided to pull off bug release with each change like this. So, this will be 0.6.2 because the LROC change was 0.6.1, but I didn't change the version number.

PR #305, opened 2019-10-21
Updated HRSC start sample and line rate table

jessemapel:
Changes:

* Removed the detector start sample offset because it isn't used in the ISIS sensor model and the HRSC mission team has verified the ISIS sensor model
* Removed extraneous entries in the line scan rate table. This will make the ISD files smaller and the sensor model a tiny bit faster as it doesn't have to search through as much.

### DIFF
--- a/ale/drivers/mex_drivers.py
+++ b/ale/drivers/mex_drivers.py
@@ -321,6 +321,11 @@ class MexHrscPds3NaifSpiceDriver(LineScanner, Pds3Label, NaifSpice, RadialDistor
         For HRSC, the ephemeris times and exposure durations are
         stored in the image data.
 
+        In the image, every line has an entry. This method goes through
+        and removes conescutive lines with the same exposure duration.
+        There are also potentially missing lines in the image which have
+        to be accounted for. 
+
         Returns
         -------
         : list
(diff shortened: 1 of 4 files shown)


ale/drivers/mex_drivers.py
```diff
@@ -336,8 +326,17 @@ def line_scan_rate(self):
         : list
           Line scan rates
         """
-        times = [time - self.center_ephemeris_time for time in self.binary_ephemeris_times]
-        return (self.binary_lines, times, self.binary_exposure_durations)
+        relative_times = [time - self.center_ephemeris_time for time in self.binary_ephemeris_times]
```
scsides: Suggest adding documentation on what and why this is so much more than the simple case. Reading the code without the knowledge of why (unless it is obvious) forces the next programmer to go looking for documentation before they can decide if it is working as intended.

kberryUSGS: Agreed that some documentation would be good here!

ale/drivers/mex_drivers.py
```diff
@@ -324,7 +324,7 @@ def line_scan_rate(self):
         In the image, every line has an entry. This method goes through
         and removes conescutive lines with the same exposure duration.
         There are also potentially missing lines in the image which have
-        to be accounted for. 
+        to be accounted for.
```
scsides: Suggest adding doc for how the missing lines are detected

PR #306, opened 2019-10-23
First pass at Kaguya MI driver

jessemapel:
Turns out a lot of this is the same as Kaguya TC 🎉 

This still needs to be properly tested by projecting some output and looking at it, but all values appear to be reasonable.


ale/drivers/selene_drivers.py
```diff
@@ -216,27 +216,6 @@ def detector_center_sample(self):
...
-        -------
-        : str
-          Reference frame
-        """
-        if self.target_name.lower() == "moon":
-            return "MOON_ME"
-        else:
-            # TODO: How do we handle no target?
-            return "NO TARGET"
-
```
jessemapel: Changes from #302 mean this is no longer needed. It will now pull whatever your kernels say should be used for the Moon

ale/drivers/selene_drivers.py
```diff
@@ -435,3 +414,318 @@ def sensor_model_version(self):
...
+        No NAIF code exists for the spacecraft name 'SELENE-M.'  The NAIF code
+        exists only for 'SELENE' or 'KAGUYA' -- 'SELENE' is captured as
+        'MISSION_NAME'
+
+        Returns
+        -------
+        : str
+          mission name
+        """
+        return self.label.get('MISSION_NAME')
```
jessemapel: This is a little different from Kaguya TC. MI labels do not have the keyword that TC does, or a keyword with the same value. This returns the full name of the spacecraft including spaces.

ale/drivers/selene_drivers.py
```diff
@@ -435,3 +414,318 @@ def sensor_model_version(self):
...
+        : dict
+          radial distortion model
+
+        """
+        return {
+            "kaguyatc": {
+                "x" : self._odkx,
+                "y" : self._odky
+            }
+        }
```
jessemapel: Needs to be tested, but based on Hiro's gist, TC and MI appear to use the same equations for the distortion model. Maybe we should rename this to the Kaguya distortion model instead of Kaguya TC

tests/pytests/test_kaguya_drivers.py
```diff
@@ -14,13 +14,203 @@
...
+            'interpolation_method': 'lagrange',
+            'line_scan_rate': [[0.5, -1.300000011920929, 0.006500000000000001]],
+            'starting_ephemeris_time': 292234259.82293594,
+            'center_ephemeris_time': 292234261.12293595,
+            't0_ephemeris': -1.300000011920929,
+            'dt_ephemeris': 0.5200000047683716,
+            't0_quaternion': -1.300000011920929,
+            'dt_quaternion': 0.5200000047683716
+        }
+    },
```
jessemapel: Old image load test converted to use the more generic load parameterization

tests/pytests/test_kaguya_drivers.py
```diff
@@ -14,13 +14,203 @@
...
+            "name_model": "USGS_ASTRO_LINE_SCANNER_SENSOR_MODEL",
+            "interpolation_method": "lagrange",
+            "line_scan_rate": [[0.5, -6.239987522363663, 0.012999974000000001]],
+            "starting_ephemeris_time": 261664552.50899568,
+            "center_ephemeris_time": 261664558.7489832,
+            "t0_ephemeris": -6.239987522363663,
+            "dt_ephemeris": 0.8914267889090947,
+            "t0_quaternion": -6.239987522363663,
+            "dt_quaternion": 0.8914267889090947
+        }
```
jessemapel: This ISD is purely the current driver output. It has not be vetted yet.

tests/pytests/test_kaguya_drivers.py
```diff
@@ -80,9 +270,6 @@ def test_detector_center_sample(driver):
         assert driver.detector_center_sample == 54320.5
         gdpool.assert_called_with('INS-12345_CENTER', 0, 2)
 
-def test_reference_frame(driver):
-    assert driver.reference_frame == 'MOON_ME'
-
```
jessemapel: Code is no longer in the driver so the test can go

ale/drivers/selene_drivers.py
```diff
@@ -435,3 +414,318 @@ def sensor_model_version(self):
...
+        Expects tc_id to be defined. This should be a string of the form
+        LISM_TC1 or LISM_TC2.
+
+        Returns
+        -------
+        : list
+          focal plane to detector samples
+        """
+        pixel_size = spice.gdpool('INS{}_PIXEL_SIZE'.format(self.ikid), 0, 1)[0]
+        return [0, 0, -1/pixel_size]
```
jessemapel: From IK, Y- is cross-track

ale/drivers/selene_drivers.py
```diff
@@ -435,3 +414,318 @@ def sensor_model_version(self):
...
+        Expects tc_id to be defined. This should be a string of the form
+        LISM_TC1 or LISM_TC2.
+
+        Returns
+        -------
+        : list
+          focal plane to detector lines
+        """
+        pixel_size = spice.gdpool('INS{}_PIXEL_SIZE'.format(self.ikid), 0, 1)[0]
+        return [0, 1/pixel_size, 0]
```
jessemapel: From IK, X+ id down-track

jessemapel:
Confirmed that the four image corners project to within 0.01 degrees of the values from the image label when projected onto the ellipsoid.

ale/drivers/selene_drivers.py
scsides: The + seems odd, not that it matters, but did they not fly both directions?

ale/drivers/selene_drivers.py
```diff
@@ -435,3 +414,318 @@ def sensor_model_version(self):
...
+
+class KaguyaMiPds3NaifSpiceDriver(LineScanner, Pds3Label, NaifSpice, Driver):
+    """
+    Driver for a PDS3 Kaguya Multiband Imager (Mi) images. Specifically level2b2 Vis and Nir images.
+
+    NOTES
+    -----
+
+    * Kaguaya has adjusted values for some of its keys, usually suffixed with `CORRECTED_`.
+      These corrected values should always be preffered over the original values.
```
kberryUSGS: Small typo: preffered -> preferred

ale/drivers/selene_drivers.py
```diff
@@ -435,3 +414,318 @@ def sensor_model_version(self):
...
+          The detector sample of the principle point
+        """
+        return spice.gdpool('INS{}_CENTER'.format(self.ikid), 0, 2)[0] - 0.5
+
+    @property
+    def focal2pixel_samples(self):
+        """
+        Calculated using 1/pixel pitch
+        Expects tc_id to be defined. This should be a string of the form
+        LISM_TC1 or LISM_TC2.
```
kberryUSGS: TC->MI as well?

ale/drivers/selene_drivers.py
jessemapel: I think they only flew in one direction and that the same part of the spacecraft was always "forward".

jessemapel: Here's the ascii art from the IK
```

   The direction along the row of pixels of each band is defined to be 
   Y axis. The +Y is the direction toward the radiator side from the 
   detectors. Z axis is defined to be "ideal" boresight. 


                       Corner pixel
                      view direction 
                           ^
         Center pixel       \             X(Along-track)
            view   <.        \           ^   CCD plane
          direction  `-.      \          |
                        `-.    \         |     .^ -Y(Cross-track)
                           `-.  \        |   .'
                              `-.\    F  | .'
                     <-----------`o.-----o'  Detector line in
                   Z        Focal  \`-.  | .  the CCD plane
                            point   \  `-*'
                                     \ .'| Center pixel
                                      *  |
                                  Corner
                                   pixel           
```

jessemapel:
I compared his image with a projected Kaguya TC image and the global Kaguya Mosaic. Both seem to be off the global kaguya mosaic by about the same amount, 2500 meters.

jessemapel:
#306 needs to be merged before this. I will then merge the two changes as it will have merge conflicts with this.

jessemapel:
I've gotta head to Phoenix, but I'll get this cleaned up tonight or tomorrow.

PR #309, opened 2019-10-30
Updated Kaguya driver for LISM distortion

jessemapel:
This adds the boresight focal plane coordinates that we were missing from before.

This goes along with a PR on usgscsm, https://github.com/USGS-Astrogeology/usgscsm/pull/246


jessemapel:
Once https://github.com/USGS-Astrogeology/usgscsm/pull/246 is ready to merge, I will merge both at the same time to keep things in sync.

PR #311, opened 2019-11-22
Adds Rotation class

jessemapel:
This adds the base level Rotation object. It still needs docs and a little clean-up, but I wanted it up incase anyone wanted to work on it.

### DIFF
--- a/include/Rotation.h
+++ b/include/Rotation.h
@@ -30,6 +30,7 @@ namespace ale {
       // Type specific accessors
       std::vector<double> toQuaternion() const;
       std::vector<double> toRotationMatrix() const;
+      std::vector<double> toStateRotationMatrix(const std::vector<double> &av) const;
       std::vector<double> toEuler(const std::vector<int>& axes) const;
       std::pair<std::vector<double>, double> toAxisAngle() const;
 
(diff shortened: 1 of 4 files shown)


src/Rotation.cpp
```diff
@@ -0,0 +1,348 @@
...
+        }
+        quat = Eigen::Quaterniond::Identity();
+
+        for (size_t i = 0; i < angles.size(); i++) {
+          quat *= Eigen::Quaterniond(Eigen::AngleAxisd(angles[i], axis(axes[i])));
+        }
+      }
+
+
+      Impl(const std::vector<double>& axis, double theta) {
```
Kelvinrr: Axis and angle seem swapped compared to the above contractor.

include/Rotation.h
```diff
@@ -0,0 +1,81 @@
...
+
+      // Generic rotation operations
+      std::vector<double> operator()(const std::vector<double>& vector, const std::vector<double>& av = {0.0, 0.0, 0.0}) const;
+      Rotation inverse() const;
+      Rotation operator*(const Rotation& rightRotation) const;
+      Rotation interpolate(const Rotation& nextRotation, double t, RotationInterpolation interpType) const;
+
+    private:
+      // pimpl
+      class Impl;
```
Kelvinrr: Why make this a separate class?

include/Rotation.h
```diff
@@ -0,0 +1,81 @@
...
+      Rotation();
+      Rotation(double w, double x, double y, double z);
+      Rotation(const std::vector<double>& matrix);
+      Rotation(const std::vector<double>& angles, const std::vector<int>& axes);
+      Rotation(const std::vector<double>& axis, double theta);
+      ~Rotation();
+
+      // Special member functions
+      Rotation(Rotation && other) noexcept;
+      Rotation& operator=(Rotation && other) noexcept;
```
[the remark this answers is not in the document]
jessemapel: This is a move assignment operator. It's used in compound expressions to prevent creating and deleting multiple objects when executing things like `a + (b * c)`. For rotations, this comes into play when chaining multiple rotations together.

include/Rotation.h
jessemapel: This hides the third party dependency from the API which helps isolate this better.

https://www.geeksforgeeks.org/pimpl-idiom-in-c-with-examples/

jessemapel: Mostly, this makes it so that down-stream projects don't need to include the Eigen headers.

src/Rotation.cpp
jessemapel: These are largely two separate things. The first angle + axes constructor takes Euler angles with explicit X, Y, Z axes. This constructor is for a rotations around an arbitrary axis (i.e. a 120 degree rotation around the axis 1, 1, 1).

I can swap these arguments so that they are more similar, but I don't want users to assume that the axes argument in the first constructor and axis argument in the second constructor are the same.

jessemapel:
This is now ready to go.

include/Rotation.h
Kelvinrr: If we want to not expose the Eigen headers, as far as I know we only require for the Eigen types/functions to exist in the CPP and the interface without the nested class doesn't really need to expose Eigen classes/functions.

src/Rotation.cpp
Kelvinrr: That seems more of a documentation/naming issue, swapping them isn't going to do much to make it seem they are different things.

PR #313, opened 2019-11-22
Removed old numba and quaternion deps

jessemapel:
We don't use this library anymore. All of our rotations are handled by scipy now.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -6,11 +6,11 @@ channels:
 
 dependencies:
   - cmake>=3.10
+  - pytest
   - eigen
   - gsl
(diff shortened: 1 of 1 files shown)


environment.yml
```diff
@@ -6,18 +6,17 @@ channels:
 
 dependencies:
   - cmake>=3.10
+  - pytest
```
jessemapel: Also, it doesn't make sense to have cmake and jupyter in here, but not pytest.

jessemapel:
note that I've already removed these dependencies from the feedstock build recipe too.

PR #314, opened 2019-11-25
ale::States implementation

kberryUSGS:
This is an implementation of ale::States.


kberryUSGS:
I updated state to have a vec3d position and vec3d velocity but I guess failed to update this PR.

kberryUSGS:
Updated so that @ladoramkershner can pull down and add tests for what exists while I work on `minimizeCache()`

src/States.cpp
```diff
@@ -0,0 +1,472 @@
...
+  /** Checks to see if an individual value, a, is within the doman specified by x **/
+  bool InsideDomain(const double a, std::vector<double> x) {
+    if(a + DBL_EPSILON < *min_element(x.begin(), x.end())) {
+      return false;
+    }
+    if(a - DBL_EPSILON > *max_element(x.begin(), x.end())) {
+      return false;
+    }
+    return true;
+  }
```
ladoramkershner: Where is DBL_EPSILON defined?

src/States.cpp
```diff
@@ -0,0 +1,472 @@
...
+
+  /** Find the index of the x-value in the data set that is just
+   * below the input value, a. If a is below the domain minimum,
+   * the method returns 0 as the lower index.  If a is above the
+   * domain maximum, it returns the second to last index of the
+   * data set, Size()-2, as the lower index. **/
```
ladoramkershner: If a is below the lower end of x, then the function returns the lowest index possible (0), which makes sense to me. Why does it not return the highest index value (Size()-1) when a is above the higher end x?

src/States.cpp
scsides: It's in the c++ header file:
<cfloat>  or <float.h>

ladoramkershner: Which cpp header file, aren't there a bunch? I also meant what is its value? Does it have some standard value in that header file?

scsides: For true c++ the header is: cfloat
The C header is: float.h

ladoramkershner: Ohhh that is what you meant, thank you.

include/States.h
```diff
@@ -0,0 +1,78 @@
...
+#define ALE_STATES_H
+
+#include<vector>
+#include <gsl/gsl_interp.h>
+ 
+namespace ale {
+
+enum interpolation {
+  /// Interpolate using linear interpolation
+  linear,
```
kaitlyndlee: This is definitely a personal preference, but I think it is easier to identify enum values when they are all caps, almost treating them like a constant.  After looking at ISIS, it doesn't seem like we have a standard for defining enums.

tests/ctests/StatesTest.cpp
```diff
@@ -0,0 +1,488 @@
...
+  EXPECT_NEAR(positions[0].y, 1.0, 1e-10);
+  EXPECT_NEAR(positions[0].z, 4.0, 1e-10);
+  EXPECT_NEAR(positions[3].x, 7.0, 1e-10);
+  EXPECT_NEAR(positions[3].y, 4.0, 1e-10);
+  EXPECT_NEAR(positions[3].z, 1.0, 1e-10);
+}
+
+// minimizeCache
+// one test where the cache can be minimized
+// one test where the test cannot be minimized
```
scsides: test = cache?

src/States.cpp
scsides: Correct me if I'm wrong: The algorithm uses the two bonding values (n, n+1). So if my data set is: {1,2,3,4,5}, and I ask for 3.2 the algorithm needs (3 and 4) aka indexes (2 and 3). If the requested values is below the min, say -10, it needs (1 and 2) aka index (0 and 1) so it can extrapolate down. If the requested values is above the max, say 9, it needs the last two (4 and 5), aka index (3 and 4) to extrapolate up.

tests/ctests/StatesTest.cpp
```diff
@@ -0,0 +1,488 @@
...
+
+TEST(StatesTest, ConstructorPositionNoVelocity) {
+  std::vector<double> ephemTimes = {0.0, 1.0, 2.0, 3.0};
+  std::vector<ale::Vec3d> positions = {
+    ale::Vec3d(4.0, 1.0, 4.0),
+    ale::Vec3d (5.0, 2.0, 3.0),
+    ale::Vec3d (6.0, 3.0, 2.0),
+    ale::Vec3d (7.0, 4.0, 1.0)};
+
+  States noVelocityState(ephemTimes, positions, 1);
```
kaitlyndlee: Is it possible to parametrize these tests?

include/States.h
```diff
@@ -0,0 +1,78 @@
...
+  Vec3d position;
+  Vec3d velocity;
+
+  State(ale::Vec3d position, ale::Vec3d velocity) : position(position), velocity(velocity) {}; 
+  State() {};
+};
+
+class States {
+  public:
+    // Constructors
```
scsides: Questions for all:
Where and how are we going to document classes? class.h or class.cpp or ....?
What format are the comments in?

kberryUSGS: I second this question. I wasn't sure what to do, so I ended up using a confusing mixture of different formats, which is certainly not ideal.

kberryUSGS: @scsides I don't seem to be able to respond to your above comment about the algorithm in-line, but your understanding as described in your comment matches mine.

jessemapel: I prefer documenting in the header files because that is what I look at if I'm developing against something.

include/States.h
```diff
@@ -0,0 +1,78 @@
...
+};
+
+/** A 3D cartesian vector */
+struct Vec3d {
+  double x;
+  double y;
+  double z;
+
+  Vec3d(double x, double y, double z) : x(x), y(y), z(z) {};
+  Vec3d() : x(0.0), y(0.0), z(0.0) {};
```
jessemapel: It would be nice to have a vector or array constructor.

tests/ctests/StatesTest.cpp
```diff
@@ -0,0 +1,488 @@
...
+  EXPECT_NEAR(linear_velocity.x, -4.5, 1e-10);
+  EXPECT_NEAR(linear_velocity.y, -2, 1e-10);
+  EXPECT_NEAR(linear_velocity.z, -3.75, 1e-10);
+  EXPECT_NEAR(spline_velocity.x, -5.05, 1e-10);
+  EXPECT_NEAR(spline_velocity.y, -3.7125, 1e-10);
+  EXPECT_NEAR(spline_velocity.z, -1.7875, 1e-10);
+}
+
+// getState() and interpolateState() are tested when testing getPosition and
+// getVelocity, because they are derived from those methods
```
jessemapel: I'd like to see a separate getState test. Right now, the implementation of getPosition and getVelocity uses getState, but if that changes then it won't be tested.

jessemapel: I think interpolateState doesn't need a separate test.

tests/ctests/StatesTest.cpp
```diff
@@ -0,0 +1,488 @@
...
+  EXPECT_EQ(states.size(), 9);
+  ale::State catState = statesState.getState(-1.0); 
+  EXPECT_EQ(catState.position.x, -3.0);
+  vector<ale::State> states2 = statesState.getStates();
+  ale::State dogState = statesState.getState(7.0); 
+  EXPECT_EQ(states2.size(), 6);
+  EXPECT_EQ(dogState.position.x, -3.0);
+  EXPECT_EQ(dogState.velocity.x, 4.0);
+}
+*/
```
jessemapel: This commented out part can be removed.

include/States.h
scsides: Agreed, IMHO, all API docs go in the header. If I want to use a class, it is nice to not have all the code cluttering up what I'm looking for. All I should need to code against an API is the libraries and headers. If I'm looking for a bug, fine, I need the source, and good internal "WHY THIS IS..." comments.
FYI, this is opposite to the current ISIS style doc.

include/States.h
kaitlyndlee: We define an interpolation enum in ale.h. Should we redefine the enum in states.h or use the same enum from ale.h?

kberryUSGS: @kaitlyndlee I moved this into States because I wasn't sure whether the content in ale.h was going to stay in there or not, so this is a good question.

src/States.cpp
```diff
@@ -0,0 +1,472 @@
...
+    int lowerIndex = ale::FindIntervalLowerIndex(a, x);
+
+    // if a is exactly equal to the time at either endpoint, return the 
+    // value at the appropriate endpoint
+    if(a == x[lowerIndex]) {
+      return deriv[lowerIndex];
+    }
+    if(a == x[lowerIndex+1]) {
+      return deriv[lowerIndex+1];
+    }
```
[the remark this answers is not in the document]
kberryUSGS: I suppose it doesn't save much time to skip calculating the value in this case? (Even though we know it will have the right value.

include/States.h
```diff
@@ -0,0 +1,78 @@
...
+  State() {};
+};
+
+class States {
+  public:
+    // Constructors
+    States();
+    States(std::vector<double> ephemTimes, std::vector<ale::Vec3d> positions, int refFrame=1);
+    States(std::vector<double> ephemTimes, std::vector<ale::Vec3d> positions, 
+           std::vector<ale::Vec3d> velocities, int refFrame=1);
```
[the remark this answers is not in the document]
kberryUSGS: Good point.

src/States.cpp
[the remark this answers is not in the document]
scsides: Assuming that most of the time the request is not for a value in the data set, then the "if", although fast, is slowing down the calculations for the majority of cases.

include/States.h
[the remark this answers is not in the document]
kberryUSGS: Actually, not sure how to best do this cleanly. The size of the velocity vector needs to be the same size as the positions vector to not break things, so even if I set a default value of {0.0, 0.0, 0.0}, the body of the constructor would need to create a list of these the same size as the positions vector.

src/States.cpp
```diff
@@ -0,0 +1,472 @@
...
+  }
+
+  //! Returns the current positions inside the m_states object
+  std::vector<ale::Vec3d> States::getPositions() const {
+    // extract positions from state vector
+    std::vector<ale::Vec3d> positions; 
+    
+    for(ale::State state : m_states) {
+        ale::Vec3d position = state.position;
+        positions.push_back(position);
```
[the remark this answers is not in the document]
kberryUSGS: Not sure why I left that as 2 lines, thanks.

src/States.cpp
```diff
@@ -0,0 +1,472 @@
...
+    inputIndices.push_back(currentSize);
+
+    // find all indices needed to make a hermite table within the appropriate tolerance
+    std::vector <int> indexList = HermiteIndices(tolerance, inputIndices, baseTime, timeScale);
+
+    // Remove all lines from cache vectors that are not in the index list
+    for(int i = currentSize; i >= 0; i--) {
+      if(!std::binary_search(indexList.begin(), indexList.end(), i)) {
+        m_states.erase(m_states.begin() + i);
+        m_ephemTimes.erase(m_ephemTimes.begin() + i);
```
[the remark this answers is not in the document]
kberryUSGS: I agree, especially since for many of these, the initial # of indices is over 1,000, and the ultimate index list often only has a few entries.

tests/ctests/StatesTest.cpp
kberryUSGS: getState is tested right now in the isis comparison. Is that sufficient?

tests/ctests/StatesTest.cpp
```diff
@@ -0,0 +1,488 @@
...
+  EXPECT_NEAR(states[3].position.y, 4.0, 1e-10);
+  EXPECT_NEAR(states[3].position.z, 1.0, 1e-10);
+  EXPECT_NEAR(states[3].velocity.x, -7.0, 1e-10);
+  EXPECT_NEAR(states[3].velocity.y, -4.0, 1e-10);
+  EXPECT_NEAR(states[3].velocity.z, -1.0, 1e-10);
+}
+
+class TestState : public ::testing::Test {
+protected:
+  States *states;
```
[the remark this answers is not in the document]
kberryUSGS: I couldn't get it to work without a pointer. I needed to be able to set positions and velocities, but there deliberately aren't setters in this class, so the only  way I could figure out how to make this accessible in the test was to make this a pointer and then create a new object in the `Setup`. Is there a better way to handle this?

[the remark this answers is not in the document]
jessemapel: You should be able to have it default construct and then assign over it.

src/States.cpp
```diff
@@ -0,0 +1,404 @@
...
+      }
+
+      double sTime = (time - baseTime) / timeScale;
+      position.x = ale::evaluateCubicHermite(sTime, vxs, scaledEphemTimes, xs);
+      position.y = ale::evaluateCubicHermite(sTime, vys, scaledEphemTimes, ys);
+      position.z = ale::evaluateCubicHermite(sTime, vzs, scaledEphemTimes, zs);
+
+      velocity.x = ale::evaluateCubicHermiteFirstDeriv(sTime, vxs, scaledEphemTimes, xs);
+      velocity.y = ale::evaluateCubicHermiteFirstDeriv(sTime, vys, scaledEphemTimes, ys);
+      velocity.z = ale::evaluateCubicHermiteFirstDeriv(sTime, vzs, scaledEphemTimes, zs);
```
jessemapel: See my comments above about changing these functions to operate on state objects.

src/States.cpp
```diff
@@ -0,0 +1,404 @@
...
+    h = x1 - x0;
+    t = (interpTime - x0) / h;
+    if(h != 0.0) {
+      return ((6 * t * t - 6 * t) * y0 + (3 * t * t - 4 * t + 1) * h * m0 + (-6 * t * t + 6 * t) * y1 + (3 * t * t - 2 * t) * h * m1) / h;
+    }
+    else {
+      throw std::invalid_argument("Error in evaluating cubic hermite velocities, values at"
+                                  "lower and upper indicies are exactly equal.");
+    }
+  }
```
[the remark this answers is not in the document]
jessemapel: For bonus perks, we could inline the first function as it's just doing addition and multiplication.

src/States.cpp
```diff
@@ -0,0 +1,404 @@
...
+
+  double States::getStopTime() {
+    int len = m_ephemTimes.size(); 
+    return m_ephemTimes.back();
+  }
+      
+
+  void States::minimizeCache(double tolerance) {
+    if (m_states.size() == 1) {
+      throw std::invalid_argument("Cache size is 1, cannot minimize");
```
jessemapel: I think size needs to be greater than 2, we cannot evaluate a spline with 1 point.

src/States.cpp
```diff
@@ -0,0 +1,404 @@
...
+      ys.push_back(state.position.y);
+      zs.push_back(state.position.z);
+      vxs.push_back(state.velocity.x);
+      vys.push_back(state.velocity.y);
+      vzs.push_back(state.velocity.z);
+    }
+
+    ale::Vec3d position, velocity; 
+
+    if (!hasMinimizedCache()) {
```
jessemapel: This should be a switch statement that uses the passed in interpolation type. If linear is passed in then use the linear GSL interpolation. If spline is passed in then use the hermite spline interpolation.

tests/ctests/StatesTest.cpp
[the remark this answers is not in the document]
kberryUSGS: Ah, good point. I'll fix this.

src/States.cpp
kberryUSGS: @jessemapel Should we let the user ask for a linear interpolation if the cache has been reduced to within some tolerance specifically calculated for the hermite cache? Couldn't that result in worse-than-intended values?

jessemapel:
[Finish States Object](https://trello.com/c/0SJk8DR8/147-finish-states-object)

jessemapel:
Most of the comments on here are minor optimizations and can be addressed later. For right now, getting this large PR merged and getting everything in sync is the priority. We'll likely run into issues as we try to integrate this into USGSCSM and we'll address those then.

kberryUSGS:
@acpaquette If you want to push directly to this PR and don't have permission let me know and I'll add you as a contributor on my fork. Also, I'm available to answer any questions on any of this. I also checked and I don't have any additional work anywhere on this, everything I had is up on this PR.

acpaquette:
@jessemapel This should be good for another review. The coveralls decrease is due to vector size checks, I don't know if it's worth while writing tests for them to prevent the coverage decrease.

src/States.cpp
```diff
@@ -0,0 +1,291 @@
...
+    bool allZero = std::all_of(velocities.begin(), velocities.end(), [](ale::Vec3d vec)
+                               { return vec.x==0.0 && vec.y==0.0 && vec.z==0.0; });
+    return !allZero;
+  }
+
+
+  ale::State States::getState(double time, ale::interpolation interp) const {
+    std::vector<double> xs, ys, zs, vxs, vys, vzs;
+    int lowerBound = ale::interpolationIndex(m_ephemTimes, time);
+    std::vector<double> interpTimes = {};
```
jessemapel, line 109: Move the definitions of xs, ys, zs, vxs, vys, vzs, and interpTimes down to just before they are filled. It'd confusing to have them separated by the interpolation range logic.

src/States.cpp
```diff
@@ -0,0 +1,291 @@
...
+                  interpolateState(ys,  interpTimes, time, interp, 0),
+                  interpolateState(zs,  interpTimes, time, interp, 0)};
+
+      velocity = {interpolateState(vxs, interpTimes, time, interp, 0),
+                  interpolateState(vys, interpTimes, time, interp, 0),
+                  interpolateState(vzs, interpTimes, time, interp, 0)};
+    }
+    else if (interp == SPLINE && hasVelocity()){
+      // Do hermite spline if velocities are available
+      double baseTime = (m_ephemTimes.at(0) + m_ephemTimes.at(m_ephemTimes.size() - 1)) / 2.;
```
jessemapel, line 160: I would change this to `(interpTimes.front() + interpTimes.back()) / 2`. It won't change the output of the function, but it will make it a little more numerically stable by interpolating closer to 0.

src/States.cpp
```diff
@@ -0,0 +1,291 @@
...
+      double timeScale = 1.0;
```
jessemapel, line 161: It's also a good idea to change this to `(interpTimes.back() - interpTimes.front()) / 2`. Combined with the baseTime change, this will re-scale the time to be from -1 to 1.

src/States.cpp
```diff
@@ -0,0 +1,291 @@
...
+        scaledEphemTimes.push_back((interpTimes[i] - baseTime) / timeScale);
+      }
+      double sTime = (time - baseTime) / timeScale;
+      position.x = ale::evaluateCubicHermite(sTime, vxs, scaledEphemTimes, xs);
+      position.y = ale::evaluateCubicHermite(sTime, vys, scaledEphemTimes, ys);
+      position.z = ale::evaluateCubicHermite(sTime, vzs, scaledEphemTimes, zs);
+
+      velocity.x = ale::evaluateCubicHermiteFirstDeriv(sTime, vxs, scaledEphemTimes, xs);
+      velocity.y = ale::evaluateCubicHermiteFirstDeriv(sTime, vys, scaledEphemTimes, ys);
+      velocity.z = ale::evaluateCubicHermiteFirstDeriv(sTime, vzs, scaledEphemTimes, zs);
```
jessemapel, line 174: These are technically incorrect, but because our time scaling has been 1, it hasn't shown up yet. We're taking a derivative in scaled time (these values would be `dx/dt_s, dy/dt_s, dz/dt_s)`) so we have to do the chain rule to recover the derivative in regular time. If `t_s = (t - baseTime) / timeScale` then `dt_s/dt = 1 / timeScale` and `dx/dt = dx/dt_s * dt_s/dt = dx/dt_s * 1 / timeScale`.

tests/ctests/StatesTest.cpp
```diff
@@ -0,0 +1,444 @@
...
+      ale::Vec3d(4, 0, 0),
+      ale::Vec3d(5, 3, 0.064),
+      ale::Vec3d(6, 4, 0.512),
+      ale::Vec3d(7, 3, 1.728)};
+
+    std::vector<ale::Vec3d> velocities = {
+      ale::Vec3d(1, 4, 0),
+      ale::Vec3d(1, 2, 0.48),
+      ale::Vec3d(1, 0, 1.92),
+      ale::Vec3d(1, -2, 4.32)};
```
jessemapel, line 132: Can you add a comment with what polynomial functions you used for these so that if people need a different time range in the future they know what the values should be.

jessemapel:
@acpaquette I'm not concerned about the coverage decrease. Some of this code will be moved around once #334 is merged and we can clean up the testing then.

src/States.cpp
acpaquette: @jessemapel Is this happening for both the position and velocity values?

jessemapel: Just the velocity need to account for this.

acpaquette: Talked about this more over Zoom but since the velocity isn't being time scaled at all the time itself doesn't need to account for scaling. No need to worry about removing it

PR #315, opened 2019-11-26
ISD object without Rotation and State support. 

Kelvinrr:


Kelvinrr:
The only header we had before was ale.h, lowercase, Rotation made the headers mixed so I made it lower case.

jessemapel:
I would greatly prefer upper-case headers. We can leave `ale.h` for compatibility, but it's much more common in c++ to have Upper Camel Case header names.

tests/ctests/isd_tests.cpp
```diff
@@ -0,0 +1,679 @@
+#include <string>
```
jessemapel: This file should be renamed to IsdTests.cpp

tests/ctests/IsdTests.cpp
```diff
@@ -0,0 +1,679 @@
...
+#include "Util.h"
+
+void ASSERT_DOUBLE_VECTOR_EQ(std::vector<double> v1, std::vector<double> v2) {
+  if (v1.size() != v2.size()) {
+    EXPECT_TRUE(::testing::AssertionFailure() << "The two input vectors are different in size");
+  }
+
+  for(size_t i = 0; i < v1.size(); i++) {
+    if(v1[i] != v2[i]) {
+      EXPECT_TRUE(::testing::AssertionFailure() << "Arrays at " << i << " are not equal.");
```
jessemapel: I would use EXPECT_DOUBLE_EQUAL here

tests/ctests/IsdTests.cpp
```diff
@@ -0,0 +1,679 @@
...
+
+#include "gtest/gtest.h"
+
+#include "ale.h"
+#include "Isd.h"
+#include "Util.h"
+
+void ASSERT_DOUBLE_VECTOR_EQ(std::vector<double> v1, std::vector<double> v2) {
+  if (v1.size() != v2.size()) {
+    EXPECT_TRUE(::testing::AssertionFailure() << "The two input vectors are different in size");
```
jessemapel: I would assert true here. You could seg_fault in the for loop below if this is false.

tests/ctests/IsdTests.cpp
```diff
@@ -0,0 +1,679 @@
...
+      EXPECT_TRUE(::testing::AssertionFailure() << "The two input vectors at " << i << "are different in size");
+    }
+
+    for(size_t j = 0; j < v1[i].size(); j++) {
+      if(v1[i][j] != v2[i][j]) {
+        EXPECT_TRUE(::testing::AssertionFailure() << "Arrays at " << i << ", " << j <<  " are not equal.");
+      }
+    }
+  }
+}
```
jessemapel: Same as above, I would use ASSERT_TRUE for the size checks and EXPECT_DOUBLE_EQUAL for the comparison.

tests/ctests/IsdTests.cpp
```diff
@@ -0,0 +1,679 @@
...
+
+TEST(Isd, TransverseDistortion) {
+  ale::json trans;
+  trans["optical_distortion"]["transverse"]["x"] = {1};
+  trans["optical_distortion"]["transverse"]["y"] = {2};
+
+  std::vector<double> coeffs = ale::getDistortionCoeffs(trans);
+  EXPECT_EQ(ale::getDistortionModel(trans), ale::DistortionType::TRANSVERSE);
+  EXPECT_DOUBLE_EQ(coeffs[0], 1);
+  EXPECT_DOUBLE_EQ(coeffs[10], 2);
```
jessemapel: Before accessing anything at an index, you should assert the vector size is what you expect.

Kelvinrr:
We should probably also use hpp instead of `.h`. But it's better to just stick to something. Wasting effort to swap things 3 times over is silly and end up inconsistent is silly.

jessemapel:
:+1: to hpp. Should we just update everything now and fix the include in ISIS.

scsides:
> I would greatly prefer upper-case headers. We can leave `ale.h` for compatibility, but it's much more common in c++ to have Upper Camel Case header names.

I really don't have a big preference for either way, but I disagree with the "much more common in c++ to have Upper Camel Case header names". All lower case is 10 times more common.

PR #317, opened 2019-12-02
Fixed installing headers

jessemapel:
The PUBLIC_HEADER option only allows for a single file. This change ensures all of the headers are installed properly and the find_package config still works.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,16 +24,17 @@ find_package(nlohmann_json REQUIRED)
 add_library(ale SHARED
             ${CMAKE_CURRENT_SOURCE_DIR}/src/ale.cpp
             ${CMAKE_CURRENT_SOURCE_DIR}/src/Rotation.cpp)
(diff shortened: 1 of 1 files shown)


kberryUSGS:
Followup question: Where is it specified that PUBLIC_HEADER only allows for a single file? I found the following which seems to indicate that it allows multiple files if in quotes? http://cmake.3232098.n2.nabble.com/Using-PUBLIC-HEADER-property-to-install-headers-but-only-1st-one-is-installed-td7588373.html

And are there any other consequences to moving away from `PUBLIC_HEADER`, or is this update functionally equivalent (but with more than one header file working) ?

jessemapel:
@kberryUSGS I did some more research and the PUBLIC_HEADER property complains when I pass in a list variable. Also, the docs and what I read online say it's mostly for frameworks on OSX.

The only difference with this is that we can't be a framework right now, which is totally fine.

PR #318, opened 2019-12-05
🚧Added clang-tidy to linux ci🚧

jessemapel:


jessemapel:
nah I don't like this

PR #319, opened 2019-12-05
Add clang-format to enforce Google style

jessemapel:
This adds a check that all modified code conforms to the Google style format. Not necessarily the full style guide (for example we can throw exceptions).


jessemapel:
If we decide on a different style, we can always update the `.clang-format` file and the travis lines will still work.

jessemapel:
Closing this for now. If we decide to come back later to this it can be re-opened.

PR #325, opened 2019-12-11
Removed py finalize call

jessemapel:
This is causing assertion errors in Python threading under certain conditions. Python will already clean-up at exit, so I don't think we need to clean up the interpreter ourselves.


jessemapel:
Travis passed, it just didn't report back. I'm going to merge this to get Jenkins building.

PR #326, opened 2019-12-11
fikid vars now included in naif_keywords. 

Kelvinrr:
Mostly based on https://github.com/USGS-Astrogeology/ale/pull/321

Instruments with filters will want naif keywords with fikids in them (right now HRSC and messenger)

### DIFF
--- a/ale/base/data_naif.py
+++ b/ale/base/data_naif.py
@@ -535,8 +535,7 @@ class NaifSpice():
             self._naif_keywords = {**self._naif_keywords, **util.query_kernel_pool(f"*{self.ikid}*"),  **util.query_kernel_pool(f"*{self.target_id}*")}
 
             try:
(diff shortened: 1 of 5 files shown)


Kelvinrr:
Still looking into it, but it seems like the code for getting test kernels does not guarantee that IAKs get listed after IKs.

jessemapel:
We could add a sort into the kernel loading logic to ensure this.

Kelvinrr:
I think I would prefer keeping the kernel loading logic the same and make the image_2_data dict sorted on creation. I don't see how the kernel loading logic would have enough information to make such a sort unless they are labeled going in. The functionality for sorted labeled kernels exists but image2data is sorted in whatever order listdir comes up with. We can either update all the test data to have labeled kernels (I'd rather not keep the current testing setup for too long and rather not put too much effort into it) or simply make the addendum kernels always be last in the list.

PR #328, opened 2019-12-13
re-enabled mex

Kelvinrr:
This puts MEX back in as ale::load-able


Kelvinrr:
Kind of? The MEX test was failing due to small ALE vs ISIS differences. The only thing that wasn't a small change in some decimal was Solar Lon which was ISISNULL. Didn't feel comfortable saying "We don't support Solar Lon computations for now" so I decided to try to fix it, re-enabled and then update test data. 

So probably best to leave this open until https://github.com/USGS-Astrogeology/ISIS3/pull/3599 gets tested locally and merged.

PR #329, opened 2020-02-07
Kaguya and LRO Image Inversion Changes, and HRSC Update

acpaquette:
Updates the Kaguya TC and Lro NAC models to handle image inversions when the spacecraft is flying opposite of it's originally intended flight direction.

Also updates the MEX HRSC cameras starting detector line.

### DIFF
--- a/ale/drivers/lro_drivers.py
+++ b/ale/drivers/lro_drivers.py
@@ -130,6 +130,24 @@ class LroLrocPds3LabelNaifSpiceDriver(LineScanner, NaifSpice, Pds3Label, Driver)
         return super().detector_center_sample - 0.5
 
     @property
+    def focal2pixel_lines(self):
+        """
+        Expects ikid to be defined. This must be the integer Naif id code of
+        the instrument. For LROC NAC this is flipped depending on the spacecraft
+        direction.
+
+        Returns
+        -------
+        : list<double>
+          focal plane to detector lines
+        """
+        focal2pixel_lines = np.array(list(spice.gdpool('INS{}_ITRANSL'.format(self.ikid), 0, 3)))
+        if self.spacecraft_direction < 0:
+            return focal2pixel_lines
+        else:
+            return -focal2pixel_lines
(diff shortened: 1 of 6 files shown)


tests/pytests/test_kaguya_drivers.py
```diff
@@ -283,5 +283,5 @@ def test_focal2pixel_samples(driver):
 def test_focal2pixel_lines(driver):
     with patch('ale.drivers.selene_drivers.spice.gdpool', return_value=np.array([2])) as gdpool, \
          patch('ale.drivers.selene_drivers.spice.bods2c', return_value=-12345) as bods2c:
-        assert driver.focal2pixel_lines == [0, -1/2, 0]
+        assert driver.focal2pixel_lines == [0, 1/2, 0]
         gdpool.assert_called_with('INS-12345_PIXEL_SIZE', 0, 1)
```
jessemapel: You should test with and without the flipping flag.

acpaquette: :+1: I put these two checks under one test. It could be converted into two separate tests but I figured it was fine under one test, but I can change it if need be.

ale/drivers/lro_drivers.py
```diff
@@ -222,6 +240,34 @@ def sampling_factor(self):
...
+        Returns the x axis of the first velocity vector relative to the
+        spacecraft. This indicates of the craft is moving forwards or backwards.
+
+        From LROC Frame Kernel: lro_frames_2014049_v01.tf
+        "+X axis is in the direction of the velocity vector half the year. The
+        other half of the year, the +X axis is opposite the velocity vector"
+
+        Hence we rotate the first velocity vector into the sensor reference
+        frame, but the X component of that vector is inverted compared to the
+        spacecraft so a +X indicates backwards and -X indicates forwards
```
jessemapel: This is only true for the NACL reference frame, has this been tested on both left and right images?

PR #333, opened 2020-03-23
LRO Direction Fix + Summing Mode Fix

acpaquette:
Updated spacecraft direction for lro and  adds the necessary checks where needed. 

Also addresses an issue with the summing mode. The sampling factor from LRO needs to be accounted for in the focal2pixel translation in the new line scanner groundToImage algorithm in CSM.

### DIFF
--- a/ale/base/data_naif.py
+++ b/ale/base/data_naif.py
@@ -411,7 +411,9 @@ class NaifSpice():
                 v_vec = rotated_velocities
 
                 velocity_axis = 2
-                trans_x = self.focal2pixel_lines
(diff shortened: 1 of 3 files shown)


jessemapel:
@acpaquette What's wrong with the sampling factor? There should already be something in the ground to image to account for cross-track summing, are we getting that in the driver?

acpaquette:
There's nothing wrong with the sampling factor itself but it needs to be applied to the focal2pixel translation. I couldn't tell you exactly why but for images with a sampling factor > 1 you have to divide the focal2pixel line translation by the sampling factor

PR #334, opened 2020-03-27
ale::Orientation implementation

jessemapel:
This doesn't use the vectors defined by the new States in #314 and will need to be updated once that is done. For right now, this is functional and tested so it's ready to go.


jessemapel:
[Finish Orientations Object](https://trello.com/c/jCrNtHXl/148-finish-orientations-object)

PR #336, opened 2020-03-30
single isd format + mro test conversion

Kelvinrr:


ale/drivers/__init__.py
```diff
@@ -32,7 +33,8 @@
 __driver_modules__ = [importlib.import_module('.'+m, package='ale.drivers') for m in __all__]
 
 __formatters__ = {'usgscsm': to_usgscsm,
-                  'isis': to_isis}
+                  'isis': to_isis,
+                  'ale' : to_isd}
```
Kelvinrr: Seems like a way to get functionality for instruments without full support for both ISIS and CSM formatters is to simply make the singular ISD a new formatter and slowly start phasing out the other two.

jessemapel: It looks like to_isd is missing from the PR.

Kelvinrr:
Presumably, if this looks good, this is mergeable and we can place other PRs as we move through instruments.

jessemapel:
For the sake of getting to moving to the new format, I'm okay with merging this and then adding the test in another PR.

PR #341, opened 2020-04-01
Updated Dawn FC tests

Kelvinrr:


tests/pytests/test_dawn_drivers.py
```diff
@@ -62,12 +20,14 @@ def test_kernels():
...
+def test_load(test_kernels):
     label_file = get_image_label('FC21A0038582_15170161546F6F')
-    usgscsm_isd = ale.load(label_file, props={'kernels': test_kernels}, formatter='usgscsm')
-    assert compare_dicts(usgscsm_isd, usgscsm_compare_dict) == []
-
+    compare_dict = get_isd("dawnfc")
 
+    isd_str = ale.loads(label_file, props={'kernels': test_kernels})
+    isd_obj = json.loads(isd_str)
+    print(json.dumps(isd_obj, indent=2))
```
acpaquette, line 29: This only actually shows up if verbose is on right?

Kelvinrr: It shows if the test fails.

PR #342, opened 2020-04-02
Removed GSL and changed natural cubic spline to cubic poly interp

jessemapel:
This PR does a handful of things:

1. Removes GSL
1. Changes cubic spline interpolation into cubic polynomial fits.
1. Removes old interpolation code that is obsoleted by States and Rotation classes.

Test data updates are caused by changing from a natural cubic spline to cubic polynomials.


jessemapel:
It's also worth mentioning that this moves the lagrange interpolation the usgscsm does right now into ALE and modifies to work on irregularly spaced times.

src/States.cpp
```diff
@@ -105,29 +102,9 @@ namespace ale {
...
-      interpStart = lowerBound - 2;
-      interpStop = lowerBound + 1;
-    }
-    else {
-      interpStart = lowerBound - 1;
-      interpStop = lowerBound + 2;
-    }
+    // try to copy the surrounding 8 points as that's the most possibly needed
+    int interpStart = std::max(0, lowerBound - 3);
+    int interpStop = std::min(lowerBound + 4, (int) m_ephemTimes.size() - 1);
```
acpaquette, line 107: Only problem here is that at either the max or min index values you'll only get 4 values (specifically at when lowerBound is 0 or m_ephemTimes.size() - 1). I'm trying to think of doing something different but I can't think of anything right now.

src/ale.cpp
```diff
@@ -339,41 +174,36 @@ namespace ale {
...
+       order = 2;
+       break;
+     case SPLINE:
+       order = 4;
+       break;
+     case LAGRANGE:
+       order = 8;
+       break;
+     default:
+       throw invalid_argument("Invalid derivitive option, must be 0 or 1.");
```
acpaquette, line 192: Should throw a different error message as this is selecting interp type not derivative.

src/States.cpp
jessemapel: Lagrange interpolation generally gets less stable as you start adding points that are very far away. I'd be hesitant to use 6 points if the actual interpolation time is between the first two. I.E. interpolate at 0.5 and use points 0, 1, 2, 3, 4, 5

PR #343, opened 2020-04-02
Orientation 

acpaquette:
Includes everything from #334 and the necessary changes from #314. The other orientation PR isn't updating so I am opening this one.


jessemapel:
Closing and re-opening to get Travis to fire again.

include/Rotation.h
```diff
@@ -99,12 +102,14 @@ namespace ale {
...
        * Rotate a vector
        *
-       * @param vector The vector to rotate. Cab be a 3 element position or 6 element state.
+       * @param vector The vector to rotate. Can be a 3 element position or 6 element state.
        * @param av The angular velocity to use when rotating state vectors. Defaults to 0.
        *
        * @return The rotated vector.
        */
-      std::vector<double> operator()(const std::vector<double>& vector, const std::vector<double>& av = {0.0, 0.0, 0.0}) const;
+      ale::Vec3d operator()(const ale::Vec3d &av) const;
```
jessemapel, line 110: The name of this argument should be vector. It's just rotating a vector, not rotating specifically an angular velocity vector.

src/Orientations.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+
+  ale::Vec3d Orientations::interpolateAV(double time) const {
+    int interpIndex = interpolationIndex(m_times, time);
+    double t = (time - m_times[interpIndex]) / (m_times[interpIndex + 1] - m_times[interpIndex]);
+    ale::Vec3d interpAv = ale::Vec3d(linearInterpolate(m_avs[interpIndex], m_avs[interpIndex + 1], t));
+    return interpAv;
+  }
+
+  ale::Vec3d Orientations::rotateAvAt(
```
jessemapel, line 57: This should be rotate vector at. This is for rotating a generic vector, not an angular velocity vector.

src/Orientations.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+
+  ale::State Orientations::rotateStateAt(
+    double time,
+    const ale::State &state,
+    RotationInterpolation interpType,
+    bool invert
+  ) const {
+    Rotation interpRot = interpolate(time, interpType);
+    ale::Vec3d av;
```
jessemapel, line 75: This needs to default to 0, 0, 0 so that if there's no angular velocities, then it's just the rotation.

src/Rotation.cpp
```diff
@@ -204,29 +195,33 @@ namespace ale {
   }
 
 
-  std::vector<double> Rotation::operator()(
-        const std::vector<double>& vector,
-        const std::vector<double>& av
+  ale::Vec3d Rotation::operator()(const ale::Vec3d &vector) const {
+    std::vector<double> tempVec = {vector.x, vector.y, vector.z};
+    Eigen::Map<Eigen::Vector3d> eigenVector((double *)tempVec.data());
```
jessemapel, line 200: We don't need to use a Map here anymore. This can just make an Eigen::Vector3d object and then unpack it into our Vec3d object.

src/Rotation.cpp
```diff
@@ -204,29 +195,33 @@ namespace ale {
...
-    else {
-      throw std::invalid_argument("Vector to rotate is the wrong size.");
-    }
+    ale::Vec3d position = state.position;
+    ale::Vec3d velocity = state.velocity;
+    std::vector<double> positionVec = {position.x, position.y, position.z};
+    std::vector<double> velocityVec = {velocity.x, velocity.y, velocity.z};
+
+    Eigen::Map<Eigen::Vector3d> positionVector((double *)positionVec.data());
+    Eigen::Map<Eigen::Vector3d> velocityVector((double *)velocityVec.data());
```
jessemapel, line 216: Same here, no need for a std::vector and a Map

tests/ctests/RotationTest.cpp
```diff
@@ -217,152 +217,126 @@ TEST(RotationTest, ToAxisAngle) {
...
+  EXPECT_NEAR(rotatedVY.position.z, 0.0, 1e-10);
+  EXPECT_NEAR(rotatedVY.velocity.x, 0.0, 1e-10);
+  EXPECT_NEAR(rotatedVY.velocity.y, 0.0, 1e-10);
+  EXPECT_NEAR(rotatedVY.velocity.z, 1.0, 1e-10);
+  EXPECT_NEAR(rotatedVZ.position.x, 0.0, 1e-10);
+  EXPECT_NEAR(rotatedVZ.position.y, 0.0, 1e-10);
+  EXPECT_NEAR(rotatedVZ.position.z, 0.0, 1e-10);
+  EXPECT_NEAR(rotatedVZ.velocity.x, 1.0, 1e-10);
+  EXPECT_NEAR(rotatedVZ.velocity.y, 0.0, 1e-10);
+  EXPECT_NEAR(rotatedVZ.velocity.z, 0.0, 1e-10);
```
jessemapel, line 339: We will probably want some sort of vec3d comparison tool for testing. Not right now, but eventually.

src/Orientations.cpp
acpaquette: @jessemapel Should this take an arbitrary vector then, or use the Vector3d object?

tests/ctests/RotationTest.cpp
acpaquette: I know the Vec3d object is just a struct but could we still overload the equivalence operator?

jessemapel:
This looks good to me, but I'm confused why Travis is failing to get gcc 7 on linux? I'm good merging this if no one knows what's going on here.

jessemapel:
Merging in spite of Travis Linux not knowing what it's doing. Tests passed locally for me and Mac CI passed.

PR #344, opened 2020-04-03
Made Ale::load and Ale::loads optional and renamed files

jessemapel:
I moved all of the interpolation methods into InterpUtils.h and moved load(s) into Loads.h. I also made load(s) optional on the C++ side if you don't want to link against Python and all of the Python driver dependencies.

To not include load(s) set `-DALE_BUILD_LOAD=OFF` when configuring.


tests/ctests/CMakeLists.txt
```diff
@@ -1,10 +1,19 @@
...
-file(GLOB test_source "${CMAKE_SOURCE_DIR}/tests/ctests/*.cpp")
+set (ALE_TEST_SOURCE ${CMAKE_SOURCE_DIR}/tests/ctests/IsdTests.cpp
+                     ${CMAKE_SOURCE_DIR}/tests/ctests/OrientationsTests.cpp
+                     ${CMAKE_SOURCE_DIR}/tests/ctests/RotationTests.cpp
+                     ${CMAKE_SOURCE_DIR}/tests/ctests/StatesTests.cpp
+                     ${CMAKE_SOURCE_DIR}/tests/ctests/TestInterpUtils.cpp
+                     ${CMAKE_SOURCE_DIR}/tests/ctests/TestMain.cpp)
+
+if(ALE_BUILD_LOAD)
+  list(APPEND ALE_TEST_SOURCE ${CMAKE_SOURCE_DIR}/tests/ctests/LoadTests.cpp)
```
jessemapel, line 12: Because we only want to include the tests for load(s) if we build them, we can't just glob everything.

jessemapel:
A lot of this PR is just copying and pasting code around files. The CMakeList files are the bulk of the actual changes.

include/Util.h
```diff
@@ -8,7 +8,6 @@
 #include "Distortion.h"
 
 namespace ale {
-  using json = nlohmann::json;
```
jessemapel, line 11: Found another using in a header to clean out.

jessemapel:
[Ale compilation without linking](https://trello.com/c/sPaMM3fV/198-ale-compilation-without-linking)

include/Load.h
```diff
@@ -0,0 +1,14 @@
...
+#define ALE_INCLUDE_ALE_H
+
+#include <nlohmann/json.hpp>
+
+#include <string>
+
+namespace ale {
+  std::string loads(std::string filename, std::string props="", std::string formatter="usgscsm", bool verbose=true);
+
+  nlohmann::json load(std::string filename, std::string props="", std::string formatter="usgscsm", bool verbose=true);
```
acpaquette, line 11: Could we add some doc strings here?

jessemapel:
This is updated, but do not merge it until after #345 
I anticipate there will be merge conflicts and merging #345 first and then this is safer because it will prevent re-introducing the circular dependencies.

jessemapel:
This is ready for re-review now.

CMakeLists.txt
```diff
@@ -14,36 +14,48 @@ include(GNUInstallDirs)
...
-  set(Python_ROOT_DIR $ENV{CONDA_PREFIX})
-  set(Python_FIND_STRATEGY LOCATION)
+if(ALE_BUILD_LOAD)
+  # If there is an Anaconda environment activated, search that for Python first
+  if(EXISTS $ENV{CONDA_PREFIX})
+    message("Searching $ENV{CONDA_PREFIX} for Python libraries")
+    set(Python_ROOT_DIR $ENV{CONDA_PREFIX})
+    set(Python_FIND_STRATEGY LOCATION)
+  endif()
+  find_package(Python REQUIRED COMPONENTS Development)
```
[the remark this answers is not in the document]
jessemapel: The problem is whether you add the Python library as a dependency. Cmake won't even let you compile if it can't find the Python library and you add it as a dependency.

PR #345, opened 2020-04-07
Updated isd

Kelvinrr:
The update exposed some circular dependencies among headers so I had to shuffle a lot around.


src/Util.cpp
```diff
@@ -411,62 +457,125 @@ std::vector<double> ale::getDistortionCoeffs(json isd) {
...
-    for (auto &location : jayson.at("positions")) {
-      positions.push_back(location[0].get<double>());
-      positions.push_back(location[1].get<double>());
-      positions.push_back(location[2].get<double>());
+    for (auto &location : obj) {
+      positions.push_back(location.get<double>());
     }
   } catch (...) {
-    throw std::runtime_error("Could not parse the sun positions.");
+    throw std::runtime_error("Could not parse the sensor positions.");
```
jessemapel, line 468: This error message should just be generic about double arrays

src/Util.cpp
```diff
@@ -411,62 +457,125 @@ std::vector<double> ale::getDistortionCoeffs(json isd) {
...
-    for (auto &location : jayson.at("positions")) {
-      positions.push_back(location[0].get<double>());
-      positions.push_back(location[1].get<double>());
-      positions.push_back(location[2].get<double>());
+    for (auto &location : obj) {
+      Vec3d vec(location[0].get<double>(),location[1].get<double>(), location[2].get<double>() );
+      positions.push_back(vec);
     }
   } catch (...) {
     throw std::runtime_error("Could not parse the sensor positions.");
```
jessemapel, line 482: This error message also needs to be updated

include/InterpUtils.h
```diff
@@ -3,9 +3,25 @@
...
 
 namespace ale {
+
+  enum RotationInterpolation {
+    SLERP, // Spherical interpolation
+    NLERP // Normalized linear interpolation
+  };
+
+  /// Interpolation enum for defining different methods of interpolation
+  enum interpolation {
```
jessemapel, line 16: We should re-name this to PositionInterpolation to match the RotationInterpolation.

include/Isd.h
```diff
@@ -59,19 +60,17 @@ namespace ale {
...
 
-    //Positions sensor_pos;
-    //Positions sun_pos;
+    States inst_pos;
+    States sun_pos;
 
-    //Rotation sensor_orientation;
-    //Rotation body_orientaion;
+    Orientations inst_pointing;
+    Orientations body_rotation;
```
jessemapel, line 73: I think the easiest way to get the constant rotations in is to just add members for them and if they are present in the ISD set them to that, but otherwise set them to the identity rotation.

include/Isd.h
```diff
@@ -8,8 +8,9 @@
 #include "Util.h"
```
jessemapel, line 8: I think this can go into the source file.

src/InterpUtils.cpp
```diff
@@ -58,4 +59,182 @@ namespace ale {
     std::sort(merged.begin(), merged.end());
     return merged;
   }
+
+
+  // Temporarily moved over from States.cpp. Will be moved into interpUtils in the future.
```
jessemapel, line 64: Remove

tests/ctests/IsdTests.cpp
```diff
@@ -148,110 +150,124 @@ TEST(Isd, BadLogFile) {
...
+  ASSERT_EQ(sunPosObj.getReferenceFrame(), 2);
+  ASSERT_EQ(sunPosObj.getTimes()[0], 300);
+  
+  ASSERT_EQ(position[0].x, 10);
+  ASSERT_EQ(position[0].y, 12);
+  ASSERT_EQ(position[0].z, 13);
+
+  ASSERT_EQ(velocity[0].x, 0);
+  ASSERT_EQ(velocity[0].y, 0);
+  ASSERT_EQ(velocity[0].z, 0);
```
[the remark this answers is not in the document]
jessemapel: There is a very large difference between 0 velocity and unknown velocity

Kelvinrr:
Mac build failed for non-test reasons, can anyone with permissions restart the build (I can't seem to)?

Kelvinrr:
There a lot of changes are are benign API changes (like changes to camel case or placing "get" in front of getters) simply because the API was very inconsistent.

include/InterpUtils.h
```diff
@@ -3,9 +3,25 @@
...
 #include <vector>
 
-#include "Util.h"
+#include "Vectors.h"
 
 namespace ale {
+
+  enum RotationInterpolation {
+    SLERP, // Spherical interpolation
+    NLERP // Normalized linear interpolation
```
acpaquette, line 12: Can we explicitly define these?

include/Isd.h
```diff
@@ -59,19 +61,17 @@ namespace ale {
...
 
-    double t0_ephemeris_time;
-    double dt_ephemeris_time;
-
-    double t0_quaternions;
-    double dt_quaternions;
-
     json naif_keywords;
+    
+    PositionInterpolation interpMethod;
```
acpaquette, line 66: Should there also be a RotationInterpolation here?

include/Rotation.h
```diff
@@ -24,8 +19,7 @@ namespace ale {
        */
       Rotation();
       /**
-       * Construct a rotation from a quaternion.
-       *
+       * Construct a rotatiorotationsquaternrotations*
        * @param w The scalar component of the quaternion.
```
acpaquette, line 23: Revert comment change

src/Util.cpp
```diff
@@ -42,7 +67,7 @@ std::string ale::getPlatformName(json isd) {
   return name;
 }
 
-std::string ale::getLogFile(nlohmann::json isd) {
+std::string getLogFile(nlohmann::json isd) {
```
acpaquette, line 70: Should all of these that use json use nlohmann::json or just json?

src/Util.cpp
```diff
@@ -411,62 +457,145 @@ std::vector<double> ale::getDistortionCoeffs(json isd) {
...
+    json ipos = isd.at("instrument_position");
+    std::vector<Vec3d> positions = getJsonVec3dArray(ipos.at("positions"));
+    std::vector<double> times = getJsonArray<double>(ipos.at("ephemeris_times")); 
+    int refFrame = ipos.at("reference_frame").get<int>();
+    
+    bool hasVelocities = ipos.find("velocities") != ipos.end();
+    if (hasVelocities) {
+      std::vector<Vec3d> velocities = getJsonVec3dArray(ipos.at("velocities")); 
+      States states(times, positions, velocities, refFrame);  
+      return states;
```
acpaquette, line 499: Could move the return states outside of the if else blocks.

src/Util.cpp
```diff
@@ -1,8 +1,22 @@
...
+#include <algorithm>
+#include <iostream>
 
+#include "ale.h"
 #include "Util.h"
 
-std::string ale::getSensorModelName(json isd) {
+namespace ale {
+bool iequals(const std::string& a, const std::string& b)
+{
```
acpaquette, line 10: Can this get moved to the line above?

src/Util.cpp
```diff
@@ -1,8 +1,22 @@
 #include <stdexcept>
+#include <algorithm>
+#include <iostream>
 
+#include "ale.h"
```
acpaquette, line 5: This seems like it could be removed but I'm not sure. Does this use anything from ale.h? Or does it need it for the ale namespace?

ale/formatters/formatter.py
```diff
@@ -26,10 +26,10 @@ def to_isd(driver):
```
kberryUSGS, line 26: Should the docstring be updated to say something like the ISIS _and_ CSM compatible meta-data now, since I think it is now compatible with both?

include/Rotation.h
```diff
@@ -24,8 +19,7 @@ namespace ale {
        */
       Rotation();
       /**
-       * Construct a rotation from a quaternion.
-       *
+       * Construct a rotatiorotationsquaternrotations*
```
kberryUSGS, line 22: What was `rotatiorotationsquaternrotations` supposed to be?

include/States.h
```diff
@@ -79,7 +80,8 @@ class States {
    * Perform a cache reduction. After running this, getStates(), getPositions(),
    * and getVelocities() will return vectors of reduced size, and getState(),
    * getPosition(), and getVelocity() will
-   * returns values interpolated over the reduced cache using a cubic hermite spline
+   * returns values interpolated over the reduced cache using a cubic hermite splineGSW#|
+   * 1111111
```
kberryUSGS, line 84: Looks like something hit the keyboard here.

include/States.h
```diff
@@ -104,7 +106,7 @@ class States {
    */
     std::vector<int> hermiteIndices(double tolerance, std::vector <int> indexList,
                                     double baseTime, double timeScale);
-    std::vector<ale::State> m_states; //! Represent at states internally to keep pos, vel together
+    std::vector<State> m_states; //! Represent at states internally to keep pos, vel together
```
kberryUSGS, line 109: at -> as. I know this wasn't your typo. It was probably mine, but while you're in here, might as well fix it.

src/Util.cpp
```diff
@@ -411,62 +457,145 @@ std::vector<double> ale::getDistortionCoeffs(json isd) {
...
+std::vector<Rotation> getJsonQuatArray(json obj) {
+  std::vector<Rotation> quats;
   try {
-    json jayson = isd.at("sensor_position");
-    for (auto &location : jayson.at("positions")) {
-      positions.push_back(location[0].get<double>());
-      positions.push_back(location[1].get<double>());
-      positions.push_back(location[2].get<double>());
+    for (auto &location : obj) {
+      Rotation vec(location[0].get<double>(),location[1].get<double>(), location[2].get<double>(), location[2].get<double>() );
```
kberryUSGS, line 478: Should last index be 3, rather than 2?

src/Util.cpp
kberryUSGS, line 499: Could also just do return `States(times, positions, velocities, refFrame);` to save a line.

src/Util.cpp
```diff
@@ -411,62 +457,145 @@ std::vector<double> ale::getDistortionCoeffs(json isd) {
...
+    
+    bool hasVelocities = ipos.find("velocities") != ipos.end();
+    if (hasVelocities) {
+      std::vector<Vec3d> velocities = getJsonVec3dArray(ipos.at("velocities")); 
+      States states(times, positions, velocities, refFrame);  
+      return states;
+    }
+    else {
+      States states(times, positions, refFrame);
+      return states;
```
kberryUSGS, line 503: Could also just do `return States states(times, positions, refFrame);` to save a line.

src/Util.cpp
```diff
@@ -411,62 +457,145 @@ std::vector<double> ale::getDistortionCoeffs(json isd) {
...
-      quaternions.push_back(quaternion[3]);
+    json spos = isd.at("sun_position");
+    std::vector<Vec3d> positions = getJsonVec3dArray(spos.at("positions"));
+    std::vector<double> times = getJsonArray<double>(spos.at("ephemeris_times")); 
+    int refFrame = spos.at("reference_frame").get<int>();
+    bool hasVelocities = spos.find("velocities") != spos.end();
+    if (hasVelocities) {
+      std::vector<Vec3d> velocities = getJsonVec3dArray(spos.at("velocities")); 
+      States states(times, positions, velocities, refFrame);  
+      return states;
```
kberryUSGS, line 521: Why not just return `States(times, positions, velocities, refFrame);` ?

src/Util.cpp
```diff
@@ -411,62 +457,145 @@ std::vector<double> ale::getDistortionCoeffs(json isd) {
...
+    }
+    else {
+      States states(times, positions, refFrame);
+      return states;
```
kberryUSGS, line 525: As above.

src/Util.cpp
kberryUSGS: If so, I also suggest updating the test so that it could catch this kind of thing.

include/InterpUtils.h
Kelvinrr: What do you mean?

include/Isd.h
Kelvinrr: Presumably, but I don't think we currently don't have a key for it. Now that I think about it am not sure we really need to pack the interpolation method? Probably leave as is until it is part of the JSON ISD

src/Util.cpp
Kelvinrr: I don't see declaring the states separately and manually populating it as much cleaner.

ale/formatters/formatter.py
Kelvinrr: I think keeping it generic "makes and ISD" is good enough since ISIS vs CSM compatibility isn't gonna be a thing in the long term. The separate formatted only exists as a stop-gap.

include/States.h
Kelvinrr: Really not sure how these mistakes happened, I can't blame a cat I don't have 🤔

include/InterpUtils.h
acpaquette: Similarly to how the PositionInterpolation is being defined as LINEAR = 0, SPLINE = 1. Can we do that with the RotationInterpolation modes?

src/Util.cpp
acpaquette: No, literally just the return statement

src/Util.cpp
Kelvinrr: definitely a bug, fixed and updated tests.

include/States.h
kberryUSGS: :cat:

Kelvinrr:
> I think I found one actual bug, otherwise, just comments about comments and docs.
> 
> The update to make all the isd kewords snake_case is a good improvement.
> 
> This is definitely beyond the scope of this PR, but I noticed that some of these classes mostly use snake_case and some mostly use camelCase. This should probably be updated to use a consistent naming format at some point.

A lot of this was me trying to make things consistent, it made it hard to figure out how the additions should look since different files were using different conventions. Eventually, I just stopped once it was outside the scope of my changes.

Kelvinrr:
I'm gonna go ahead and merge this so we can start on integration.

PR #347, opened 2020-04-09
Added Eigen and Json. Moved headers in build to match install

jessemapel:


kberryUSGS:
This looks good to me. Just curious, do you know if it's preferable, in general, for the headers for the ale library to be included as "ale/header.h", rather than -I including the path to the headers and including as just "header.h", or is it to match the style of including headers used for the other libraries in ale? I tried looking this up in c++ coding standards guidelines and found conflicting things.

jessemapel:
@kberryUSGS It's not standardized across packages. I prefer <package>/<header> because it ensures you don't have conflicts. For example, we have a header called Vectors that could conflict with other headers from other packages.

Kelvinrr:
Our names are generic enough that the chances of conflicts are pretty high. Another alternative is to add a prefix to headers (e.h. AleVectors, AleStates, etc.). No preference either way.

kberryUSGS:
I'm happy with this change with including the directory in the include. Thanks @Kelvinrr and @jessemapel for pointing out that the header names are so generic that including including the ale directory helps to prevent conflicts.

PR #349, opened 2020-04-10
Made States use NaNs if it doesn't have velocity

jessemapel:
Previously, States stored 0 velocity if it didn't have velocities. This resulted in some weird behavior were positions without velocities could be interpolated or rotated incorrectly. I changed it to use NaNs so that there is a clear distinction between 0 velocity and unknown velocity.


include/States.h
```diff
@@ -23,8 +25,17 @@ namespace ale {
...
+    State(Vec3d position) : position(position) {
+      velocity = {std::numeric_limits<double>::quiet_NaN(),
+                  std::numeric_limits<double>::quiet_NaN(),
+                  std::numeric_limits<double>::quiet_NaN()};
+    };
     State(Vec3d position, Vec3d velocity) : position(position), velocity(velocity) {};
     State() {};
+
+    bool hasVelocity() const {
+      return !(std::isnan(velocity.x) && std::isnan(velocity.y) && std::isnan(velocity.z));
```
kberryUSGS, line 37: This makes sense literally and logically, but since NaNs kind of infect any calculation they're involved in, is it  useful to have a velocity with 1-2 components, but not 3, that are NaN? So then, ALE will just do the calculations requested, any calculation with a NaN ends up NaN, and that's just the output from ALE This seems reasonable to me, but I wanted to bring it up anyway.

jessemapel: There's no reason why you'd have 1-2 NaNs. I could change these to ors instead.

kberryUSGS: Yeah, having 1-2 NaNs seemed like a weird (impossible?) case. I'd be pro-changing them to ors.

PR #350, opened 2020-04-10
Fixed constant rotations in Orientations

jessemapel:
This fixed a couple things in Orientations:

1. Orientations don't have a single references frame, they are the transformations between reference frames.
1. The constant rotation wasn't actually being used anywhere. Now, when you use rotateAt, it will properly include the constant rotation.

### DIFF
--- a/src/Orientations.cpp
+++ b/src/Orientations.cpp
@@ -8,12 +8,11 @@ namespace ale {
     const std::vector<Rotation> &rotations,
     const std::vector<double> &times,
     const std::vector<Vec3d> &avs,
-    const int refFrame,
     const Rotation &const_rot,
     const std::vector<int> const_frames,
     const std::vector<int> time_dependent_frames
   ) :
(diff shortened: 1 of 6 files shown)


src/Orientations.cpp
```diff
@@ -109,8 +107,10 @@ namespace ale {
     std::vector<Rotation> mergedRotations;
     std::vector<Vec3d> mergedAvs;
     for (double time: mergedTimes) {
+      // interpolate includes the constant rotation, so invert it to undo that
+      Rotation inverseConst = m_constRotation.inverse();
       Rotation rhsRot = rhs.interpolate(time);
-      mergedRotations.push_back(interpolate(time)*rhsRot);
+      mergedRotations.push_back(inverseConst*interpolate(time)*rhsRot);
```
acpaquette, line 113: Does this also need to be done in the rotateStateAt and rotateVectorAt functions?

jessemapel: No, those methods use Orientations::interpolate which already includes the constant rotation.

PR #352, opened 2020-04-15
Adds MiniRF driver to ale, Radar sensor type, and to_usgscsm Radar formatter

kberryUSGS:
Adds a MiniRF ale driver: `LroMiniRfIsisLabelNaifSpice` driver, based on using an ingested ISIS cube, but getting spice from naif

Adds a Radar sensor model type

Adds the ability to write out a Radar sensor model ISD.


ale/base/type_sensor.py
```diff
@@ -112,3 +113,72 @@ def ephemeris_stop_time(self):
...
+        -------
+        : str
+          USGS SAR (synthetic aperture radar) model
+        """
+        return "USGS_ASTRO_SAR_MODEL"
+
+    @property
+    def ephemeris_time(self):
+        newstart = spice.str2et(str(self.utc_start_time))
+        newend = spice.str2et(str(self.utc_stop_time))
```
jessemapel, line 134: SPICE calls shouldn't be in the sensor type mix-in. These should only be in the SPICE mix-in or any driver the inherits from it.

ale/base/type_sensor.py
```diff
@@ -112,3 +113,72 @@ def ephemeris_stop_time(self):
...
+    def wavelength(self):
+        """
+        Returns the wavelength used for image acquistion. 
+
+        Returns
+        -------
+        : double
+          Wavelength used to create an image in meters
+        """
+        return self.wavelength
```
jessemapel, line 147: This is cyclic

ale/base/type_sensor.py
```diff
@@ -112,3 +113,72 @@ def ephemeris_stop_time(self):
...
+    def line_exposure_duration(self):
+        """
+        Returns the exposure duration for each line.
+
+        Returns
+        -------
+        : double
+          Exposure duration for a line
+        """
+        return self.line_exposure_duration
```
jessemapel, line 159: This is also cyclic

ale/base/type_sensor.py
```diff
@@ -112,3 +113,72 @@ def ephemeris_stop_time(self):
...
+    def range_conversion_coefficients(self):
+        """
+        Returns the range conversion coefficients
+
+        Returns
+        -------
+        : list
+          Coefficients needed for range conversion
+        """
+        return self.range_conversion_coefficients
```
jessemapel, line 183: All of these methods are cyclic and will end up just recursively calling themselves until they hit the max depth.

ale/drivers/lro_drivers.py
```diff
@@ -452,7 +452,7 @@ def constant_time_offset(self):
     def additional_preroll(self):
         """
         Returns the addition preroll defined in an IAK.
-
+LroMiniRfIsisLabelNaifSpiceDriver
```
jessemapel, line 455: What is this?

ale/drivers/lro_drivers.py
```diff
@@ -505,3 +505,59 @@ def spacecraft_direction(self):
         rotation = frame_chain.compute_rotation(1, lro_bus_id)
         rotated_velocity = spice.mxv(rotation._rots.as_dcm()[0], velocity)
         return rotated_velocity[0]
+
+
+class LroMiniRfIsisLabelNaifSpiceDriver(Radar, NaifSpice, IsisLabel, Driver):
```
jessemapel, line 510: This should inherit from IsisSpice because we're going to allow ISIS to spiceinit right now.

ale/drivers/lro_drivers.py
```diff
@@ -505,3 +505,59 @@ def spacecraft_direction(self):
...
+    def line_exposure_duration(self):
+        """
+        Line exposure duration in seconds. The sum of the burst and the delay for the return. 
+
+        Returns
+        -------
+        : double
+          scaled pixel width
+        """
+        return self.label['IsisCube']['Instrument']['LineExposureDuration'];
```
jessemapel, line 550: This is already implemented in the IsisLabel class.

ale/drivers/lro_drivers.py
```diff
@@ -505,3 +505,59 @@ def spacecraft_direction(self):
...
+        """
+        Range conversion coefficients
+
+        Returns
+        -------
+        : List
+          range conversion coefficients
+        """
+        return self.label['IsisCube']['Instrument']['RangeCoefficientSet']; 
+
```
jessemapel, line 563: What about the Reference Radii?

jessemapel:
IsisLabelIsisSpice is the current sub-classing to use. Also, use the same routine as the Base class and raise not implemented errors.

ale/drivers/lro_drivers.py
kberryUSGS: Accidental middle-click  paste.

ale/drivers/lro_drivers.py
kberryUSGS: The reference radii actually do come from the PCK so we don't need to grab them separately in here, the default works.

tests/pytests/test_lro_drivers.py
```diff
@@ -10,14 +10,10154 @@
...
 from ale.drivers.lro_drivers import LroLrocIsisLabelNaifSpiceDriver
+from ale.drivers.lro_drivers import LroMiniRfIsisLabelNaifSpiceDriver
 from ale.transformation import TimeDependentRotation
 
 from conftest import get_image_label, get_isd, get_image_kernels, convert_kernels, compare_dicts
 
 image_dict = {
-    'M103595705LE': get_isd("lrolroc")
+    'M103595705LE': get_isd("lrolroc"),
+    '03821_16N196_S1': {
```
jessemapel, line 20: Can you split this into a separate file and use get_isd so that the test file is cleaner.

kberryUSGS: Sure!

PR #354, opened 2020-04-24
Update format of range conversion coefficients isd 

kberryUSGS:
Updated Radar sensor type, driver, and usgscsm formatter output to split the range_conversion_coefficients into a separate range_conversion_times and range_conversion_coefficients. (Previously, the format was: [time, coeff1, coeff2, coeff3, coeff4])

Also, definitely meant to make a branch, but accidentally committed to my own master, so just decided to go with this.


kberryUSGS:
[Update ALE isd generation and all code in usgscsm that handles the range coefficients to split the values into a "times" list and a "coefficients" list](https://trello.com/c/iE12A3VD/349-update-ale-isd-generation-and-all-code-in-usgscsm-that-handles-the-range-coefficients-to-split-the-values-into-a-times-list-and)

PR #358, opened 2020-06-11
Segfault fix

Kelvinrr:
This should fix the issue of segfaults happening when C++ load/loads is called multiple times in a row.


src/Load.cpp
```diff
@@ -84,44 +84,53 @@ namespace ale {
...
+     Py_INCREF(pStringFileName); // take ownership of reference
 
      PyObject *pStringProps = PyUnicode_FromString(props.c_str());
      PyTuple_SetItem(pArgs, 1, pStringProps);
-
+     Py_INCREF(pStringProps); // take ownership of reference
+     
      PyObject *pStringFormatter = PyUnicode_FromString(formatter.c_str());
      PyTuple_SetItem(pArgs, 2, pStringFormatter);
+     Py_INCREF(pStringFormatter); // take ownership of reference
```
Kelvinrr, line 106: Once the strings are added to the tuple, the tuple takes ownership of the reference, so we need to increment the reference to explicitly maintain ownership.

src/Load.cpp
```diff
@@ -84,44 +84,53 @@ namespace ale {
...
      char *temp_str = PyBytes_AS_STRING(temp_bytes); // Borrowed pointer
      cResult = temp_str; // copy into std::string
 
      Py_DECREF(pResultStr);
+
      Py_DECREF(pStringFileName);
      Py_DECREF(pStringProps);
      Py_DECREF(pStringFormatter);
+     
+     Py_DECREF(pArgs);
```
Kelvinrr, line 133: Decrementing the tuple will finally reduce the reference count to 0, as we own a copy and so does the tuple.

PR #359, opened 2020-06-15
Updated Python doc config

jessemapel:
The docs haven't been updated in a while and some files have changed. This gets the Python docs building. The C++ docs still aren't working correctly.


kberryUSGS:
Any idea what's up with the `119: Intel MKL FATAL ERROR: Cannot load libmkl_intel_thread.dylib` error on the Travis Mac build?

jessemapel:
Looks like this https://github.com/JuliaPy/PyPlot.jl/issues/315#issuecomment-334179119 because it's working when we run pytest, but failing when we run through c++.

PR #360, opened 2020-06-15
Added Orientations operations

jessemapel:
This adds inversion and multiplication to orientation objects. Note that the inversion is not perfect as the way we store them isn't full invertible. The frames do not fully invert and the constant rotation gets folded into the time dependent rotation and it becomes fully time dependent.

### DIFF
--- a/include/ale/Orientations.h
+++ b/include/ale/Orientations.h
@@ -88,7 +88,15 @@ namespace ale {
     Orientations &operator*=(const Rotation &rhs);
 
     /**
-     * Invert the orientations
+     * Invert the orientations.
+     * Note that inverting a set of orientations twice does not result in
+     * the original orientations. the constant rotation is applied after the
+     * time dependent rotation. This means in the inverse, the constant
+     * rotation is applied first and then the time dependent rotation.
+     * So, the inverted set of orientations is entirely time dependent.
+     * Then, the original constant rotations cannot be recovered when inverting
+     * again. The set of orientations will still operate the same way, but its
+     * internal state will not be the same.
      */
      Orientations inverse() const;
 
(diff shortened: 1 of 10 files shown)


src/Orientations.cpp
```diff
@@ -144,4 +150,60 @@ namespace ale {
...
+    for (size_t i = 0; i < m_rotations.size(); i++) {
+      newRotations.push_back(m_rotations[i].inverse() * constInverseRotation);
+    }
+
+    std::vector<Vec3d> rotatedAvs;
+    for (size_t i = 0; i < m_avs.size(); i++) {
+      Vec3d rotatedAv = (m_constRotation * m_rotations[i])(m_avs[i]);
+      rotatedAv.x *= -1.0;
+      rotatedAv.y *= -1.0;
+      rotatedAv.z *= -1.0;
```
jessemapel, line 170: It may be worth it to just add a * operator to the vector struct.

include/ale/Orientations.h
```diff
@@ -71,15 +71,27 @@ namespace ale {
...
      */
     Orientations &operator*=(const Orientations &rhs);
+
     /**
      * Multiply this orientation by a constant rotation
      */
     Orientations &operator*=(const Rotation &rhs);
 
+    /**
+     * Invert the orientations
```
jessemapel, line 91: Note for myself that I need to note all of the stuff that doesn't perfectly invert in here.

src/Orientations.cpp
```diff
@@ -144,4 +150,60 @@ namespace ale {
...
+
+
+  Orientations operator*(Orientations lhs, const Rotation &rhs) {
+    return lhs *= rhs;
+  }
+
+
+  Orientations operator*(const Rotation &lhs, Orientations rhs) {
+    return rhs.addConstantRotation(lhs);
+  }
```
acpaquette, line 203: This seems odd compared to the other two multiplication implementations.

jessemapel: How so? There is a little strangeness about what's a reference and what's not. That's just trying to be specific about what needs to be copied and what doesn't.

PR #362, opened 2020-06-19
Adds a "convenience constructor" to States

kberryUSGS:
Adds a constructor to States which conveniently takes the format that `p_cache` is in on ISIS. (a vector of a vector of doubles.)


src/States.cpp
```diff
@@ -99,6 +111,11 @@ namespace ale {
 
 
   State States::getState(double time, PositionInterpolation interp) const {
+// consider adding "just return position if only one" in here. 
+//    std::cout << "Num of times is : " << m_ephemTimes.size() << std::endl; 
+//    if ( (m_ephemTimes.size() == 1) && (m_states.size() == 1) ) {
+//     }
```
jessemapel, line 117: Remove this

PR #363, opened 2020-06-19
Orientations now only need 1 rotation

jessemapel:
This makes it possible to work with only a single point in time and make small interpolations around that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,10 @@ install:
   - conda env create -n ale python=3.7.3
   - conda env update -f environment.yml -n ale
   - source activate ale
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -43,6 +43,10 @@ install:
   - conda env create -n ale python=3.7.3
   - conda env update -f environment.yml -n ale
   - source activate ale
+  - |
+    if [ "$TRAVIS_OS_NAME" == "osx" ]; then
+      install_name_tool -change @rpath/libiomp5.dylib @loader_path/libiomp5.dylib ${CONDA_PREFIX}/lib/libmkl_intel_thread.dylib;
```
jessemapel, line 48: This is a known issue with some of the numpy versions served by conda-forge. In actual use this doesn't pop up, just Travis.

PR #365, opened 2020-06-20
Docs!

acpaquette:
Large documentation update to get things closer to what ALE currently is. Once this is merged the documentation can be updated on the gh-pages branch.


src/Util.cpp
```diff
@@ -148,8 +148,7 @@ std::vector<std::vector<double>> getLineScanRate(json isd) {
       lines.push_back(scanRate.get<std::vector<double>>());
     }
   } catch (...) {
-    throw std::runtime_error("Could not parse the integration start lines in "
-                             "the integration time table.");
+    throw std::runtime_error("Could not parse the line scan rate from the isd.");
```
jessemapel, line 151: The test needs to be updated to expect the new error message

PR #366, opened 2020-06-22
Adds windows build CI

jessemapel:
This adds windows build CI for both hand building and conda-build. The actual publishing of conda for windows will happen on the ale-feedstock. Once this is done, I will move the conda-build changes over there.


recipe/meta.yaml
```diff
@@ -13,33 +13,32 @@ source:
...
+    - eigen
+  host:
+    - pip
+    - python
+    - nlohmann_json
   run:
-  - eigen
-  - networkx
-  - numpy
-  - openblas
```
kberryUSGS, line 33: Why is `openblas` still in the `environment.yml` but eliminated here? Is it only needed for development work?

jessemapel: I don't know why it's in the environment.yml. It's not in our [conda-forge feedstock](https://github.com/conda-forge/ale-feedstock/blob/master/recipe/meta.yaml) so I think I'll take it out of the environment.yml here.

PR #373, opened 2020-06-26
Pvl 1.0.0 Update

### LINKED ISSUE #356: PVL Decoder in Dawn Driver needs to be updated for PVL 1.0
The decoder structure in PVL is changing with their upcoming 1.0 release. The Dawn drivers use a custom decoder to ignore some non-standard characters and needs to be updated to work with the new PVL decoder API. We may not even need to subclass because PVL now supports a bunch of different grammars.

acpaquette:
This PR is ready for review as of the pvl 1.0.0 release! This PR is currently up to date with pvl 1.0.0

Closes #356


environment.yml
```diff
@@ -11,6 +11,7 @@ dependencies:
   - nlohmann_json
   - numpy
   - pvl
+  - pytz
```
jessemapel, line 14: Why do we need a new dependency here? Does DateTime not handle local time?

jessemapel (approved):
From this morning's discussion, pytz is not a large dependency so this is good to go.

PR #374, opened 2020-07-01
Add code to States to not interpolate if an exact solution exists

kberryUSGS:
... this fixes a divide-by-zero error in the actual interpolation code.


src/States.cpp
```diff
@@ -111,6 +111,14 @@ namespace ale {
 
 
   State States::getState(double time, PositionInterpolation interp) const {
+
+    // If time is in times, don't need to interpolate!
+    for (int i=0; i<m_ephemTimes.size(); i++){
```
jessemapel, line 116: A linear search is going to be very slow here. This should use std::lower_bound and then compare the output instead.

src/States.cpp
```diff
@@ -111,6 +111,16 @@ namespace ale {
 
 
   State States::getState(double time, PositionInterpolation interp) const {
+
+
+    // If time is in times, don't need to interpolate!
+    auto candidate_time = std::lower_bound(m_ephemTimes.begin(), m_ephemTimes.end(), time);
+
+    if (*candidate_time == time) {
```
jessemapel, line 119: lower_bound can return `m_ephemTimes.end()` which you don't want try and access. This if should be `candidate_time != m_ephemTimes.end() && *candidate_time == time` which will short-circuit in that case.

kberryUSGS: Good point, thanks

kberryUSGS:
Good point, I'll also add a test, thanks!

PR #382, opened 2020-07-13
ALE Spice Directory Fix

acpaquette:
Set mission directories to an empty list if ALESPICEROOT is not set


jessemapel:
There's a test that checks for the environment variable warning not being set in test_env.py that'll need to be updated.

acpaquette:
@jessemapel This should be good to go now!

jessemapel:
I'll get a PR in tomorrow to tick the version numbers and such for a release.

PR #385, opened 2020-07-22
Updated Mini RF to use the unified formatter

jessemapel:
I checked the new test output and it looks correct, but I want to test it out ~in a sensor model~ more before this gets merged.


jessemapel:
I checked the ephemerides in this against ISIS and there are lining up. So, I'm going to say this is ready to go.

jessemapel:
Once the USGSCSM work is done I've got a knoten comparison notebook I'm working on too that I'll put up.

jessemapel:
@acpaquette The json ISD didn't get removed, it just got greatly reduced because I made the image smaller. It went from 10,000 lines to a few hundred.

acpaquette:
Oh, my bad. I didn't try to actually load the diff. Looks all good then!

PR #392, opened 2020-11-17
Fixed Viking and Cassini ISIS label ISIS SPICE drivers

jessemapel:
These drivers were succeeding for images that were not from their sensors. These changes both make sure they properly check the instrument_id and fix some timing issues.


ale/drivers/co_drivers.py
```diff
@@ -329,6 +329,29 @@ def frame_chain(self):
...
+
+    id_lookup = {
+        "ISSNA" : "CASSINI_ISS_NAC",
+        "ISSWA" : "CASSINI_ISS_WAC"
+    }
+
+    name_lookup = {
+        "ISSNA" : "Imaging Science Subsystem Narrow Angle Camera",
+        "ISSWA" : "Imaging Science Subsystem Wide Angle Camera"
+    }
```
acpaquette, line 341: Can we make these look ups part of the module rather than member variables in the class? Similar to the viking drivers. This would also need to happen for the CassiniIssPds3LabelNaifSpice

acpaquette (commented):
Overall looks good! Small thing to change but thank you for fixing my hastily written drivers. The only other thing that really needs to happen is tests but these fixes would probably need to happen anyway

tests/pytests/data/isds/cassiniiss_isis_isd.json
```diff
@@ -0,0 +1,201 @@
+{
```
jessemapel, line 1: I had to add a separate ISD for the ISIS label because the Cassini kernels set has been updated since the pds3 data was generated so values have shifted a little bit.

tests/pytests/conftest.py
```diff
@@ -95,6 +98,21 @@ def get_image_label(image, label_type='pds3'):
 
     return label_file[0]
 
+def get_table_data(image, table_name):
```
jessemapel, line 101: This function is designed to be used with the output of isis's blobdump app which dumps a  BLOB/Table to a separate file with the Table's PVL label and then the binary data.

acpaquette (approved):

PR #402, opened 2021-02-11
Fixed no av seg fault when multiplying orientations

### LINKED ISSUE #401: Multiplying two orientations together when one doesn't have angular velocities causes a seg fault
This is from https://github.com/USGS-Astrogeology/usgscsm/issues/324

When multiplying two orientations together it calls interpolateAV at the combine cache times. This seg faults when no angular velocities are available. So, we should do two things:

1. If you call interpolateAV with no angular velocities it should throw an exception.
2. If you multiply two orientations and either of them lacks angular velocity, the output should lack angular velocity.

jessemapel:
I also added an exception when trying to interpolate the angular velocity for orientations without any angular velocities.

Fixes #401

### DIFF
--- a/src/Orientations.cpp
+++ b/src/Orientations.cpp
@@ -84,7 +84,10 @@ namespace ale {
 
   Vec3d Orientations::interpolateAV(double time) const {
     Vec3d interpAv;
-    if (m_times.size() > 1) {
+    if (m_avs.empty()) {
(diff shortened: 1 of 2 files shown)


scsides (approved):
makes sense

kberryUSGS (approved):
:+1: Looks good!

PR #404, opened 2021-03-10
Readthedocs

ladoramkershner:
Getting readthedocs to work on my local branch. This involved removing an old version of the conf.py file, so I want to make sure that actually sticks when we merge.


docs/index.rst
```diff
@@ -0,0 +1,11 @@
+#############################
+Ephemeris Abstraction Library
```
kberryUSGS, line 2: Maybe update this to: `Abstraction Layer for Ephemerides (ALE)` ? (This project was briefly called "EAL" in the past.)

docs/make.bat
```diff
@@ -0,0 +1,35 @@
+@ECHO OFF
```
kberryUSGS, line 1: It probably makes sense to also commit the `Makefile` generated by sphinx to the repo, since we're more likely to build on Mac/Linux than Windows.

docs/index.rst
ladoramkershner: Good catch! I just copied up the old index.rst that was in docs/source/ folder (as opposed to just docs/). So I fixed it in there as well!

docs/make.bat
kberryUSGS: Never mind, it's already in there.

scsides (approved):

jessemapel:
If this is getting pushed up to readthedocs, you should pull down the github pages documentation on the gh_pages branch:

https://github.com/USGS-Astrogeology/ale/tree/gh-pages

Also, point the badge on the Readme to readthedocs.

Finally, 👍 for moving to something better than github pages.

PR #405, opened 2021-03-11
Doc badge

ladoramkershner:
I got the doc badge working. There was also a weird thing on the RTDs page where you could select the project name to go to the EXACT same page, just with out the link you just clicked on (TLDR; redundant link). I think it was something I added with the docs/index.rst pointing to the docs/source/index.rst. I got rid of it and moved the source/ files up.


jessemapel (approved):

jessemapel:
Looks good. I looked at the actual docs page and it generated the tree properly, but it didn't pull in any of the actual docs for classes or functions. Is there something else that needs to be setup to get that working?

kberryUSGS (approved):
Looks good + same comment as Jesse. I also noticed that the function docs aren't on there.

scsides (approved):

PR #407, opened 2021-03-17
Added breathe and sys path to try and get docs building

jessemapel:
Used official docs and an example:

https://breathe.readthedocs.io/en/latest/quickstart.html

https://github.com/Cruel/readthedocs-breathe

Also, added the system path so that it can find the ale module

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -10,9 +10,10 @@
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 #
-# import os
-# import sys
-# sys.path.insert(0, os.path.abspath('.'))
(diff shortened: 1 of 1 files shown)


ladoramkershner:
I feel like with these kinds of PRs you have to merge them before you can see if they work. That is unless you have your personal repo set up with a RTD page. Are we okay to  have quick merges on these type of PRs?

ladoramkershner:
I did notice that in the breathe quickstart it does mention 
"The documentation path should contain a folder `source` containing the conf.py file"
This is something I undid, because the source folder seemed redundant

jessemapel:
I don't know if we need the source folder but I'm good with just merging this and iterating on it.

PR #414, opened 2021-04-05
Fixed pre-parse try except

jessemapel:
Fixed the pre-parse try except so that it properly fails through instead of only failing on value errors.


kberryUSGS (approved):
:+1: Looks good!

PR #418, opened 2021-06-10
Docs Fix

acpaquette:
Fixed project definition in docs rst files preventing the documentation from building correctly


ladoramkershner (approved):

jessemapel:
I both love and hate that the solution was this simple

ladoramkershner:
If it helps/hurts we needed one more super simple change to get the python side going as well!

PR #421, opened 2021-07-08
Have CSM write camera model files on multiple lines, for readability

oleg-alexandrov:
This is a proposed fix for https://github.com/USGS-Astrogeology/usgscsm/issues/337.

I've worked extensively with .json files produced by ALE recently and having the whole file in a single line makes it very hard to understand what is in the model and do comparisons, etc.

This fix makes the .json files be written on multiple lines and nicely indented. (The whole point of json is to provide a pretty and simple format, unlike XML, so this way, with pretty indentation, json achieve its full potential.)


oleg-alexandrov:
And this was very carefully tested.

jessemapel (approved):

PR #425, opened 2021-08-10
added missing HRSC properties

Kelvinrr:
HRSC driver for ISISLabel + NaifSpice was missing some required overrides, so this should patch those in. 

fixes https://github.com/USGS-Astrogeology/ale/issues/424


jessemapel (approved):

Kelvinrr:
Was trying to compare to knoten notebooks but it seems like the HRSC notebook never ran as it is using a CTX image? Explains why this driver is in non-working order, probably never was finished. As I don't have much to compare against, and this makes an ISD possible, probably fine to merge.

PR #428, opened 2021-09-10
Save center_ephemeris_time to the SAR sensor model

oleg-alexandrov:
The center_ephemeris_time should be saved in the SAR sensor model, or else the SAR model will fail to be loaded, when the following code is encountered: 

https://github.com/USGS-Astrogeology/usgscsm/blob/da685687d4842d4f60d0a427348288517237664b/src/UsgsAstroSarSensorModel.cpp#L62


oleg-alexandrov:
I think there's a bit of urgency to this. Latest usgscsm code cannot load a model file without this fix. It says:

CSM warning: Could not create model [USGS_ASTRO_SAR_SENSOR_MODEL] with error [Could not construct model [USGS_ASTRO_SAR_SENSOR_MODEL] with error [ISD is invalid for creating the sensor model with error [Could not parse the center image time.]]]

PR #432, opened 2022-01-07
Adds Cassini Vims driver for use within ISIS

acpaquette:
Added cassini vims vis/ir drivers. Something is still off in the driver, since we are about +-.1 pixels in the x and y directions.

As an aside, this driver will likely never see a CSM driver written for it.

Still needs:
- [ ] To be fixed
- [ ] Tests written


jessemapel:
I will be modifying the repo history as part of #440, if this PR is ready to go please flag it as such and we can get it merged before I modify the history. Otherwise, it will need to be rebased onto the new master branch.

PR #442, opened 2022-03-04
Updates for spiceypy 5

jessemapel:
As of spiceypy 5.0, nrows and ncols params are gone. CSPICE 67 also appears to have uncovered a bug in the Kaguya SCLK so I've corrected it in our test data. I e-mailed JAXA about the issue and resolution. If they push a new SCLK file, then I'll follow up with a PR to replace with that.


tests/pytests/data/MVA_2B2_01_02329N002E0302/SEL_M_V01.TSC
```diff
@@ -11,7 +11,7 @@ SCLK_KERNEL_ID                = ( @2009-06-10T12:57:27.4670 )
 SCLK_DATA_TYPE_131       = ( 1 )
 SCLK01_N_FIELDS_131      = ( 1 )
 SCLK01_MODULI_131        = ( 4294967296 )
-SCLK01_OFFSETS_131       = ( 0 0 )
+SCLK01_OFFSETS_131       = ( 0 )
```
jessemapel, line 14: According to the [SCLK Required Reading](https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/FORTRAN/req/sclk.html#Format%20constant%20assignments), this is now correct. There should be 1 offset per field and `SCLK01_N_FIELDS_131      = ( 1 )` only indicates 1 field.

environment.yml
```diff
@@ -15,7 +15,7 @@ dependencies:
   - python
   - python-dateutil
   - scipy>=1.4.0
-  - spiceypy>=2.3.0
+  - spiceypy>=4.0.1
```
jessemapel, line 18: ncols, nrows was deprecated in 4.0.1 https://spiceypy.readthedocs.io/en/main/changelog.html#id5

acpaquette (approved):

PR #455, opened 2022-05-13
Print clear error messages when cannot look up things

oleg-alexandrov:
When ALE parses some metadata, it can fail with the cryptic message "TypeError: 'NoneType' object is not iterable". 

This pull request checks for the output of parsing for something, and if that fails, spells out what exactly went wrong.

### DIFF
--- a/ale/drivers/lro_drivers.py
+++ b/ale/drivers/lro_drivers.py
@@ -573,7 +573,10 @@ class LroLrocNacIsisLabelIsisSpiceDriver(LineScanner, IsisSpice, IsisLabel, Driv
         : list
           Radial distortion coefficients. There is only one coefficient for LROC NAC l/r
         """
-        return [self.naif_keywords.get('INS{}_OD_K'.format(self.ikid), None)]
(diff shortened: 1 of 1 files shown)


ladoramkershner:
This is a really great addition! Do you think printing out the keys that failed the lookup would be a good addition to the print outs? For example, the logic for pixel2focal_x() (starting line 961) could become something like:

```
key = 'INS{}_TRANSX'.format(self.fikid)
ans = self.naif_keywords.get(key, None)
if ans is None:
    raise Exception(f'Could not parse detector to focal plane x using key {key}.')
return ans
```

I think this PR is a definite improvement already, but I am curious to know your thoughts on this idea!

oleg-alexandrov:
I guess this is for this package's maintainers to say. I guess a little more information in the error message can be of help.

jessemapel (commented):
I agree with Lauren, the error should report the key it fails to find.

oleg-alexandrov:
I pushed that in a second commit. Key be it. 

A quick test yields: "Failed: Could not parse the focal length using key: INS-85636_FOCAL_LENGTH"

ladoramkershner (approved):

Issue #440, opened 2022-03-04
Administrative Security Review


jessemapel:
As part of the USGS Release process the repo needs and administrative code review as per the [USGS Software Management website](https://www.usgs.gov/software-management/types-software-review).

jessemapel:
I've combed through the repository and cannot find any sensitive information.

scsides:
Admin review for ALE:

- There are no hostnames that are not public servers
- There are no IP addresses
- There are no username/passwords
- There are many absolute paths. Without having the hostname and access to it, these paths are of little value, but they are "full paths." The examples below have been shorted to show enough to find the lines with the full paths without further adding to the problem with this issue's text. This is not an exhaustive list.

  - ale/util.py:def   cpkgs/isis3/data/
  - notebooks/write_MexHrscPds3LabelNaifSpiceDriver.ipynb:    h5270_0000_ir2.cub
  - notebooks/write_MessengerMDISPds3LabelNaifSpiceDriver.ipynb:  spice/mess-e_v_h-spice-6-v1.0/messsp_1000/extras/mk
  - notebooks/write_MexHrscPds3LabelNaifSpiceDriver.ipynb:    h5270_0000_ir2.img
  - notebooks/write_CassiniIssPds3LabelNaifSpiceDriver.ipynb:  cpkgs/isis3/data/cassini/kernels

A search that shows the potential problems (413 including false possitives) is:
grep -r -e " /" -e "\"/" * | grep  -vi  -e "}\/" -e " \/ " -e "\/\/" -e http -e "\/\*" -e "\*\/" -e "\.tf:" -e "\.ti:"
Some files identified with this grep are SPICE and .cub files.

jessemapel:
I think the notebook directory can go. All of them are out of date and we're pushing newer versions of that documentation over to USGSCSM.

I will put in a PR to remove the util.py path.

jessemapel:
@scsides Is this good now that #461 is done?

jlaura:
@jessemapel The paths were removed, but the fix is not complete. For example: https://github.com/USGS-Astrogeology/ale/blob/da2e7b2faadcefe0d33153e3c89b97a436699a5e/ale/util.py

When one goes back through the history on the repo, the absolute paths are maintained. The remediation (which is good practice if we ever have to do this for something more serious than these absolute paths) is to do a history rewrite.

jessemapel:
[...]

PR #462, opened 2022-05-24
Initial EIS WAC drivers

jessemapel:
Initial WAC drivers based on our test labels. These are very rough as the initial SPICE is very brief and the labels are missing several key values.

I'm marking this as draft for now and will follow up with tests. If we get NAC labels then I'll add those too.


jessemapel:
This PR will not be ready for the work for #440 is complete, I will rebase it onto the new master branch.

scsides (approved):

PR #463, opened 2022-05-25
Adds CassiniIssNac/Wac IsisLabel NaifData Driver

acpaquette:
Adds the Cassini ISS NAC/WAC IsisLabel NaifSpice driver for use in ISIS. Also tests the driver with the expected tests on the added properties and a loads test.


jessemapel:
I will be modifying the repo history as part of https://github.com/USGS-Astrogeology/ale/issues/440, if you can fix the failing tests and we can get it merged before I modify the history. Otherwise, it will need to be rebased onto the new master branch.

I can preform the re-base if you allow maintainers permission to modify your PR.

ale/drivers/co_drivers.py
```diff
@@ -110,6 +110,186 @@
...
+        so we aquired updated NAC filter data from ISIS's IAK kernel.
+
+        """
+        # default focal defined by IAK kernel
+        try:
+            default_focal_len = super(CassiniIssPds3LabelNaifSpiceDriver, self).focal_length
+        except:
+            default_focal_len = float(spice.gdpool('INS{}_DEFAULT_FOCAL_LENGTH'.format(self.ikid), 0, 2)[0])
+
+        filters = tuple(self.label["IsisCube"]["BandBin"]['FilterName'].split("/"))
```
Kelvinrr, line 217: since this can still git the spice kernels, should we start cacheing the values with an underscore prefixed variable? Or maybe use the naif keywords property and have that be cached in the background.

acpaquette: Yes, I think anything that uses a spice call should probably be cached in an underscore

PR #475, opened 2022-06-10
Marci Driver

acpaquette:
Adds the Marci driver for use within ISIS. There are changes that need to go into ISIS for this driver to work but this gets it working in ALE.

A small bug was also fixed in the comparison function to handle arrays of strings.


ale/drivers/mro_drivers.py
```diff
@@ -5,9 +5,195 @@
...
+            self._ephemeris_stop_time = max(self.compute_marci_time(line))
+        return self._ephemeris_stop_time
+
+    @property
+    def detector_center_line(self):
+        return 0
+
+    @property
+    def detector_center_sample(self):
+        return 0
```
Kelvinrr, line 142: not to be a pain, but can we get doc strings for these?

Kelvinrr, line 142: not to be a pain, but can we get doc strings for these?

ale/drivers/mro_drivers.py
```diff
@@ -5,9 +5,195 @@
...
+
+    @property
+    def sensor_model_version(self):
+        """
+        Returns
+        -------
+        : int
+          ISIS sensor model version
+        """
+        return 1
```
Kelvinrr, line 195: Probably be useful to have some boilerplate for docs here, but adding a boilerplate doc string is probably better than nothing?

tests/pytests/conftest.py
```diff
@@ -60,8 +60,13 @@ def compare_dicts(ldict, rdict):
         elif isinstance(item, np.ndarray) or isinstance(item, list):
             if len(item) != len(rdict[key]):
                 differences.append(f'Array sizes of key {key} are not equal {item} : {rdict[key]}.')
-            elif not np.allclose(item, rdict[key]):
-                differences.append(f'Array values of key {key} are not almost equal {item} : {rdict[key]}.')
+            else:
+                if isinstance(item[0], int) or isinstance(item[0], float):
```
Kelvinrr, line 64: you can probably clean this up by doing something like:

```python
import numbers
isinstance(num, numbers.Number)
```

Kelvinrr, line 64: you can probably clean this up by doing something like:

```python
import numbers
isinstance(num, numbers.Number)
```

Kelvinrr (commented):
just some comments.

PR #476, opened 2022-07-20
Osirisrex Driver

amystamile-usgs:


ale/drivers/osirisrex_drivers.py
```diff
@@ -0,0 +1,101 @@
...
+import ale
+from ale.base.data_naif import NaifSpice
+from ale.base.label_isis import IsisLabel
+from ale.base.type_sensor import Framer
+from ale.base.base import Driver
+from ale.base.type_distortion import RadialDistortion
+
+class OsirisRexCameraIsisLabelNaifSpiceDriver(Framer, IsisLabel, NaifSpice, RadialDistortion, Driver):
+    @property
+    def instrument_id(self):
```
acpaquette, line 12: Could we get doc strings on all of these properties?

acpaquette (commented):
:+1: This looks awesome!

ale/drivers/osirisrex_drivers.py
amystamile-usgs: Good catch. I forgot to go back and add these! They are in now.

PR #478, opened 2022-07-26
Exact Instrument Pointing Data Extraction

acpaquette:
Update to mirror how ISIS extracts instrument pointing from kernels. This was due to issues in MVIC TDI when trying to get it close to ISIS.

This has larger impacts on drivers as a whole, largely line scanners that do not record an ephemeris record per line. All tests have been updated but all drivers should be tested in knoten to ensure accuracy when compared with the ISIS drivers.


ale/transformation.py
```diff
@@ -96,38 +96,63 @@ class FrameChain(nx.DiGraph):
...
     @classmethod
-    def from_spice(cls, sensor_frame, target_frame, center_ephemeris_time, ephemeris_times=[], nadir=False):
+    def from_spice(cls, sensor_frame, target_frame, center_ephemeris_time, ephemeris_times=[], nadir=False, exact_ck_times=False):
         frame_chain = cls()
+        sensor_times = []
+        target_times = np.array(ephemeris_times)
 
-        times = np.array(ephemeris_times)
+        if exact_ck_times and not nadir:
+            print("Computing New Times")
```
jessemapel, line 105: This should only print if it's verbose.

ale/transformation.py
```diff
@@ -96,38 +96,63 @@ class FrameChain(nx.DiGraph):
...
 
-        time_dependent_frames = list(zip(sensor_time_dependent_frames[:-1], sensor_time_dependent_frames[1:]))
+        sensor_time_dependent_frames = list(zip(sensor_time_dependent_frames[:-1], sensor_time_dependent_frames[1:]))
         constant_frames = list(zip(sensor_constant_frames[:-1], sensor_constant_frames[1:]))
         target_time_dependent_frames = list(zip(target_time_dependent_frames[:-1], target_time_dependent_frames[1:]))
         target_constant_frames = list(zip(target_constant_frames[:-1], target_constant_frames[1:]))
 
-        time_dependent_frames.extend(target_time_dependent_frames)
+        # time_dependent_frames.extend(target_time_dependent_frames)
+        # print(time_dependent_frames)
```
jessemapel, line 124: extra debug comments.

ale/transformation.py
```diff
@@ -271,3 +296,85 @@ def last_time_dependent_frame_between(self, source, destination):
                 return path[i+1], path[i], edge
 
         return None
+
+    @staticmethod
+    def extract_exact_ck_times(observStart, observEnd, targetFrame):
```
jessemapel, line 301: At least a simple docs string will be very helpful here.

ale/transformation.py
```diff
@@ -271,3 +296,85 @@ def last_time_dependent_frame_between(self, source, destination):
...
+                    et = spice.sct2e(sclkSpCode, sclkdp[instance])
+
+                    while (instance < (ninstances - 1) and et < observEnd):
+                        times.append(et)
+                        instance = instance + 1
+                        et = spice.sct2e(sclkSpCode, sclkdp[instance])
+                    times.append(et)
+
+                    if not observationSpansToNextSegment:
+                        timeLoaded = True
```
jessemapel, line 373: This looks unused now

ale/transformation.py
```diff
@@ -271,3 +296,85 @@ def last_time_dependent_frame_between(self, source, destination):
...
+
+        FILESIZ = 128;
+        TYPESIZ = 32;
+        SOURCESIZ = 128;
+
+        currentTime = observStart
+        timeLoaded = False
+
+        count = spice.ktotal("ck")
+        file, filtyp, source, handle = spice.kdata(0, "ck", FILESIZ, TYPESIZ, SOURCESIZ)
```
jessemapel, line 312: We could do some more complex logic where we open all of the CKs and then directly build the frame chain from that, but emulating ISIS here is sufficient.

ale/util.py
```diff
@@ -495,7 +500,7 @@ def duckpool(naifvar, start=0, length=10, default=None):
     return default
 
 
-def query_kernel_pool(matchstr="*"):
+def query_kernel_pool(matchstr="*", max_length=10):
```
jessemapel, line 503: Why is this being added? I could see some instances where we need more than 10 values.

jessemapel (commented):
Changes look reasonable. I wonder if the test failures are because of different kernel load ordering on the different OS. The new logic means that you will get the times for whichever CK is loaded first, so the case-sensitive difference between mac and ubuntu could be causing different ordering when we glob all of the kernel files to load for testing.

ale/util.py
acpaquette: 10 was the previous default and works in *most* situations but I needed to add if for a particular driver. I just can't remember if it was MVIC TDI or Marci

amystamile-usgs (approved):
👍

Issue #480, opened 2022-08-05
Define the use of Mars using as Sphere


thareUSGS:
As described in this RFC, we plan to transition stereo-processing in SOCET SET/GXP to _"1. Initialize projects in SOCET SET using the IAU-recommended sphere for Mars (radius = 3396190 meters)"_  and a couple other updates as described here:
https://astrodiscuss.usgs.gov/t/appl-rfc-use-iau-sphere-for-hirise-dtm-projects-in-socet-set/428

BTW, one additional benefit not described in the RFC, is that we will not need to support the various ocentric/ographic jumps to properly support SOCET. Our current workflow, requires ocentric -> ographic -> ocentric transformations during our full stereo workflow as SOCET only recognizes ographic latitudes. Using a sphere, removes this burden (and data resampling).

So as we test ALE-generated metadata (ISDs) for usgscsm (in GXP), the current ISDs for all Mars (HiRISE and CTX) are defined using the default Mars ellipse which is listed as '"radii": {"semimajor": 3396.19, "semiminor": 3376.2, "unit": "km"}'. I assume the defined ellipse may also be impacting the ISD calculations.

For ISIS's spiceinit, I believe a simple update to Mars within pck00009.tpc should help to define Mars as a sphere. For ALE, I am not sure how to override these Mars defaults. Perhaps updating the same file with the NAIF kernels? But if the kernels are within a common area read-only area, can I update the path to the pck file? Or do I need to copy all the kernels and update the kernel path to grab the updated pck file?

I hope this question makes sense? Now, if there is not an easy way to point to an updated pck file (outside the main geneally read-only kernel data area), I guess this could be considered as an enhancement request to allow for a pck/radius override.

jessemapel:
The body radii are pulled in from the PCK or the ISIS cube depending on how you are using it. Making sure that your data sets are consistent is outside the scope of ALE. So, I'm going to say we don't make any changes in ALE and instead recommend some actions for users.

1. If you want to update an existing ISD just modify the "radii" part of the JSON. Change the "semiminor" value to be the same as the "semimajor" value.
2. If you are making a new ISD, then either add an extra kernel that redefines Mars as a sphere when you specify the kernel set in loads or in spiceinit before you run loads.

jessemapel:
[...]

PR #486, opened 2022-08-18
Update to isd_generate.py to allow for a NAIF radius override

thareUSGS:
For our Mars pipeline in SOCET GXP, we still are planning to use a sphere (using the Mars' areoid, [see](https://stereopipeline.readthedocs.io/en/latest/tools/dem_geoid.html)).

Currently ALE uses the Mars elliptical defaults from the NAIF kernels (or ISIS cubes). This update allows the user to override the radius values and set a spherical radius for both semimajor and semiminor in the ISD.

This is not only needed for Mars. For triaxial bodies, the IAU recommends a best-fit sphere should also be set for map products. References (DOI) are listed in the updated help for the current IAU values.

This is attempting to address https://github.com/USGS-Astrogeology/ale/issues/480


thareUSGS:
@rbeyer You might want to have a look at this also.

ale/isd_generate.py
```diff
@@ -143,6 +160,15 @@ def file_to_isd(
...
         usgscsm_str = ale.loads(file)
 
+    if radius is not None:
+        usgscsm_json = json.loads(usgscsm_str)
+        usgscsm_json["radii"]["semimajor"] = radius
+        usgscsm_json["radii"]["semiminor"] = radius
+        logger.info(f"Overriding radius to:")
+        logger.info(usgscsm_json["radii"])
+        usgscsm_str = json.dumps(usgscsm_json, indent=2)
+        isd_file.write_text(usgscsm_str)
```
rbeyer, line 170: Remove line 170, otherwise you'll write the isd_file twice.

thareUSGS: good catch

rbeyer (approved):

ale/isd_generate.py
```diff
@@ -84,9 +97,12 @@ def main():
         except KeyError:
             k = [args.kernel, ]
 
+    if args.radius is not None:
+        rad = args.radius
+
```
jlaura, line 102: why not omit this and just pass args.radius in on line 105? Also, just reading the code, doesn't this error if you omit a radius because rad is undefined when you hit line 105?

ale/isd_generate.py
```diff
@@ -51,6 +51,19 @@ def main():
...
+        default=None,
+        help="Optional spherical radius (km) override.  If not specified, the default "
+             "radius values, as defined in the NAIF kernels, will be used.  "
+             "There often cases were the defined NAIF radii are set as a triaxial "
+             "when the IAU WGCCRE report recommends a best-fit sphere for a "
+             "derived map product. For current IAU spherical recommendations see: "
+             "https://doi.org/10.1007/s10569-017-9805-5 or "
+             "http://voparis-vespa-crs.obspm.fr:8080/web/. "
+             "Make sure radius value is in kilometers."
+    )
```
jlaura, line 66: What about nargs="+" (must pass one or more values, e.g., 3396190, 3376250 OR just 3396190) or nargs="2" (must pass two values, e.g.,3396190 3396190) here and making the radius fully overridable? When the args come, as a list, you would need some logic to determine the length and apply to semi major / semi minor axes properly.

I know that the current addition fits the Mars use case, but what if I want to use old (or new) IAU radii. It seems like it would be quite nice to be able to pass these fully dynamically.

ale/isd_generate.py
```diff
@@ -51,6 +51,19 @@ def main():
...
              ".json will be used to generate the output file paths."
     )
+    parser.add_argument(
+        "-r", "--radius",
+        type=float,
+        default=None,
+        help="Optional spherical radius (km) override.  If not specified, the default "
+             "radius values, as defined in the NAIF kernels, will be used.  "
+             "There often cases were the defined NAIF radii are set as a triaxial "
+             "when the IAU WGCCRE report recommends a best-fit sphere for a "
```
jessemapel, line 61: Remove the comments about the default NAIF kernels as it is not correct. Instead mention that this is for when the radii you want are not programmatically available (could be NAIF kernels, could be the cube labels, could be some other data source).

ale/isd_generate.py
jessemapel: If we want to add the ability to pass multiple radii, then we should defer to the 1 radii case as it's the most common. I do not like forcing users to enter the same radii twice when we can just add some extra logic.

jessemapel (commented):
This needs tests.

jessemapel:
Fixes #480

ale/isd_generate.py
thareUSGS: I will omit, but no it won't error as the default is set to None.

rbeyer: Yes this if-statement is redundant and will lead to an error as described.

Just passing `args.radius` to the `radius=` parameter of `file_to_isd()` is the right pattern here.

ale/isd_generate.py
thareUSGS: allow user to pass --semiminor as sphere or --semimajor --semiminor for ellipse.  Fighting way too with argparse to send send "only" two values (as floats) so I went with two parameters. Sort of annoying but works.

rbeyer:
Man, I hesitate to pile more on here, but while `required='--semiminor' in sys.argv` is elegant, if I'm reading it correctly, it is doing the "opposite" thing the `help=` text for `--semiminor` and `--semimajor` are saying, since `--semiminor` has no `required=` parameter.

I have a suggestion based on the use pattern of proj (which maybe I love a little too much).  Provide these three command line arguments:
- -a, --semimajor
- -b, --semiminor
- -r, --radius

Then write some logic after `parser.parse_args()` that does the following:
- If a user gives -a and not -b (or vice versa): error.
- If a user gives -r and also -a or -b: error.

Then assemble the sequence of axes to give to `file_to_isd()` as appropriate depending on what you got.

This allows someone to give identical values to -a and -b, or use -r as a shortcut.

Also, the `help=` parameters use the term *ellipse* but we're talking about an *ellisoid* here, right?

thareUSGS:
I tested `required='--semiminor' in sys.argv` and I placed it exactly as you stated, and it works the opposite. completely confusing but the current code tests correctly (must have semimajor if semiminor sent).

I like the -a, -b, -r options (ala PROJ). And yes ellipsoid - I will update.

Now to figure out how to add some tests.

jessemapel (commented):
The existing tests for this script are here: https://github.com/USGS-Astrogeology/ale/blob/master/tests/pytests/test_isd_generate.py

ale/isd_generate.py
```diff
@@ -51,6 +51,34 @@ def main():
              "and the default strategy of replacing their final suffix with "
              ".json will be used to generate the output file paths."
     )
+    parser.add_argument(
+        "--semimajor", "-a", "-r", "--radius",
+        required="--semiminor" in sys.argv,
```
AustinSanders, line 56: I tested this locally with -b and no -a. It doesn't throw an error if -b is passed without -a, but it silently uses the default radii.

The "required" option does seem to work if we pass the full "--semiminor" flag without "--semimajor" i.e. it throws an error and an explanation that both values are required.

It looks like changing this line to
```         required="--semiminor" in sys.argv or "-b" in sys.argv,```
will get this working correctly.

**Review by AustinSanders — CHANGES_REQUESTED** [review]
A few questions and 2 quick changes, then this should be good to go.

File: ale/isd_generate.py [hunk 6]
```diff
@@ -51,6 +51,34 @@ def main():
...
+        type=float,
+        default=None,
+        help="Optional spherical radius (km) override.  Setting "
```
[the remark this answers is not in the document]
thareUSGS: That is a good point. Perhaps the added "-a" and "-b" and "-r" could be easily confused to PROJ (which requires meters). 

I am not wedded to keeping those alternative flags. I guess another way to approach this is to always require meters (for --semimajor, etc.) and then write the radius into the ISD as km.  I always prefer using meters myself.

ale/isd_generate.py
```diff
@@ -51,6 +51,34 @@ def main():
              "and the default strategy of replacing their final suffix with "
              ".json will be used to generate the output file paths."
     )
+    parser.add_argument(
+        "--semimajor", "-a", "-r", "--radius",
```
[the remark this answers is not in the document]
thareUSGS: The use "-r" or "--radius" is meant to follow the use of "+R" which is only the semimajor as a sphere. https://proj.org/en/9.5/usage/ellipsoids.html#ellipsoid-size-parameters

I am leaning toward requiring meters for all these.

ale/isd_generate.py
thareUSGS: ahh perfect. I fought with that for a longtime and had issues getting that to work. Good find and fix.

AustinSanders (dismissed):

AustinSanders (commented):
We'll have to work out the conflicts before merge, but the changes themselves look good to go.

AustinSanders (approved):

PR #487, opened 2022-09-09
Add Kaguyatc IsisLabel, Naifspice driver

acpaquette:
Adds the kaguyatcIsisLabelNaifSpice driver to ale for use with ISIS. Also fixes a few issues with the kaguyami load test


jlaura:
@acpaquette  I pulled this PR local to test it on KaguyaTC spiceinited cubes. I am getting the following error from the new driver:

```bash
Trying <class 'ale.drivers.selene_drivers.KaguyaTcIsisLabelNaifSpiceDriver'>
Failed: Failed to find metakernels. mission: selene, year:2007, versions="latest" spice root = "/isisdata/"
```

Why is the `KaguyaTcIsisLabelNaifSpiceDriver` looking for metakernels? The kernels are already listed on the ISIS cube.

jessemapel:
@jlaura The drivers do not have any kernel selection logic in them. That is the job of SpiceQL. It sounds like you want the IsisLabelIsisSpice driver.

jlaura:
Yup, I realize now that this driver is not going to work either. I do need the ISISSpice ISISLabel driver for these data.

ale/base/label_isis.py
```diff
@@ -170,14 +170,22 @@ def spacecraft_clock_start_count(self):
...
+            if 'SpacecraftClockStartCount' in self.label['IsisCube']['Instrument']:
+                self._clock_start_count = self.label['IsisCube']['Instrument']['SpacecraftClockStartCount']
+            elif 'SpacecraftClockCount' in self.label['IsisCube']['Instrument']:
+                self._clock_start_count = self.label['IsisCube']['Instrument']['SpacecraftClockCount']
+            elif 'SpacecraftClockStartCount' in self.label['IsisCube']['Archive']:
+                self._clock_start_count = self.label['IsisCube']['Instrument']['SpacecraftClockStartCount']
+            else:
+                self._clock_start_count = None
+
+            if isinstance(self._clock_start_count, pvl.Quantity):
```
jlaura, line 183: Is this because of a potential PVL version change or can we get back different types?

ale/base/label_isis.py
```diff
@@ -190,12 +198,20 @@ def spacecraft_clock_stop_count(self):
...
+                self._clock_stop_count = self.label['IsisCube']['Instrument']['SpacecraftClockStopCount']
+            elif 'SpacecraftClockStopCount' in self.label['IsisCube']['Archive']:
+                self._clock_stop_count = self.label['IsisCube']['Archive']['SpacecraftClockStopCount']
+            else:
+                self._clock_stop_count = None
+
+            if isinstance(self._clock_stop_count, pvl.Quantity):
+                self._clock_stop_count = self._clock_stop_count.value
+
+            self._clock_stop_count = str(self._clock_stop_count)
```
jlaura, line 212: Does it make sense to have a generic accessor for start/stop times and then call that from these properties with an appropriate arg? Looks like a fair bit of duplicative code, but perhaps that is more readable and preferable?

ale/drivers/selene_drivers.py
```diff
@@ -24,6 +24,30 @@ def instrument_id(self):
...
+          Line exposure duration
+        """
+        # It's a list, but only sometimes.
+        # seems to depend on whether you are using the original zipped archives or
+        # if its downloaded from JAXA's image search:
+        # (https://darts.isas.jaxa.jp/planet/pdap/selene/product_search.html#)
+        try:
+            return self.label['IsisCube']['Instrument']['LineSamplingInterval'][0].value * 0.001 # Scale to seconds
+        except:
+            return self.label['IsisCube']['Instrument']['LineSamplingInterval'].value * 0.001  # Scale to seconds
```
jlaura, line 49: Ouch! Different files based on DL location? That stinks!

ale/drivers/selene_drivers.py
```diff
@@ -278,7 +295,7 @@ def ephemeris_start_time(self):
         : float
           ephemeris start time of the image
         """
-        return spice.sct2e(self.spacecraft_id, self.spacecraft_clock_start_count)
+        return spice.sct2e(self.spacecraft_id, float(self.spacecraft_clock_start_count))
```
jlaura, line 298: Does it make sense to cast here or should the spacecraft_clock_start_count always be a float? So, cast in that property.

ale/drivers/selene_drivers.py
```diff
@@ -520,33 +530,302 @@ def detector_start_sample(self):
...
+        """
+        Returns the name of the platform containing the sensor. This is usually
+        the spacecraft name.
+
+        Returns
+        -------
+        : str
+          Name of the platform which the sensor is mounted on
+        """
+        return self.label['IsisCube']['Instrument']['MissionName']
```
jlaura, line 607: Shouldn't one of these be spacecraft name and the other mission name? I remember these labels being a little weird in that regard.

ale/drivers/selene_drivers.py
```diff
@@ -520,33 +530,302 @@ def detector_start_sample(self):
...
+        be defined. This must be the integer Naif Id code for the spacecraft. Expects
+        spacecraft_clock_start_count to be defined. This must be a string
+        containing the start clock count of the spacecraft
+
+        Returns
+        -------
+        : double
+          Starting ephemeris time of the image
+        """
+        return spice.sct2e(self.spacecraft_id, float(self.spacecraft_clock_start_count))
```
jlaura, line 626: Same float question as above.

ale/drivers/selene_drivers.py
```diff
@@ -520,33 +530,302 @@ def detector_start_sample(self):
...
+        return spice.gdpool('INS{}_CENTER'.format(self.ikid), 0, 2)[1] - 0.5
+
+    @property
+    def detector_center_sample(self):
+        """
+        Returns the center detector sample of the detector. Expects ikid to be
+        defined. This should be the NAIF integer ID code for the sensor.
+
+        We subtract 0.5 from the center sample because as per the IK:
+        Center of the first pixel is defined as "1.0".
```
jlaura, line 722: and ISIS needs to recenter to 0.5, correct?

jlaura:
@acpaquette Sorry for the delay!

ale/base/label_isis.py
jessemapel: PvL returns different types depending on if it's just a value or a value with units.

jessemapel: Note the difference between:

```
self._clock_start_count
```
and
```
self._clock_start_count.value
```

ale/base/label_isis.py
jessemapel: This code is part of the generic start/stop time accessors. There are a variety of different ways stop times are computed/recorded on the labels. This property handles the spacecraft clock stop count which can then be turned into the stop time by a NAIF call. See [NaifSpice::ephemeris_stop_time](https://github.com/USGS-Astrogeology/ale/blob/master/ale/base/data_naif.py#L495).

We could write a generic label search function. There is a lot of repeated code in this class to handle the multiple different places in a label a value can appear.

ale/drivers/selene_drivers.py
jessemapel: Space craft clock counts are actually strings most of the time:

https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/req/time.html#Spacecraft%20Clock%20SCLK

What's happening here is that we're calling sct2e which takes space craft clock ***ticks***. There is some inconsistency in labels between the spacecraft clock counts being strings or floats. If it's a string we should be using [scs2e](https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/cspice/scs2e_c.html) instead of [sct2e](https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/cspice/sct2e_c.html).

ale/base/label_isis.py
acpaquette: Agree with Jesse here, it also may not be worth it as accessing some of these values is so specific that a generic function could only work in a few instances.

It may be worth exploring but some duplicate code is so small that maintaining a function that gets used in like two places may be just as annoying to maintain as the duplicate code

ale/drivers/selene_drivers.py
acpaquette: See response from Jesse

ale/drivers/selene_drivers.py
acpaquette: Yes, and knoten handles the conversion if are comparing between ISIS and USGSCSM

ale/drivers/selene_drivers.py
acpaquette: I will look into this

ale/drivers/selene_drivers.py
```diff
@@ -520,33 +549,302 @@ def detector_start_sample(self):
...
+        We subtract 0.5 from the center line because as per the IK:
+        Center of the first pixel is defined as "1.0".
+
         Returns
         -------
         : int
-          ISIS sensor model version
+          The detector line of the principle point
+        """
+        return spice.gdpool('INS{}_BORESIGHT_LINE'.format(self.ikid), 0, 1)[0]
```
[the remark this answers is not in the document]
jlaura: or is something be handled in ISIS/knoten/??. If so, what if I access direct via swigcsm? My big question is just clarity on whether or not a -0.5 offset needs to be applied and whose responsibility it is to apply it.

ale/drivers/selene_drivers.py
```diff
@@ -520,33 +549,302 @@ def detector_start_sample(self):
...
+        Returns the x coefficients of the optical distortion model.
+        Expects tc_id to be defined. This should be a string of the form
+        LISM_TC1 or LISM_TC2.
+
+        Returns
+        -------
+        : list
+          Optical distortion x coefficients
         """
+        return spice.gdpool('INS{}_DISTORTION_COEF_X'.format(self.ikid),0, 4).tolist()
```
jlaura, line 762: Same spice caching question down the rest of this file.

ale/drivers/selene_drivers.py
```diff
@@ -520,33 +549,302 @@ def detector_start_sample(self):
...
+        : list
+          Optical distortion y coefficients
+        """
+        return spice.gdpool('INS{}_DISTORTION_COEF_Y'.format(self.ikid), 0, 4).tolist()
+
+    @property
+    def boresight_x(self):
+        """
+        Returns the x focal plane coordinate of the boresight.
+        Expects ikid to be defined. This should be the NAIF integer ID for the
```
jlaura, line 782: Do we needs these expects? The API called doesn't need to know this information because this class has an ikid property. The code is explicit that self.ikid is needed. Maybe these 'expects` in the docstring can be removed?

ale/drivers/selene_drivers.py
```diff
@@ -694,7 +992,7 @@ def ephemeris_start_time(self):
         : float
           ephemeris start time of the image
         """
-        return spice.sct2e(self.spacecraft_id, self.spacecraft_clock_start_count)
+        return spice.scs2e(self.spacecraft_id, self.spacecraft_clock_start_count)
```
jlaura, line 995: A doc string comment here or in spacecraft_clock_start_count that this needs to be a string (and not numeric ticks) might be a good idea?

jlaura (approved):

ale/drivers/selene_drivers.py
```diff
@@ -97,7 +121,23 @@ def boresight_y(self):
...
+        Calculated using 1/pixel pitch
+        Expects tc_id to be defined. This should be a string of the form
+        LISM_TC1 or LISM_TC2.
+
+        Returns
+        -------
+        : list
+          focal plane to detector lines
+        """
+        focal2pixel_lines = super().focal2pixel_lines
```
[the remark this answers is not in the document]
acpaquette: In this case no, the focal2pixel values come from whatever spice mixin is being used

ale/drivers/selene_drivers.py
```diff
@@ -201,8 +236,10 @@ def instrument_host_name(self):
         : str
           Spacecraft name as a proxy for instrument host name.
         """
-        return self.label.get("SPACECRAFT_NAME", None)
-
+        try:
+          return super().instrument_host_name
```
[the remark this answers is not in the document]
acpaquette: Similar to above but the instrument_host_name will only ever be exposed by whatever label mixin is being used

ale/drivers/selene_drivers.py
```diff
@@ -341,11 +376,7 @@ def focal2pixel_lines(self):
           focal plane to detector lines
         """
         pixel_size = spice.gdpool('INS{}_PIXEL_SIZE'.format(self.ikid), 0, 1)[0]
-        if self.spacecraft_direction < 0:
-            return [0, -1/pixel_size, 0]
-        elif self.spacecraft_direction > 0:
-            return [0, 1/pixel_size, 0]
-
+        return [0, 1/pixel_size, 0]
```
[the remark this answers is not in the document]
acpaquette: Correct, that is what partially drove the issues with KaguyaTC

ale/drivers/selene_drivers.py
```diff
@@ -201,8 +236,10 @@ def instrument_host_name(self):
         : str
           Spacecraft name as a proxy for instrument host name.
         """
-        return self.label.get("SPACECRAFT_NAME", None)
-
+        try:
+          return super().instrument_host_name
+        except:
+          return self.label.get("MISSION_NAME", None)
```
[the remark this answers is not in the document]
acpaquette: I think it can be removed?

ale/drivers/selene_drivers.py
```diff
@@ -520,33 +549,302 @@ def detector_start_sample(self):
...
+        """
+        start_sample = 1
+        swath_mode = str(self.label["IsisCube"]["Instrument"]["SwathModeId"])
+        if (swath_mode == "FULL"):
+          start_sample = 1
+        elif (swath_mode == "NOMINAL"):
+          start_sample = 297
+        elif (swath_mode == "HALF"):
+          start_sample = 1172;
+        return start_sample - 0.5
```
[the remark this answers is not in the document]
acpaquette: In this case we select the start_sample then subtract 0.5 from whatever is selected. So they all get subtracted by 0.5 in the end

ale/drivers/selene_drivers.py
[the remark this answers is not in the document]
acpaquette: The pixel difference between ISIS and CSM is really only at the pixel level while this w.r.t. the detector. In that case, both ISIS and CSM are likely going to agree on a given value. In this case it's the BORESIGHT_LINE with no modifications there after

ale/drivers/selene_drivers.py
acpaquette: They were/are a way to tell people what other properties a property relies on, since many of the properties have default implementations

ale/drivers/selene_drivers.py
acpaquette: There is actually a difference in some spacecraft. In Kaguya TC it has more to do with the formatting of the `spacecraft_clock_start_count`. When `spacecraft_clock_start_count` is in the "integer" format and you try to wrap it as a string the string routine is unhappy with the value. 

The default assumption is that the `spacecraft_clock_start_count` will be the string formatted version. For KaguyaTC it was the integer format which required the change. Kaguya MI which is where this change is, seems to have the string formatted `spacecraft_clock_start_count`

ale/drivers/selene_drivers.py
acpaquette: I was able to remove the `spacecraft_name` override here so that resolves this

ale/drivers/selene_drivers.py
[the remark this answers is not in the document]
acpaquette: This is a case of old KaguyaTC labels vs new. I pulled a new label but now I have to regenerate the sliced kernels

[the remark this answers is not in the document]
acpaquette: Removed!

ale/drivers/selene_drivers.py
jlaura: I see that, but this is a 3 line function and the code clearly needs the ikid.

jlaura (approved):

PR #492, opened 2022-10-18
Msl Focal2Pixel Fix

acpaquette:
Fixed value order for focal2pixel_samples vector


Kelvinrr:
Tested locally and was able to csminit

jessemapel:
@Kelvinrr @acpaquette This looks good, but it seems the test data needs to be updated.

jessemapel (approved):
Python 3.7 environment is struggling. I'm not sure why, but it's not supported by conda-forge anymore, so we could be on our own there.

PR #501, opened 2022-10-24
Adds IsisDataIsisLabel driver for Kaguya

### LINKED ISSUE #493: Add Kaguya TC from ISIS Cube Driver
According to #375 ALE can generate a valid ISD from the PDS3 label, but not from the ISIS label. There are instances when it is necessary to use the ISIS cube (for example after adjustment). Please add a Kaguya TC driver (for all 3 flavors of TC: monoscopic, stereoscopic, and SP support) that accepts the ISIS cube as the input.

jlaura:
Adds a KaguyaTC IsisData IsisLabel driver that supports TC1 and TC2 data.

As indicated in #493, I have some questions about how best to handle the need to furnish kernels inorder to get the distortion coefficients. I looked at the MDIS driver thinking it might furnish kernels, but I did not see how the IsisData IsisLabel was furnishing the IKs.

Fixes #493.

Marked as draft because this needs tests. I'd appreciate a review on the implementation to know if I need to furnish kernels in a different way.

### DIFF
--- a/ale/drivers/selene_drivers.py
+++ b/ale/drivers/selene_drivers.py
@@ -3,13 +3,300 @@ from glob import glob
 import numpy as np
 import spiceypy as spice
 
+from ale import util
 from ale.base import Driver
 from ale.base.data_naif import NaifSpice
+from ale.base.data_isis import IsisSpice
 from ale.base.label_pds3 import Pds3Label
 from ale.base.label_isis import IsisLabel
 from ale.base.type_sensor import LineScanner
 
 
+class KaguyaTcIsisLabelIsisSpiceDriver(LineScanner, IsisLabel, IsisSpice, Driver):
+    """
+    """
+    # Comes from IsisSpice
+    '''@property
+    def instrument_id(self):
+        """
+        Id takes the form of LISM_<INSTRUMENT_ID>_<SD><COMPRESS><SWATH> where
+
+        INSTRUMENT_ID = TC1/TC2
+        SD = S/D short for single or double, which in turn means whether the
+        label belongs to a mono or stereo image.
+
+        COMPRESS = D/T short for DCT or through, we assume image has been
+        decompressed already
+
+        SWATCH = swatch mode, different swatch modes have different FOVs
+
+        Returns
+        -------
+        : str
+          instrument id
+        """
+        instrument = super().instrument_id
+        swath = self.label["IsisCube"]["Instrument"]["SwathModeId"][0]
+        sd = self.label["IsisCube"]["Archive"]["ProductSetId"].split("_")[1].upper()
+
+        inst_id = "LISM_{}_{}T{}".format(instrument, sd, swath)
+        return inst_id'''
+
+    # In data_isis
+    '''@property
+    def sensor_frame_id(self):
+        """
+        Returns the sensor frame id.  Depends on the instrument that was used to
+        capture the image.
+
+        Returns
+        -------
+        : int
+          Sensor frame id
+        """
+        return spice.namfrm("LISM_{}_HEAD".format(super().instrument_id))'''
+    
+    # In data_isis
+    '''@property
+    def ikid(self):
+        """
+        Read the ikid from the cube label
+        """
+        if not hasattr(self, "_ikid"):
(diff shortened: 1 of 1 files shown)


ale/drivers/selene_drivers.py
```diff
@@ -3,13 +3,300 @@
...
+        Expects tc_id to be defined. This should be a string of the form
+        LISM_TC1 or LISM_TC2.
+
+        Returns
+        -------
+        : list
+          Optical distortion x coefficients
+        """
+        self.kernels
+        return spice.gdpool('INS{}_DISTORTION_COEF_X'.format(self.ikid),0, 4).tolist()
```
acpaquette, line 204: @jlaura Any spice call instance should be accessible via the ISIS label under the `NaifKeywords` group. Looking over a spiceinit'd kaguyatc cube, that group should look something like this:

Screen Shot 2022-10-24 at 3 05 06 PM

Where the keyword has already been pulled out of the kernel, all you need to do is access it

ale/drivers/selene_drivers.py
```diff
@@ -3,13 +3,300 @@
...
+
+    @property
+    def kernels(self):
+      if not hasattr(self, '_kernels'):
+        self._kernels = util.get_kernels_from_isis_pvl(self.label['IsisCube'])
+        print(self._kernels)
+        [spice.furnsh(k) for k in self._kernels]
```
acpaquette, line 211: If the previous comment is handled this should just be able to go away. Typically we use the kernels property to load kernels passed in via the `props` argument via a context

ale/drivers/selene_drivers.py
```diff
@@ -3,13 +3,300 @@
...
+
+        """
+        return {
+            "kaguyalism": {
+                "x" : self._odkx,
+                "y" : self._odky,
+                "boresight_x" : self.boresight_x,
+                "boresight_y" : self.boresight_y
+            }
+        }
```
acpaquette, line 294: This distortion should probably get pulled out into it's own mixin but can be merged as is

acpaquette (changes requested):
Overall looks good, most of the commented out properties can go but it's nice to see what was removed from the copied driver. Once the spice calls are fixed this can be merged.

If the distortion mixin is addressed then that's just bonus points, but if not we can just make a new issue

acpaquette:
@jlaura Best way to handle something like that is to pull it into its own property on the concrete driver. Something like:
```
@property
def swath_mode(self):
    return swath_mode
```
It's a little janky but it will allow you to mock that element and return whatever value you want from the label.

Another option would be to replace the last bracket accesser with a get call. Then you could mock the get call to return whatever.

Both options aren't ideal and inject some weird code smell but either will accomplish what you are looking to test

jlaura:
@acpaquette Updated. I went with a `_swath_mode` property that is easily mocked and is then consumed by the `detector_start_sample`.

ale/drivers/__init__.py
```diff
@@ -125,7 +125,7 @@ def load(label, props={}, formatter='ale', verbose=False):
             res = driver(label, props=props, parsed_label=parsed_label)
             # get instrument_id to force early failure
             res.instrument_id
-
+            print(res.instrument_id)
```
acpaquette, line 128: One last print to remove and this PR is golden!

acpaquette (commented):
One small thing to handle

acpaquette (approved):

jlaura:
@jessemapel  I assume this has to fail out if not 'TC1' or 'TC2'? I see the Messenger ISIS/ISIS driver using a lookup and the TGO-CaSSIS driver using a lookup implemented slightly differently. Would something like the following work?


```python
INSTRUMENT_LIST = ['TC1', 'TC2']
@property
def instrument_id(self):
    iid = self.label['IsisCube']['Instrument']['InstrumentId']
    if iid not in INSTRUMENT_LIST:
        raise ValueError(f'{iid} not in {INSTRUMENT_LIST})
    return iid
```

jessemapel:
Yeah, we use the instrument_id property as a first check if a driver is valid so that we can quickly iterate through them in load.

PR #509, opened 2022-11-15
ALE Load Driver Selection

acpaquette:
Initial stab at exclusively selecting drivers based on component classes.

Still needs:
- [x] tests
- [x] exposed on the C++ side

### DIFF
--- a/ale/drivers/__init__.py
+++ b/ale/drivers/__init__.py
@@ -54,7 +54,7 @@ class AleJsonEncoder(json.JSONEncoder):
             return obj.isoformat()
         return json.JSONEncoder.default(self, obj)
 
-def load(label, props={}, formatter='ale', verbose=False, only_isis_label=False,  only_isis_spice=False,  only_pds3_label=False,  only_naif_spice=False):
+def load(label, props={}, formatter='ale', verbose=False, only_isis_label=False, only_isis_spice=False, only_pds3_label=False, only_naif_spice=False):
     """
     Attempt to load a given label from possible drivers.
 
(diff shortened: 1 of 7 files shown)


ale/drivers/__init__.py
```diff
@@ -82,15 +100,39 @@ def load(label, props={}, formatter='ale', verbose=False):
...
     """
+    print(only_isis_label,  only_isis_spice,  only_pds3_label,  only_naif_spice)
     if isinstance(formatter, str):
         formatter = __formatters__[formatter]
-
-    drivers = chain.from_iterable(inspect.getmembers(dmod, lambda x: inspect.isclass(x) and "_driver" in x.__module__) for dmod in __driver_modules__)
+    
+    driver_mask = [only_isis_label, only_pds3_label, only_isis_spice, only_naif_spice]
+    class_list = np.array([IsisLabel, Pds3Label, IsisSpice, NaifSpice])
+    class_list = list(class_list[driver_mask])
```
jlaura, line 130: style suggestion, please ignore.

what about using a list comprehension or itertools.compress instead of np? Just because the next long call in here is that chain.from_iterable one.

```python
>>> x = ['a', 'b', 'c']
>>> y = [True, True, False]
>>> from itertools import compress
>>> compress(x, y)
<itertools.compress object at 0x7f7ff0827a60>
>>> list(compress(x, y))
['a', 'b']
```

jlaura (approved):

ale/drivers/__init__.py
acpaquette: I wanted to avoid numpy here so this is great!

jlaura:
@acpaquette I really like the interface on this compared to the original! :+1:

PR #510, opened 2022-11-22
transition to new build system for Python 3.10+

antonhibl:
Adds setup.cfg and pyproject.toml files that allow for building with pip in accordance with the recommended changes in PEP 518, PEP 517, PEP 620, and PEP 660. These files contains build system requirements, dependencies, metadata, entry point methods and more information which is used by pip to build the package. the command to utilize this new build system is currently: `pip3 install --trusted-host files.pythonhosted.org --trusted-host pypi.org --trusted-host pypi.python.org -e .`

Going forward, eventually setup.py will be depreciated so this is the first step to transitioning to the new recommended build system to replace setup.py. More changes are probably needed to optimize building but it seems to build and install correctly using the new system with these additions. The main purpose of this for now is to add support to Python 3.10 and 3.11 in a more robust way which supports their recommended build systems.

### DIFF
--- a/ale/drivers/__init__.py
+++ b/ale/drivers/__init__.py
@@ -106,7 +106,6 @@ def load(label, props={}, formatter='ale', verbose=False, only_isis_spice=False,
     dict
          The ISD as a dictionary
     """
-    print("Banana", only_isis_spice, only_naif_spice)
     if isinstance(formatter, str):
         formatter = __formatters__[formatter]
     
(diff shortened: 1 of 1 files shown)


antonhibl:
@acpaquette any idea what the reason might be these tests are failing?  it doesn't make much sense unless that new setup is being used to install ale before testing in the CI, I only added 2 files(setup.cfg and pyproject.toml) that are only used with a specific pip command.

acpaquette:
@antonhibl Yes, this seems to be failing due to the CI setup. You can change how the CI installs the package in the ci_testing.yml under .github/workflows

acpaquette:
NVM, the setup.py file is still there. Not sure why this is happening. Have you tried running locally first?

antonhibl:
@acpaquette yeah I ran and installed Ale just fine it seems with the new setup.cfg and pyproject.toml with pip on mac, I tested in some notebooks to and it seemed everything was working. I can also test on Ubuntu if necessary.

PR #514, opened 2023-01-26
Some potential fixes for MSL

oleg-alexandrov:
I've been working with @acpaquette looking at some issues with the MSL implementation in ALE. I think I found a handful of potential issues, and the results become plausible after they are fixed. This will need a review and some more testing. The summary of fixes is the following, corresponding to the changes in the files:
 
 - Flipped the sign of two rows of the rotation matrix. Before this fix, the image was projecting strangely, with the sky projecting close to the camera and the ground projecting in the distance. I believe this is needed because the [paper we follow](https://agupubs.onlinelibrary.wiley.com/doi/epdf/10.1029/2003JE002199) uses the convention that x and y change sign (see equation (8a) and (8b)) and I think CSM does not do that.
 - Added 0.5 to optical center, because CSM uses the convention that upper-left pixel is (0.5, 0.5). I know other ALE drivers do that too. This likely needs some more thinking.
 - Fixed the functions focal2pixel_lines() and focal2pixel_samples(). These were not swapping x and y after going from line/sample to distorted sensor pixels. The fixed approach is consistent with the Dawn driver and what I had to do for DigitalGlobe CSM logic. Before this fix, projecting onto the ground was resulting in the sky being sideways. 
 - Fixed the functions detector_center_sample() and detector_center_line(). These were actually correct in the Cahvor class, but were overridden in the MslMastcamPds3NaifSpiceDriver class by some generic spice function which is not aware of the Cahvor() implementation.
 - Fixed the functions detector_start_line() and detector_start_sample(). Now setting the start line and start sample to 0. These values are used in CSM's computeDistortedFocalPlaneCoordinates() function together with the detector center values to shift a pixel. Setting them to something non-zero is inconsistent with the CAHVOR implementation.

Here's the result of map-projecting onto the MOLA DEM the MSL images 1099ML0048660020500398C00 and 1099ML0048660010500397E01 with ASP's mapproject and updated CSM ISD cameras.

image

Things are as expected. In each image, the hole in the middle is where the camera is. The camera looks towards the lower-left corner. The sky is in the distance and the view expands that way. Some rays from the ground hit the sensor from behind, hence the junky thing in the upper-right. 

When these two images are overlaid with georeference info, they agree quite well, which is a good sign.

I think these changes go the right way. What is needed is to get a stereo pair with good convergence angle and with prior knowledge of the site to do a fool-proof sanity check.

I must say this work wasn't something assigned or funded. Spending a couple of days on this seems about right. If the changes make sense, hopefully more testing can be done. Later in the spring I am funded to do some CSM work when I can revisit this if need be.

### DIFF
--- a/ale/base/type_sensor.py
+++ b/ale/base/type_sensor.py
@@ -479,7 +479,10 @@ class Cahvor():
                                                       ephemeris_times=self.ephemeris_time,
                                                       nadir=False, exact_ck_times=False)
             cahvor_quats = Rotation.from_matrix(self.cahvor_rotation_matrix).as_quat()
-            cahvor_rotation = ConstantRotation(cahvor_quats, self.final_inst_frame, self.sensor_frame_id)
+            #cahvor_rotation = ConstantRotation(cahvor_quats, self.final_inst_frame, self.#sensor_frame_id)
+            cahvor_rotation = ConstantRotation(cahvor_quats, self.target_frame_id,
+                                               self.sensor_frame_id)
+
             self._frame_chain.add_edge(rotation = cahvor_rotation)
         return self._frame_chain
 
--- a/ale/base/type_sensor.py
+++ b/ale/base/type_sensor.py
@@ -397,7 +397,7 @@ class Cahvor():
     def compute_h_c(self):
         """
         Computes the h_c element of a cahvor model for the conversion
-        to a photogrametric model
+        to a photogrammetric model
 
         Returns
         -------
@@ -409,7 +409,7 @@ class Cahvor():
     def compute_h_s(self):
         """
         Computes the h_s element of a cahvor model for the conversion
-        to a photogrametric model
+        to a photogrammetric model
 
         Returns
         -------
@@ -421,7 +421,7 @@ class Cahvor():
     def compute_v_c(self):
         """
         Computes the v_c element of a cahvor model for the conversion
-        to a photogrametric model
+        to a photogrammetric model
 
         Returns
         -------
@@ -433,7 +433,7 @@ class Cahvor():
     def compute_v_s(self):
         """
         Computes the v_s element of a cahvor model for the conversion
-        to a photogrametric model
+        to a photogrammetric model
 
         Returns
         -------
@@ -460,6 +460,7 @@ class Cahvor():
             H_prime = (self.cahvor_camera_dict['H'] - h_c * self.cahvor_camera_dict['A'])/h_s
             V_prime = (self.cahvor_camera_dict['V'] - v_c * self.cahvor_camera_dict['A'])/v_s
             self._cahvor_rotation_matrix = np.array([H_prime, V_prime, self.cahvor_camera_dict['A']])
+        print("----cahvor rotation matrix is ", self._cahvor_rotation_matrix)
         return self._cahvor_rotation_matrix
 
     @property
@@ -489,6 +490,7 @@ class Cahvor():
         X = np.array(X.split(),  dtype=np.float64)
         X = X.reshape(3, 3)
         X = np.matmul(r.as_matrix(), X)
+        print("----rover frame rotation is ", X)
         return X
 
     @property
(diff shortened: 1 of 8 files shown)


ale/base/type_sensor.py
```diff
@@ -466,13 +473,16 @@ def frame_chain(self):
...
+            self._frame_chain = FrameChain.from_spice(sensor_frame=self.final_inst_frame,
                                                       target_frame=self.target_frame_id,
                                                       center_ephemeris_time=self.center_ephemeris_time,
                                                       ephemeris_times=self.ephemeris_time,
                                                       nadir=False, exact_ck_times=False)
             cahvor_quats = Rotation.from_matrix(self.cahvor_rotation_matrix).as_quat()
-            cahvor_rotation = ConstantRotation(cahvor_quats, self.spacecraft_id * 1000, self.sensor_frame_id)
+            #cahvor_rotation = ConstantRotation(cahvor_quats, self.final_inst_frame, self.#sensor_frame_id)
+            cahvor_rotation = ConstantRotation(cahvor_quats, self.target_frame_id,
+                                               self.sensor_frame_id)
```
acpaquette, line 484: @oleg-alexandrov This is overwriting the frame chain to short circuit from the target to the sensor, I take it as intentional?

oleg-alexandrov: This is what gives plausible results. After your fix, the orientation changed, but the camera points in the opposite direction of where that hill being imaged is, and there is a very large perspective distortion, and still some shift between the images. Adding my fix on top, the camera now points up, rather than to the left, where the hill is in the DEM, so that's wrong, but there is no shift between successive images and no perspective distortion. So, there is still some missing link in that frame chain.

oleg-alexandrov: Also note that I removed the temporary hack transform in an earlier commit, so things are more correct now. I also modified one sign in an earlier commit the cahvor_rotation_matrix() function. That was not returning a true rotation, as it had a negative determinant. So, now it really looks if we can get another rotation in the rover frame we should be good.

acpaquette: I guess I am more confused now, the `final_inst_frame` was something I added. In doing so I removed the hack rotation that was previous present

oleg-alexandrov: I see that my independent removal of the hack transform is not even in any history commits, which surely adds to the confusion. I think when I pulled your code I had a conflict with my local commit, and after fixing that things got squashed. Anyhow, I surely agree that your changes made that hack unnecessary so it was right for you to remove it, and we are now incrementally going the right way.

acpaquette: 👍 That all makes more sense. I am trying to get mapproject working again on my end right now. Running into package versioning issues 😬

oleg-alexandrov: Fetching ASP with conda in an existing ISIS env is pretty much guaranteed not to work unless your ISIS version is the same which is used with ASP, because every time ISIS gets bumped up packages change and there's so many of them. 

The easiest thing is to fetch and unzip a tarball, as described here: https://stereopipeline.readthedocs.io/en/latest/installation.html#precompiled-binaries-linux-and-macos

acpaquette: Will do, running into some weird path binary issues locally with opencv and x11. Worked on friday but an OS update may have caused some issues

oleg-alexandrov:
I pushed the latest code to this branch. 

There is a writeup with how current functionality can be used to create DEMs at: https://stereopipeline.readthedocs.io/en/latest/examples/csm.html#csm-msl

By now, I am quite sure that the following is correct:

- Creation and incorporation of Cahvor cameras
- Population of CSM model (I compared full model with ASP Pinhole and there's great agreement with cam_test)
- Creation of the rig for the navcam stereo pair

Future steps may be:

- Invest time in understanding the SPICE/NAIF documentation
- Find accurately the missing transform in the rover frame. The final transform from Mars to camera is: Cahvor * UnknownTransform * MarsToRover.
- Is it even a fixed transform or variable (given that the rover has moving parts)
- Understand the 60 m vertical jump for SOL597 (also test with SOL603; these are good datasets next to a notable ground feature, Kimberly outcrop).
 - Create a tool to plot the rover path and the various orientations on the Mars surface. Hard to do things without that. 
  
I am very happy to be kept in the loop and help when you folks need anything here.

oleg-alexandrov:
I played with things a bit more, and it is also unclear if the Mars-to-rover rotation should be applied as it is, or if its inverse should be used. So, the root of the problem seems to be in the rover frame, as all images acquired with a fixed rover agree with each other.

PR #517, opened 2023-02-03
MGS MOC Narrow Angle Camera

jrcain-usgs:
Added MGS MOC Narrow Angle Camera Isis Label Naif Spice Driver and Tests


ale/drivers/mgs_drivers.py
```diff
@@ -12,6 +12,140 @@
...
+        """
+        Returns
+        -------
+        : String
+          The name of the sensor
+        """
+        return self.instrument_id
+
+    @property
+    def ephemeris_start_time(self):
```
acpaquette, line 51: I don't think this function needs to be overwritten and should be handled by the IsisLabel and NaifSpice mixins

ale/drivers/mgs_drivers.py
```diff
@@ -12,6 +12,140 @@
...
+        """
+        Returns the center detector sample. Expects ikid to be defined. This should
+        be an integer containing the Naif Id code of the instrument.
+
+        Returns
+        -------
+        : float
+          Detector sample of the principal point
+        """
+        return float(spice.gdpool('INS{}_CENTER'.format(self.ikid), 0, 1)[0])
```
acpaquette, line 98: Should probably change this to getting 2 elements instead of 1 to be consistent with detector_center_line

acpaquette (commented):
Did a quick review of the driver with a few things to address. Great work!

ale/drivers/mgs_drivers.py
jrcain-usgs: It looks like I actually should have overwritten Spacecraft Clock Start Count here.

acpaquette: I guess I am still confused as to why? If you check how ISIS gets the start count it looks for the keyword you are looking for here:
https://github.com/USGS-Astrogeology/ale/blob/e689abf9e5391aa578156a2f957e1499daedb75e/ale/base/label_isis.py#L176

Does the label contain both `SpacecraftClockCount` and `SpacecraftClockStartCount`?

ale/drivers/mgs_drivers.py
jrcain-usgs: I'm only using element [0] in detector_center_sample, so shouldn't I just get one element, the first one?
In detector_center_line I get two elements because I need to get both in order to access the second one.

ale/drivers/mgs_drivers.py
jrcain-usgs: Oh, you're right.  I missed that, I thought it only looked for `SpacecraftClockStartCount`.  Removed with commit.

PR #518, opened 2023-02-06
initial UVVIS Driver notebook and code commit

antonhibl:
Start of notebook and driver for UVVIS image labels, needs more wokr and currently throws spiceinit error with clementine1 UVVIS cubes.


acpaquette:
This may be more of an ISIS issue than an ALE issue. I will try on my version of ISIS to see what I get

PR #523, opened 2023-03-14
Allow ALE NaifSpice Mixins to Access SpiceQL service

acpaquette:
This update enables the default NaifSpice mixin to access the SpiceQL service, with an example driver updated to test (MroCtxIsisLabelNaifSpiceDriver and MroCtxPds3LabelNaifSpiceDriver).

Both work as expected  and can be tested with:
`MroCtxIsisLabelNaifSpiceDriver("/Path/to/my/isis/cube.cub", props={"web": True})`
or
`MroCtxPds3LabelNaifSpiceDriver("/Path/to/my/pds3/label.[lbl|img], props={"web": True})`

Keep in mind that so far, this process is much slower over the net and we are still working to improve the speed of the SpiceQL service.


ale/base/data_naif.py
```diff
@@ -435,47 +443,74 @@ def sensor_position(self):
...
-                                                       "NONE")
-                    ssb_tar_state = np.array(list(ssb_tar.starg))
-                    state = ssb_tar_state - ssb_obs_state
-
-                    matrix = spice.sxform("J2000", self.reference_frame, time)
+            # spkezr returns a vector from the observer's location to the aberration-corrected
+            # location of the target. For more information, see:
+            # https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/FORTRAN/spicelib/spkezr.html
+            if self.correct_lt_to_surface and self.light_time_correction.upper() == 'LT+S':
+                print("BANANA", ephem)
```
jlaura, line 450: 🍌 ?

ale/base/data_naif.py
```diff
@@ -435,47 +443,74 @@ def sensor_position(self):
...
+                                                                "observer": observer,
+                                                                "frame": "J2000",
+                                                                "abcorr": self.light_time_correction,
+                                                                "mission": self.spiceql_mission})
+                # obs_tar = pyspiceql.getTargetState(time,
+                #                                     target,
+                #                                     observer,
+                #                                     'J2000',
+                #                                     self.light_time_correction)
+                obs_tar_lts = np.array(obs_tars)[:,-1]
```
jlaura, line 462: Why are these (and below) commented out?

ale/base/data_naif.py
```diff
@@ -685,44 +723,19 @@ def spiceql_mission(self):
...
+        Returns
+        -------
+        : json
+          Json data from the SpiceQL call
+        """
+        # Ideally this would work if a user passed no kernels but still
+        # set ISISDATA
+        if (not hasattr(self, "kernels")):
+            function_args["searchKernels"] = True
+        return kernel_access.spiceql_call(function_name, function_args, self.use_web)
```
jlaura, line 741: JSON is not a python return type. Is this a string that happens to be JSON formatted or is it a dict? The caller will want to know that and then can parse to JSON or not as appropriate.

ale/drivers/mro_drivers.py
```diff
@@ -344,23 +343,6 @@ def sensor_name(self):
         """
         return "CONTEXT CAMERA"
 
-    @property
```
jlaura, line 347: This seems a weird change to include in the PR? CHANGELOG entry might clear up if this is a bug fix for something else or what.

ale/kernel_access.py
```diff
@@ -295,3 +298,65 @@ def remove_dups(listofElements):
...
+                Any return from a SpiceQL function
+    """
+    if not use_web:
+        func = getattr(pyspiceql, function_name)
+
+        # Ensure that in offline mode we anticipate the user loading/passing their own kernels
+        # to ALE
+        return func(**function_args)
+
+    url = "https://spiceql-dev.prod-asc.chs.usgs.gov/v1/"
```
jlaura, line 332: Even though this is only an incremental step towards turning this on, I would strongly advocate using an ENV variable here to encode. That seems the most likely way a user might want to set / change this. Sure, this might eventually be hard coded, but I would advocate against that since it tightly couples this ALE version to a specific spiceql server version.

ale/kernel_access.py
```diff
@@ -295,3 +298,65 @@ def remove_dups(listofElements):
...
+    url += function_name
+    headers = {
+        'accept': '*/*',
+        'Content-Type': 'application/json'
+    }
+
+    r = requests.get(url, params=function_args, headers=headers, verify=False)
```
jlaura, line 339: Yes, this verify=False is needed on a DOI network if the SSL certificates are not in place, but out in the wild this is a bad idea.

ale/kernel_access.py
```diff
@@ -295,3 +298,65 @@ def remove_dups(listofElements):
...
+    r.raise_for_status()
+    response = r.json()
+
+    if r.status_code != 200:
+        raise requests.HTTPError(f"Recieved code {response['statusCode']} from spice server, with error: {response}")
+    elif response["statusCode"] != 200:
+        raise requests.HTTPError(f"Recieved code {response['statusCode']} from spice server, with error: {response}")
+    return response["body"]["return"]
+
+    # Code for accessing SpiceQL docker container
```
jlaura, line 349: This can be removed. Put it into docs?

ale/transformation.py
```diff
@@ -123,15 +124,20 @@ def from_spice(cls, sensor_frame, target_frame, center_ephemeris_time, ephemeris
 
         constant_frames.extend(target_constant_frames)
 
-        frame_chain.compute_time_dependent_rotiations(sensor_time_dependent_frames, sensor_times)
-        frame_chain.compute_time_dependent_rotiations(target_time_dependent_frames, target_times)
+        # Add all time dependent frame chains to the graph
```
jlaura, line 127: It seems that `use_web` is becoming ubiquitous in function signatures. Have other options been considered to not extend the signatures? For example, a module level variable or something?

jlaura (changes requested):
I think this needs:
- CHANGELOG entries
- tests?
- thought about the use of `use_web` all over the place now
- tests that handle edge cases introduced by having 3 spice sources or docs describing how this was tested. My concern is unintended breakages due to the new features.

ale/base/data_naif.py
acpaquette: What, got a problem with fruit in your code? 😄

ale/base/data_naif.py
acpaquette: I left these here to remind myself of what was being replaced, they could be removed

ale/base/data_naif.py
acpaquette: It's a dict, brain was havin a hard time

ale/kernel_access.py
acpaquette: 100% agree, I think providing some way to set it then defaulting might be a good middle ground?

ale/kernel_access.py
acpaquette: Both good catches!

ale/kernel_access.py
```diff
@@ -295,3 +298,65 @@ def remove_dups(listofElements):
...
+    headers = {
+        'accept': '*/*',
+        'Content-Type': 'application/json'
+    }
+
+    r = requests.get(url, params=function_args, headers=headers, verify=False)
+    r.raise_for_status()
+    response = r.json()
+
+    if r.status_code != 200:
```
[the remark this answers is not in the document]
acpaquette: Kinda sorta, this is a weird quirk of our lambda setup that I think needs to change. Basically if the lambda function called fails, we still get a "valid" web return as we catch the failure.

I think we can just let the function itself fail and that should be handled as expected(?)

It may also be that all of our endpoints only handle returning 200 codes.

This likely needs to be ironed out before a full release

ale/kernel_access.py
acpaquette: Likely, ideally the interfaces would be unified but keeping this is nice for testing. Likely goes in docs as you mention

ale/transformation.py
acpaquette: The function structure is definitely scuffed. Currently the `use_web` variable is in the NaifSpice mixin, which makes some sense as you want it on a driver to driver level.

Then we have the `kernel_access.py` module with the `spiceql_call` wrapper. This wrapper function is then further wrapped in the `NaifSpice` mixin where the `use_web` variable is grabbed from the driver when the function is called.

The issue here is that `transformation.py` and specifically the `FrameChain.from_spice` code is not part of the driver and exists to support getting the reference frame rotations. This requires that either a variable be passed through or that the FrameChain itself assigns it's own `use_web` variable which may actually be worth while.

So, to answer your question, I had not considered other options but I don't think a module level define is good. But I think we can make it that any FrameChain constructed using `from_spice` can set a member variable defining web usage.

acpaquette:
@jlaura This has not been fully tested yet as much of the mocking that happens in the test was not addressed. Basically anything that was mocking a spice call needs to now mock the `spiceql_call` function. Something that will be addressed in a future sprint where I will go in and basically fix every current NaifSpice driver/tests.

ale/transformation.py
```diff
@@ -123,15 +124,20 @@ def from_spice(cls, sensor_frame, target_frame, center_ephemeris_time, ephemeris
...
-            quat_from_rotation = spice.m2q(rotation_matrix)
-            quats[:3] = quat_from_rotation[1:]
-            quats[3] = quat_from_rotation[0]
+            quat_and_av = spiceql_call("getTargetOrientations",{"ets": str([ephemeris_times[0]]),
+                                                                "toFrame": d,
+                                                                "refFrame": s,
+                                                                "mission": mission},
+                                                                use_web=use_web)[0]
+            quats[:3] = quat_and_av[1:4]
+            quats[3] = quat_and_av[0]
```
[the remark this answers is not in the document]
acpaquette: When software development is built on hopes and prayers 😉

ale/kernel_access.py
acpaquette: Revisiting this. Isn't `requests` is a default python library?

ale/spiceql_access.py
```diff
@@ -0,0 +1,198 @@
+import asyncio
+import math
+import requests
+
+import aiohttp
```
Kelvinrr, line 5: Needs to be added to environment.yml

ale/kernel_access.py
jlaura: No, it is not. If you want to use just standard library, use urllib (though it does not have the same nice to have stuff like urllib3 or requests).

ale/spiceql_access.py
```diff
@@ -0,0 +1,198 @@
...
+        return func(**function_args)
+    url = "https://spiceql-dev.prod-asc.chs.usgs.gov/v1/"
+    url += function_name
+    headers = {
+        'accept': '*/*',
+        'Content-Type': 'application/json'
+    }
+
+    clean_function_args = stringify_web_args(function_args)
+    async with session.get(url, params=clean_function_args, headers=headers, ssl=False) as response:
```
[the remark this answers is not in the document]
jlaura: ssl=False should be removed. Do we really want to support running without SSL?

(If this is an internal network/custom certificate thing, we can setup our work stations to have the proper self-signed certs.)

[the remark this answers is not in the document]
Kelvinrr: almost every tool I use on my command line has some kind of disable-ssl option.

Kelvinrr (commented):
Probably need to wait until. https://github.com/conda-forge/staged-recipes/pull/24039 is merged to merge this.

ale/base/__init__.py
```diff
@@ -1 +1,90 @@
 from ale.base.base import Driver
+
+spiceql_mission_map = {
+    "CHANDRAYAAN-1_M3": "m3",
+    "CHANDRAYAAN-1_MRFFR": "mrffr",
+    "CASSINI_ISS_NAC": "cassini",
+    "CASSINI_ISS_WAC": "cassini",
```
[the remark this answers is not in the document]
acpaquette: It could go in spiceql, the translation has to exist somewhere. I think it's better off in ALE since ALE is using the spiceql interface rather than spiceql conforming to ALE.

Granted all of the keys are from NAIF so :shrug:

ale/base/base.py
```diff
@@ -342,13 +408,19 @@ def projection(self):
...
             else: 
                 # should be a path
                 if not os.path.exists(self._file): 
                     self._projection = "" 
                 else: 
-                    geodata = gdal.Open(self._file)
+                    try:
+                        geodata = gdal.Open(self._file)
+                    except:
+                        geodata = None
```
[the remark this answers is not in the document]
acpaquette: It may be worth reporting the exception to the user as a warning. When working with pds3 labels I was running into issues where the label was pointing to a data file causing the gdal.Open to fail but the label could be used as valid input for a driver.

ale/drivers/selene_drivers.py
```diff
@@ -1092,6 +1175,46 @@ def line_exposure_duration(self):
         except:
             return self.label['CORRECTED_SAMPLING_INTERVAL'].value * 0.001  # Scale to seconds
 
+    @property
+    def usgscsm_distortion_model(self):
+        """
+        Kaguya uses a unique radial distortion model so we need to overwrite the
+        method packing the distortion model into the ISD.
```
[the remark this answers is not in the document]
acpaquette: I saw this too and wasn't sure why this was part of the PR. I need to double check this and see if it was missed from a rebase

Kelvinrr:
Closing in favor of https://github.com/DOI-USGS/ale/pull/621

PR #524, opened 2023-03-29
added projection code

Kelvinrr:
Adds ability to add projections to usgscsm ISDs, part of the EIS projection support work.

### DIFF
--- a/ale/base/base.py
+++ b/ale/base/base.py
@@ -333,7 +333,7 @@ class Driver():
             try: 
               from osgeo import gdal 
             except: 
-                self._projection = None
+                self._projection = ""
                 return self._projection
 
             if isinstance(self._file, pvl.PVLModule):
@@ -346,7 +346,7 @@ class Driver():
             else: 
                 # should be a path
                 if not os.path.exists(self._file): 
-                    self._projection = None 
+                    self._projection = "" 
                 else: 
                     geodata = gdal.Open(self._file)
                     self._projection = geodata.GetSpatialRef()
@@ -354,7 +354,8 @@ class Driver():
             # is None if not projected
             if self._projection: 
                 self._projection = self._projection.ExportToProj4()
-                
+            else: 
+                self._projection = "" 
         return self._projection
     
 
--- a/ale/base/base.py
+++ b/ale/base/base.py
@@ -1,6 +1,9 @@
 import pvl
 import json
 
+import tempfile
+import os 
+
 class Driver():
     """
     Base class for all Drivers.
@@ -323,3 +326,33 @@ class Driver():
     @property
     def short_mission_name(self):
         return self.__module__.split('.')[-1].split('_')[0]
+
+    @property 
+    def projection(self):
+        if not hasattr(self, "_projection"): 
+            try: 
+              from osgeo import gdal 
+            except: 
+                self._projection = None
+                return self._projection
+
+            if isinstance(self._file, pvl.PVLModule):
+                # save it to a temp folder
+                with tempfile.NamedTemporaryFile() as tmp:
+                    tmp.write(pvl.dumps(self._file)) 
+
+                    geodata = gdal.Open(tempfile.name)
+                    self._projection = geodata.GetSpatialRef()
+            else: 
+                # should be a path
+                if not os.path.exists(self._file): 
+                    self._projection = None 
+                else: 
+                    geodata = gdal.Open(self._file)
+                    self._projection = geodata.GetSpatialRef()
+
+            # is None if not projected
+            if self._projection: 
+                self._projection = self._projection.ExportToProj4()
+                
+        return self._projection
--- a/ale/base/base.py
+++ b/ale/base/base.py
@@ -356,3 +356,30 @@ class Driver():
                 self._projection = self._projection.ExportToProj4()
                 
         return self._projection
+    
+
+    @property 
+    def geotransform(self):
+        if not hasattr(self, "_geotransform"): 
+            try: 
+              from osgeo import gdal 
+            except: 
+                self._geotransform = None
+                return self._geotransform
+
+            if isinstance(self._file, pvl.PVLModule):
+                # save it to a temp folder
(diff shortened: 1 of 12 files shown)


ale/base/base.py
```diff
@@ -323,3 +326,33 @@ def center_ephemeris_time(self):
...
         return self.__module__.split('.')[-1].split('_')[0]
+
+    @property 
+    def projection(self):
+        if not hasattr(self, "_projection"): 
+            try: 
+              from osgeo import gdal 
+            except: 
+                self._projection = None
+                return self._projection
```
Kelvinrr, line 337: Localized the import to keep the optional gdal stuff in one location without complicating things by adding some gdal module or something.

recipe/meta.yaml
```diff
@@ -31,7 +31,9 @@ requirements:
     - scipy >=1.4.0
     - spiceypy >=4.0.1
     - pyyaml
-
+  run_contrained: 
+    - gdal
```
Kelvinrr, line 35: will only install if it can do so without conflicts. important to co-install with ISIS.

tests/pytests/data/B10_013341_1010_XN_79S172W/B10_013341_1010_XN_79S172W_isis3.lbl
```diff
@@ -69,6 +69,25 @@ Object = IsisCube
     Source                    = isis
   End_Group
 
+  Group = Mapping
+    ProjectionName     = Sinusoidal
+    CenterLongitude    = 148.36859083039
+    TargetName         = MARS
```
Kelvinrr, line 75: just kinda shoved a projection into an MRO image for testing, I imagine the same to what a real image is like.

ale/base/base.py
```diff
@@ -323,3 +326,33 @@ def center_ephemeris_time(self):
...
+                    self._projection = None 
+                else: 
+                    geodata = gdal.Open(self._file)
+                    self._projection = geodata.GetSpatialRef()
+
+            # is None if not projected
+            if self._projection: 
+                self._projection = self._projection.ExportToProj4()
+                
+        return self._projection
```
acpaquette, line 358: This also needs to pull the geo transform off of the geodata. Should be able to get it using `GetGeoTransform()`

ale/formatters/usgscsm_formatter.py
```diff
@@ -51,6 +51,8 @@ def to_usgscsm(driver):
         'unit' : 'm'
     }
 
+    isd_data["projection"] = driver.projection
+
```
acpaquette, line 55: Similarly to the above, the geo transform needs to be added here

acpaquette (changes requested):
Overall looks good, we need to get the geoTransform attached to the isd to finish this off.

We also need to define how those fields are accessed in ALE. I have written the functions to do so in USGSCSM, but will paste them here:

```
std::vector<double> getGeoTransform(json isd) {
  std::vector<double> transform = {};
  try {
    transform = isd.at("geo_transform").get<std::vector<double>>();
  } catch (std::exception &e) {
    std::string originalError = e.what();
    std::string msg = "Could not parse the geo_transform. ERROR: " + originalError + isd.dump();
    throw std::runtime_error(msg);
  }
  return transform;
}

std::string getProjectionString(json isd) {
  std::string projection_string = "";
  try {
    projection_string = isd.at("proj_string");
  } catch (...) {
    throw std::runtime_error("Could not parse the projection string.");
  }
  return projection_string;
}
```

ale/base/base.py
```diff
@@ -323,3 +326,60 @@ def center_ephemeris_time(self):
...
+    
+
+    @property 
+    def geotransform(self):
+        if not hasattr(self, "_geotransform"): 
+            try: 
+              from osgeo import gdal 
+            except: 
+                self._geotransform = (0.0, 1.0, 0.0, 0.0, 0.0, 1.0)
```
Kelvinrr, line 367: Im preeeetty sure this is the identity, probably good as a default.

ale/formatters/usgscsm_formatter.py
```diff
@@ -51,6 +51,9 @@ def to_usgscsm(driver):
         'unit' : 'm'
     }
 
+    isd_data["projection"] = driver.projection
+    isd_data["geotransform"] = driver.geotransform
```
acpaquette, line 55: This needs to make it's way into the `to_isd` function in the generic formatter.

ale/base/base.py
thareUSGS: Yes that should identity for GDAL. https://github.com/OSGeo/gdal/blob/c642af0722d43bb337aaf86293129dbed576e534/alg/gdaltransformer.cpp#L1847  

But just a warning, this is not same order for all packages.

src/Util.cpp
```diff
@@ -37,6 +37,31 @@ std::string getImageId(json isd) {
...
 }
 
+
+std::vector<double> getGeoTransform(json isd) {
+  std::vector<double> transform = {};
+  try {
+    transform = isd.at("geotransform").get<std::vector<double>>();
+  } catch (std::exception &e) {
+    std::string originalError = e.what();
+    std::string msg = "Could not parse the geo_transform. ERROR: " + originalError + isd.dump();
```
acpaquette, line 47: Might want a new line character between the ISD and the original error. In general you may want to avoid printing the ISD. If I left that in it can be removed.

acpaquette (commented):
Overall looks good, one last comment to address

src/Util.cpp
```diff
@@ -37,6 +37,31 @@ std::string getImageId(json isd) {
...
+    throw std::runtime_error(msg);
+  }
+  return transform;
+}
+
+
+std::string getProjectionString(json isd) {
```
acpaquette, line 54: This also needs to be renamed to `getProjection` to match the h file

Kelvinrr: isd how the test didn't have an undefined symbol error????

acpaquette (approved):

PR #528, opened 2023-04-08
Projection Addition Fixes

acpaquette:
Fixes some of the issues with the new projection extraction code.

Also enforces that labels will only contain the `projection` and `geotranform` keys if a valid projection can be derived from the label


Kelvinrr (dismissed):

Kelvinrr:
error is some connection problem, probably can ignore it

Kelvinrr (commented):
before merging, we should probably add a changeling update

AustinSanders (approved):

PR #530, opened 2023-05-09
Hayabusa NIRS

jrcain-usgs:
Note: One test fails because of projection/geotransform, there is a separate PR up for that issue though.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,28 @@
+image: ubuntu:18.04
+
+clone_github:
+    stage: deploy 
+    before_script:
+        - apt-get update -y && apt-get install openssh-client -y
+        - apt install git -y
+        - eval $(ssh-agent -s)
+        - echo "$GIT_SSH_PRIVATE_KEY" | openssl base64 -A -d | ssh-add -
+        - mkdir -p ~/.ssh
+        - chmod 700 ~/.ssh
+        - ssh-keyscan $GIT_LAB_HOST >> ~/.ssh/known_hosts
+        - chmod 644 ~/.ssh/known_hosts
+        - git config --global user.name "${GIT_USER_NAME}"
+        - git config --global user.email "${GIT_USER_EMAIL}"
+        - ssh "git@$GIT_LAB_HOST"
+    script:
+        - echo $SOURCE_REPOSITORY
+        - git config --global http.sslVerify false # Epic....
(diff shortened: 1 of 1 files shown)


ale/drivers/hayabusa_drivers.py
```diff
@@ -41,4 +43,70 @@ def sensor_name(self):
...
+        : str
+          Name of the instrument
+        """
+        lookup_table = {'NIRS': 'HAYABUSA_NIRS'}
+        return lookup_table[super().instrument_id]
+
+    @property
+    def sensor_model_version(self):
+        """
+        The ISIS Sensor model number for HiRise in ISIS. This is likely just 1
```
thareUSGS, line 66: An old note pointing to HiRISE. Just remove ot update to Hayabusa Nirs

notebooks/write_HayabusaNirsIsisLabelNaifSpiceDriver.ipynb
```diff
@@ -0,0 +1,324 @@
...
+    "\n",
+    "import ale\n",
+    "from ale.util import generate_kernels_from_cube\n",
+    "from ale.drivers.hayabusa_drivers import HayabusaNirsIsisLabelNaifSpiceDriver\n",
+    "from ale.formatters.formatter import to_isd\n",
+    "from ale.drivers import AleJsonEncoder\n",
+    "\n",
+    "os.environ['ALESPICEROOT'] = '/Users/jrcain/remotedata/isis_data/'\n",
+    "os.environ['ISISDATA'] = '/Users/jrcain/remotedata/isis_data/'\n",
+    "ale.spice_root = '/Users/jrcain/remotedata/usgs_data/'"
```
acpaquette, line 27: For compliance when doing releases we have to avoid concrete paths. This also has to be changed in the commit that this is part of, which is frustrating. This is also the case for the KernelSlice notebook.

acpaquette (changes requested):
Overall looks great! We need to remove any concrete paths from the commit history here for release pruposes

notebooks/write_HayabusaNirsIsisLabelNaifSpiceDriver.ipynb
jrcain-usgs: Ah, I missed that for the NIRS notebook.  In the KernelSlice notebook, I put paths like "~/remotedata", so there shouldn't be absolute paths or usernames in there, is that compliant?

PR #531, opened 2023-05-10
UVVIS ALE Driver

antonhibl:
Additional ALE driver for Clementine-1 UV-VIS images.


antonhibl:
still need to write tests

PR #532, opened 2023-05-12
Hayabusa NIRS

jrcain-usgs:
Redo of previous PR without absolute paths.

One of the tests may fail because of the projection/geotransform bug, addressed in #528


acpaquette (changes requested):
Overall looks great! Just need a changelog entry and this can get merged

jrcain-usgs:
Changelog entry added.

acpaquette:
@jrcain-usgs looks like the nirs test is failing now due to the merge of the projection fix. Can the test here be updated? You should just need to remove the projection and geotransform keywords in the truth ISD

acpaquette (approved):

acpaquette (approved):

PR #538, opened 2023-05-24
Mgs time bias

acpaquette:
Address a time bias in the MGS NAC/WAC images resulting in poorly corresponding ephemeris data when compared to ISIS


tests/pytests/test_cassini_drivers.py
```diff
@@ -140,7 +140,7 @@ def test_custom_frame_chain_iak(self, from_spice):
             target_frame_id.return_value = -800
             frame_chain = self.driver.frame_chain
             assert len(frame_chain.nodes()) == 0
-            from_spice.assert_called_with(center_ephemeris_time=2.4, ephemeris_times=[2.4], nadir=False, sensor_frame=14082360, target_frame=-800, exact_ck_times=True)
+            from_spice.assert_called_with(center_ephemeris_time=2.4, ephemeris_times=[2.4], nadir=False, sensor_frame=14082360, target_frame=-800, exact_ck_times=True,  inst_time_bias=0)
```
AustinSanders, line 143: Why is this only necessary on cassini drivers?

AustinSanders:
Overall looks good!  Are we planning on reenabling mgs drivers when this is merged, or leaving them out until this is 100% squared away?

tests/pytests/test_cassini_drivers.py
acpaquette: Cassini has a special additional frame that is added to the end of the frame chain. It seems like that driver is the only thing testing the from_spice call.

I think that the NaifData class tests should have caught this as well, it's just hard to get all the checks.

AustinSanders (dismissed):

jrcain-usgs (approved):

PR #541, opened 2023-06-01
Mex fixes

acpaquette:
Fixes up MEX HRSC drivers for use in ISIS and USGSCSM


ale/drivers/mex_drivers.py
```diff
@@ -69,7 +69,7 @@
             [-8569.5561557859, 0.068566503695773, 142.857126402363]],
 }
 
-class MexHrscPds3NaifSpiceDriver(LineScanner, Pds3Label, NaifSpice, RadialDistortion, Driver):
+class MexHrscPds3NaifSpiceDriver(LineScanner, Pds3Label, NaifSpice, NoDistortion, Driver):
```
Kelvinrr, line 72: why disable distortion?

tests/pytests/data/h5270_0000_ir2/h5270_0000_ir2_pds3.lbl
```diff
@@ -115,147 +115,6 @@ RADIANCE_OFFSET              = 0.0 <W*m**-2*sr**-1>
 RADIANCE_SCALING_FACTOR      = 0.0169353 <W*m**-2*sr**-1>
 REFLECTANCE_SCALING_FACTOR   = 0.00215382
 
-/* DATA OBJECT DEFINITIONS */
-Object = IMAGE
-  INTERCHANGE_FORMAT = BINARY
-  LINES              = 400
-  LINE_PREFIX_BYTES  = 68
-  LINE_SAMPLES       = 1288
-  SAMPLE_TYPE        = MSB_INTEGER
```
Kelvinrr, line 124: why is all of this deleted?

ale/drivers/mex_drivers.py
acpaquette: I think this was before we had a `NoDistortion` class/mixin in ALE. In the driver we define the distortion coefficients as all zeros:
```
    @property
    def odtk(self):
        """
        The coefficients for the distortion model
        Returns
        -------
        : list
          Radial distortion coefficients. There is only one coefficient for LROC NAC l/r
        """
        return [0.0, 0.0, 0.0]
```

I just changed it to remove duplicate code but didn't remove the odtk property. I will do that

tests/pytests/data/h5270_0000_ir2/h5270_0000_ir2_pds3.lbl
acpaquette: For whatever reason two pds3 labels existed in the file. One with 400 lines and the other with 15088 lines which is the actual number of lines in the image.

Using the real number of lines made testing easier

Kelvinrr (approved):

PR #546, opened 2023-06-07
UVVIS driver and tests

antonhibl:


tests/pytests/data/LUD3125I.161/LUD3125I.161_isis3.lbl
```diff
@@ -0,0 +1,140 @@
+PDS_VERSION_ID               = PDS3
```
acpaquette, line 1: This needs to be the actual ISIS cube label instead of the lbl file from the PDS3 image. The easiest way to get the ISIS label is to run `catlab` on the ISIS cube then just overwrite this file

ale/drivers/uvvis_drivers.py
```diff
@@ -0,0 +1,111 @@
...
+from ale.base.type_distortion import RadialDistortion, NoDistortion
+from ale.base.type_sensor import Framer, LineScanner
+from ale.util import generate_kernels_from_cube
+from ale.base.type_sensor import Framer
+from ale.base.type_distortion import NoDistortion
+
+from ale import util
+
+
+class UvvisIsisLabelNaifSpiceDriver(Framer, IsisLabel, NaifSpice, NoDistortion, Driver):
```
acpaquette, line 20: Can this driver be changed to `ClementineUvvisIsisLabelNaifSpiceDriver`?

ale/drivers/uvvis_drivers.py
```diff
@@ -0,0 +1,111 @@
+import os
```
acpaquette, line 1: Can this file be changed to `clementine_drivers` as this particular driver is the clementine uvvis driver?

tests/pytests/test_clementine_drivers.py
```diff
@@ -0,0 +1,56 @@
+import os
+import json
+import unittest
+from unittest.mock import PropertyMock, patch
+
+import pytest
+
+import ale
+from ale.drivers.uvvis_drivers import ClementineUvvisIsisLabelNaifSpiceDriver
```
acpaquette, line 9: This import needs to be updated to `ale.drivers.clementine_drivers`

acpaquette (approved):
Overall looks good now! I am going to approve and hopefully that will get tests running. If those pass this can be merged

tests/pytests/data/LUA3107H.161/LUA3107H.161.clem_0.conf
```diff
@@ -0,0 +1,9 @@
+LEAPSECONDS_KERNEL     = /Users/ahibl/astro_efs/base/kernels/lsk/naif0012.tls
```
acpaquette, line 1: Could we remove these .conf and .cmt files? They aren't necessary to get things passing

acpaquette: Looks like there is still one conf, we also don't want to include the bck files

PR #548, opened 2023-07-05
ALE Windows CI

acpaquette:
Adds ALE windows builds

### DIFF
--- a/.github/workflows/ci_testing.yml
+++ b/.github/workflows/ci_testing.yml
@@ -90,3 +90,4 @@ jobs:
           name: ale
           path: |
             D:\a\ale\ale\build\Release\ale.dll
+            D:\a\ale\ale\build\Release\ale.lib
--- a/.github/workflows/ci_testing.yml
+++ b/.github/workflows/ci_testing.yml
@@ -9,7 +9,7 @@ on:
       - main
 
 jobs:
-  Build-and-Test-Linux:
+  Build-and-Test-LinuxOsx:
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
@@ -20,10 +20,10 @@ jobs:
       run:
         shell: bash -l {0}
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9
         with:
           submodules: true
-      - uses: conda-incubator/setup-miniconda@v2
+      - uses: conda-incubator/setup-miniconda@3b0f2504dd76ef23b6d31f291f4913fb60ab5ff3
         with:
           miniconda-version: "latest"
           activate-environment: ale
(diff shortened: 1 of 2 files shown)


.github/workflows/ci_testing.yml
```diff
@@ -57,3 +57,36 @@ jobs:
...
+  Build-and-Test-Win:
+    runs-on: windows-2019
+    strategy:
+      matrix:
+        python-version: ["3.8", "3.9", "3.10", "3.11"]
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: recursive
+      - uses: conda-incubator/setup-miniconda@v2
```
jlaura, line 70: This needs to have a hash, not a version.

.github/workflows/ci_testing.yml
```diff
@@ -57,3 +57,36 @@ jobs:
...
+          auto-update-conda: true
+          python-version: ${{ matrix.python-version }}
+      - name: Check build environment
+        run: |
+          conda list
+      - name: Build Package
+        run: |
+          mkdir -p build
+          cd build
+          cmake -DCMAKE_BUILD_TYPE=RELEASE -DCMAKE_WINDOWS_EXPORT_ALL_SYMBOLS=ON -G "Visual Studio 16 2019" -A x64 -DALE_BUILD_TESTS=OFF ..
```
jlaura, line 85: Does the build machine have this specific VS Studio?

.github/workflows/ci_testing.yml
```diff
@@ -57,3 +57,36 @@ jobs:
...
           verbose: true
         if: matrix.os == 'ubuntu-latest' && matrix.python-version == '3.9'
+
+  Build-and-Test-Win:
+    runs-on: windows-2019
+    strategy:
+      matrix:
+        python-version: ["3.8", "3.9", "3.10", "3.11"]
+    steps:
+      - uses: actions/checkout@v2
```
jlaura, line 67: This needs to have a hash, no a version

.github/workflows/ci_testing.yml
acpaquette: Is there a reason we need the hash instead of the version? The build failures were due to an update to scipy

jlaura: Yes - see the DGEC Github actions policy. Everything needs a SHA for security. On all repos.

.github/workflows/ci_testing.yml
jlaura: See: https://github.com/DOI-USGS/ale/actions/runs/5468659762/jobs/9956589863#step:5:19

.github/workflows/ci_testing.yml
acpaquette: Can I get the hashes from the USGSCSM repo?

jlaura: Yes. They are also discoverable by opening the tag on the action repo (e.g., https://github.com/actions/checkout/tree/v3.5.3) and then clicking the truncated commit hash (screen shot). Then copy the hash from the URL in the newly opened page.

This is required because the action at a given hash is trackable from a security perspective. A tag is mutable, a hash is not.
Screen Shot 2023-07-05 at 2 22 46 PM

.github/workflows/ci_testing.yml
acpaquette: I think so? This is the same version that was being used in the appveyor build and is the same version that USGSCSM uses

jlaura (approved):

jlaura (approved):

PR #557, opened 2023-08-30
Docs workflow

chkim-usgs:
Only builds docs for Python 3.11 on Ubuntu.

ALE Docs current url: https://asc-docs.s3.us-west-2.amazonaws.com/ale/index.html

Successful build: https://github.com/DOI-USGS/ale/actions/runs/6029661002/job/16359713495#step:12:1


.github/workflows/ci_testing.yml
```diff
@@ -3,10 +3,10 @@ name: Pull-Request-CI
 on:
   pull_request:
     branches:
-      - main
+      - finalize_docs_workflow #main
   push:
     branches:
-      - main
+      - finalize_docs_workflow #docs_workflow
```
acpaquette, line 9: Same as above

.github/workflows/ci_testing.yml
```diff
@@ -57,6 +57,24 @@ jobs:
...
+        if: matrix.os == 'ubuntu-latest' && matrix.python-version == '3.11'
+        uses: aws-actions/configure-aws-credentials@0e613a0980cbf65ed5b322eb7a1e075d28913a83
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-west-2
+      - name: Upload to S3
+        if: matrix.os == 'ubuntu-latest' && matrix.python-version == '3.11'
+        run: |
+          aws s3 sync --delete docs/build/html s3://asc-docs/ale
```
acpaquette, line 77: I am worried that this builds and push docs for every PR that comes through without being merged. Ideally we would only want to deploy docs once merged into main

.github/workflows/ci_testing.yml
```diff
@@ -3,10 +3,10 @@ name: Pull-Request-CI
 on:
   pull_request:
     branches:
-      - main
+      - finalize_docs_workflow #main
```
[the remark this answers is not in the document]
chkim-usgs: Yes, just saw that and pushed changes

.github/workflows/ci_testing.yml
chkim-usgs: Hm ok so I need to set some sort of condition that only builds/pushes docs if it's a PR being merged into `main`.

chkim-usgs:
I'll add OCAP compliance to this PR as well.

acpaquette (commented):
Did the files that got moved into the source directory need to be moved? I know that USGCSM used a directory under docs called source to basically house everything but it technically doesn't need to be that way. Would it be possible to revert the moves then update the cmake/configuration files to reflect the structure of the ale/docs directory?

chkim-usgs:
@acpaquette Reverted the file structure, how does it look?

acpaquette (commented):
@chkim-usgs Overall looks good! I think there is one last file to remove. The public/.doctrees file. I think it was a binary file that got included on accident. Once that is removed we can get this merged

acpaquette (approved):

PR #558, opened 2023-09-06
MSL Fixes/Landed Addition

acpaquette:
Allows MSL ISDs to be generated with zero'd positions and velocities and adjusted pointing information.

This change allows the rover images to be locally projected relative to the rover. 

Works in ISIS using skymap.


AustinSanders (approved):
This all seems good to me. I'm not familiar with the inner workings of the cahvor model, but the changes seem reasonable.

PR #564, opened 2023-09-15
MSL Nadir Pointing

acpaquette:
Adds the ability to request nadir pointing for MSL data.

This allows for orthographic projection in ISIS using cam2map.


acpaquette:
@oleg-alexandrov This PR would also be a good idea to look at on Monday

oleg-alexandrov:
I am not sure why you need to have it both ways with the signs. I did not encounter such an issue with MSL. But whichever way it works. 

I had my own tweaks for signs for MSL in my branch of ale, and I think I made some big changes to what actually gets saved in the .json file for the constant rotation. If you are getting unexpected cameras when saving from disk, you may want to take a look.

BTW, I also had a fix for ensuring the left and right nav cam sensor have the proper baseline. This may not be important for stitching, but it becomes important for stereo. 

And a question. You still need some kind of rotation matrix to place the rover in ellipsoid coordinates rather than at planet center. How do you get that rotation?

acpaquette:
@oleg-alexandrov the isd_generate script has an extra argument in this PR that only puts the rover at the center of the body when you set `-l or --local`. Otherwise the rover is on the ellipsoid

oleg-alexandrov:
Looks good. This brings us closer to accurate pose for the rover on the planet. By now the issue is not with cahvor but in the rover frame.

acpaquette:
@oleg-alexandrov Right, I was able to get around this by just computing nadir pointing which seems to work well. I would be curious to see how close the nadir rotations are from the rotations you are computing in your PR

Kelvinrr:
I think this only needs passing tests?

Kelvinrr (approved):

PR #566, opened 2023-09-25
DawnVIR Driver

amystamile-usgs:
~~Notebook is currently generating an ISD but still need to handle this line from the camera model https://github.com/DOI-USGS/ISIS3/blob/dev/isis/src/dawn/objs/DawnVirCamera/DawnVirCamera.cpp#L385C2-L386C80~~


amystamile-usgs:
Was able to get ISIS to spiceinit with ale driver. AVs and ETs look good but quats are still off.

image
image

ale/drivers/dawn_drivers.py
```diff
@@ -199,3 +208,227 @@ def detector_center_line(self):
...
+                target_time_dependent_frames, target_constant_frames = self._frame_chain.frame_trace(self.target_frame_id, self.center_ephemeris_time)
+
+                sensor_time_dependent_frames = list(zip(sensor_time_dependent_frames[:-1], sensor_time_dependent_frames[1:]))
+                constant_frames = list(zip(sensor_constant_frames[:-1], sensor_constant_frames[1:]))
+                target_time_dependent_frames = list(zip(target_time_dependent_frames[:-1], target_time_dependent_frames[1:]))
+                target_constant_frames = list(zip(target_constant_frames[:-1], target_constant_frames[1:]))
+
+                constant_frames.extend(target_constant_frames)
+                self._frame_chain.compute_time_dependent_rotiations(sensor_time_dependent_frames, sensor_times, 0)
+                self._frame_chain.compute_time_dependent_rotiations(target_time_dependent_frames, target_times, 0)
```
acpaquette, line 415: You should be able to remove this section of the frame chain code

ale/drivers/dawn_drivers.py
```diff
@@ -199,3 +208,227 @@ def detector_center_line(self):
...
+                  rot_mat, av = spice.xf2rav(xform2)
+                  
+                  quat_from_rotation = spice.m2q(rot_mat)
+                  quats[i,:3] = quat_from_rotation[1:]
+                  quats[i, 3] = quat_from_rotation[0]
+
+                  rotation = ConstantRotation(quats[i], spice.namfrm("J2000"), self.vir_zero_id)
+
+                  self._frame_chain.add_edge(rotation=rotation)
+
```
acpaquette, line 433: I think the rosetta PR also does this  part, I would check there and compare. I know that the constant rotation should be a time dependent rotation though

acpaquette (commented):
A couple suggestions for fixing things

ale/formatters/formatter.py
```diff
@@ -123,8 +125,8 @@ def to_isd(driver):
 
     # Reverse the frame order because ISIS orders frames as
     # (destination, intermediate, ..., intermediate, source)
-    instrument_pointing['time_dependent_frames'] = shortest_path(frame_chain, destination_frame, J2000)
-    time_dependent_rotation = frame_chain.compute_rotation(J2000, destination_frame)
+    instrument_pointing['time_dependent_frames'] = shortest_path(frame_chain, destination_frame, 1)
+    time_dependent_rotation = frame_chain.compute_rotation(1, destination_frame)
```
acpaquette, line 129: Could this be changed back to J2000 rather than 1?

acpaquette:
Could we also add a changelog entry?

amystamile-usgs:
@acpaquette test_vir_load failing.

acpaquette (approved):

PR #575, opened 2023-11-07
Add radtan distortion

oleg-alexandrov:
Add support for the radial + tangential distortion model. This has 5 distortion coefficients, stored in order k1, k2, p1, p2, k3. The k1, k2, k3 coefficients are for radial distortion, and p1 and p2 for tangential distortion. This was validated with the OpenCV implementation using the function cv::projectPoints(). More background in https://github.com/DOI-USGS/usgscsm/issues/465. 

## Licensing

### DIFF
--- a/include/ale/Distortion.h
+++ b/include/ale/Distortion.h
@@ -9,7 +9,8 @@ namespace ale {
     DAWNFC,
     LROLROCNAC,
     CAHVOR,
-    LUNARORBITER
+    LUNARORBITER,
+    RADTAN
   };
 }
 
--- a/src/Util.cpp
+++ b/src/Util.cpp
@@ -342,9 +342,9 @@ double getSemiMinorRadius(json isd) {
 // type. Defaults to transverse
 DistortionType getDistortionModel(json isd) {
   try {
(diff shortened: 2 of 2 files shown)


acpaquette:
@oleg-alexandrov Is the plan for this to just inject the distortion model after you create an ISD?

oleg-alexandrov:
My plan is to use this distortion model for CSM cameras I create in ASP. That because this does not correspond to any sensor that ALE supports. I only added enough logic to ensure ALE and CSM can load and process files having this distortion model.

acpaquette:
Is there some driver that will eventually go into ALE that uses this model?

acpaquette (approved):

oleg-alexandrov:
If the changes are approved, can they be pulled in? There's more discussion to be had at the usgscsm repo, but these ale changes appear small enough. (And without this brought in the usgscsm build fails.)

acpaquette:
@oleg-alexandrov It's approved does someone internal to the DOI-USGS still need to hit merge?

oleg-alexandrov:
I don't have write access to this repo (and no need to have this access). Thanks.

Issue #552, opened 2023-08-06
tkfram and zzdynrot matrix order issue


AndrewAnnex:
A SpiceyPy user recently opened an issue (https://github.com/AndrewAnnex/SpiceyPy/issues/473) to report that `tkfram` was returning a incorrectly transposed rotation matrix. I identified this issue to be due to the use of the direct c converted fortran spice functions and that the `zzdynrot` was also likely effected, while other similar fortran functions (irfrot etc) were not effected. After looking into this further, both zzdynrot and tkfram were added via a pull request to support ALE in 2019 https://github.com/AndrewAnnex/SpiceyPy/pull/301, so this bug has been lurking in the codebase for a while. 

I currently have a pull request https://github.com/AndrewAnnex/SpiceyPy/pull/474 that includes the fixes, but this project should evaluate the degree to which they are impacted by this change.

Kelvinrr:
SpiceyPy was updated to 6.0 in environment.yml and tests pass, next ALE release (1.0), probably after the ALE sprint a few weeks from now, we will update the recipe on conda-forge as well.

acpaquette:
@Kelvinrr can this be closed?

Kelvinrr:
I had this sitting in a local branch and forgot to PR it in 💀 

up now: https://github.com/conda-forge/ale-feedstock/pull/58

PR #578, opened 2023-11-14
Spiceypy version bump

Kelvinrr:
Addresses [#552](https://github.com/DOI-USGS/ale/issues/552)

I'll put in a PR into the recipe as well. 

## Licensing

- [ ] I dedicate any and all copyright interest in this software to the public domain. I make this dedication for the benefit of the public at large and to the detriment of my heirs and successors. I intend this dedication to be an overt act of relinquishment in perpetuity of all present and future rights to this software under copyright law.

### DIFF
--- a/code.json
+++ b/code.json
@@ -218,8 +218,8 @@
     ],
 
     "contact": {
-      "name": "Jason Laura",
-      "email": "jlaura@usgs.gov"
+      "name": "Kelvin Rodriguez",
+      "email": "krodriguez@usgs.gov"
     },
 
     "date": {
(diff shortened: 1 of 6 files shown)


Kelvinrr:
@acpaquette I think this would require a new ALE major release? Technically this change is breaking potentially for some people because the change in spiceypy. In which case, I can update this with a version tick for ALE.

code.json
```diff
@@ -222,6 +222,53 @@
...
+    ],
+
+    "languages": [
+      "C++",
+      "Python"
+    ],
+
+    "contact": {
+      "name": "Jason Laura",
+      "email": "jlaura@usgs.gov"
```
acpaquette, line 269: I don't think Jay should remain the point of contact here, either myself or potentially @Kelvinrr

acpaquette (changes requested):
There may also be a version defined in the CmakeLists.txt file at the root of the repo

acpaquette (approved):

PR #579, opened 2023-11-17
Replace affine6p transformation with numpy solution

chkim-usgs:
Addresses https://github.com/DOI-USGS/ale/issues/573

The `numpy` solution is from https://stackoverflow.com/a/20555267.

Both the `numpy` solution and `affine6p` solution uses the least squares method (repo link for `affine6p` [here](https://gitlab.com/yoshimoto/affine6p-py)).

This file [affine_differences.txt](https://github.com/DOI-USGS/ale/files/13397291/affine_differences.txt) shows the pytest difference output between the `affine6p` and `numpy` transformation estimations, respectively. The `numpy` output is fairly accurate to around the 10th decimal in comparison to the existing `affine6p` output in the `test_naif_keywords()` test. 

Updated the `test_lo_drivers.py` to assert the `TRANSX`, `TRANSY`, `ITRANSS`, and `ITRANSL` values are almost equal to the `affine6p` output.

## Licensing

### DIFF
--- a/ale/drivers/lo_drivers.py
+++ b/ale/drivers/lo_drivers.py
@@ -1,6 +1,5 @@
 import spiceypy as spice
 import numpy as np
-import affine6p
 from ale.base.data_naif import NaifSpice
 from ale.base.label_isis import IsisLabel
 from ale.base.type_sensor import Framer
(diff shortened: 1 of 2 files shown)


acpaquette (commented):
Looks good! I think this can get merge once we add a changelog entry

jlaura (dismissed):
Can `pip` be removed from the conda recipe as well? https://github.com/DOI-USGS/ale/blob/main/recipe/meta.yaml#L21

And CHANGELOG entry?

Approved because those are super minor. This looks good.

acpaquette (approved):

PR #589, opened 2024-01-12
Bufixes for MSL sensor intrinsics

oleg-alexandrov:
This is a fix for a bug I made with MSL cameras.

This bug explains part of the problem of why ISIS is having a hard time with MSL data.

A corresponding bug was fixed in ASP. The build 2024-01-12 has that. An earlier ASP build will now give wrong result.

The root cause of the problem was that for MSL the zero datum is not beneath the rover, but above it, so some heuristics was failing. 

Things were working with mapproject because two bugs were canceling each other out (since I developed both). No such luck with ISIS.

This was very carefully tested with many SOL00603 MAST and NAV images. The most relevant pair is:

SOL00603/NLB_451026746EDR_F0311094NCAM00271M1
MAST_0603/0603ML0025450040301380C00_DRCL

These can be used with the MSL DEM I shared with @acpaquette.

I am not sure a new changelog entry is needed, since the existing MSL work is already listed.

## Licensing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,7 @@ release.
 ### Fixed
 - Fixed LRO MiniRF drivers naif keywords focal to pixel and pixel to focal translations to be correct. [#569](https://github.com/DOI-USGS/ale/pull/569)
 - Bugfix for position and orientation for MSL cameras (driver MslMastcamPds3NaifSpiceDriver). Validated that Nav and Mast LBL files (for both left and right sensor) produce correctly positioned and oriented CSM cameras, that are self-consistent and consistent with a prior DEM for the site. [#580](https://github.com/DOI-USGS/ale/pull/580) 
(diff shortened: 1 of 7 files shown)


acpaquette:
Sweet, thank you Oleg. Two things:

Tests will likely need to be fixed, just started the pipelines

Can we get another changelog entry under `Fixes`. Something along the lines of inverted pixel_size and pixel_size related properties in CAHVOR/MSL

oleg-alexandrov:
Will do, in a bit.

oleg-alexandrov:
I put a fix to the tests. Only one test still fails:

tests/pytests/test_util.py::test_get_metakernels_no_alespiceroot

This is a strange one. It is supposed to warn if ALESPICEROOT is not set. I did not modify anything regarding that, so not sure.

Let us see if cloud tests pass.

acpaquette:
The 3.8 test is failing due to poor download speeds on linux. This looks good otherwise, I am going to merge this and deal with the failing ci in another PR

PR #590, opened 2024-01-16
Msl Landed Fix

acpaquette:
Updated what the "landed" props argument does for MSL. To get accurate sky projections in ISIS we just need to invert the focal length and set the spacecraft (rover) position to the center of the target body.

## Licensing


AustinSanders (commented):
Just needs a changelog entry and it should be good to go!

AustinSanders (approved):

PR #606, opened 2024-05-29
Adds ISD Compression

### LINKED ISSUE #604: Enhancement: Support for ISD Compression
When working with a large number of ISDs, the total data volume can start to become an issue. To combat this, some sort of compression for ISDs would be terrific. Right now, this is living in user land where one can compression/uncompress. This is to request a potential enhancement in ALE to standardize on one compression method. The idea being that down stream ISD consumers could choose to support compressed ISDs because they know what the compression is.

This [review of JSON](https://www.lucidchart.com/techblog/2019/12/06/json-compression-alternative-binary-formats-and-compression-methods/) suggests that simple [brotli compression](https://github.com/google/brotli) is both fast and efficient. It also seems that brotli is well supported.

amystamile-usgs:
closes #604 

## Licensing

### DIFF
--- a/ale/isd_generate.py
+++ b/ale/isd_generate.py
@@ -196,55 +196,35 @@ def file_to_isd(
     else:
         usgscsm_str = ale.loads(file, props=props, verbose=log_level>logging.INFO, only_isis_spice=only_isis_spice, only_naif_spice=only_naif_spice)
 
-    isd_file.write_text(usgscsm_str)
-
     if compress:
         logger.info(f"Writing: {os.path.splitext(isd_file)[0] + '.br'}")
-        compress_json(isd_file)
+        compress_json(usgscsm_str, os.path.splitext(isd_file)[0] + '.br')
     else:
         logger.info(f"Writing: {isd_file}")  
+        isd_file.write_text(usgscsm_str)
 
     return
 
-def write_json_file(isd_string, json_file):
+def compress_json(json_data, output_file):
     """
-    Writes out json file with inputted isd string.
+    Compresses inputted JSON data using brotli compression algorithm.
     
     Parameters
     ----------
-    isd_string : str
-        The ISD as a JSON formatted string
+    json_data : str
+        JSON data
 
-    json_file : str
-        Output json file path
-    """
-    with open(json_file, 'w') as fp:
-        json.dump(isd_string, fp, cls = AleJsonEncoder)
+    output_file : str
+        The output compressed file path with .br extension.
 
-def compress_json(uncompressed_json_file):
     """
-    Compresses inputted .json file.
+    binary_json = json.dumps(json_data).encode('utf-8')
+
+    if not os.path.splitext(output_file)[1] == '.br':
+        raise ValueError("Output file {} is not a valid .br file extension".format(output_file.split(".")[1]))
     
-    Parameters
-    ----------
-    uncompressed_json_file : str
-        .json file path
-
-    Returns
-    -------
-    str
-        Compressed .br file path
-    """
-    if not os.path.splitext(uncompressed_json_file)[1] == '.json':
-        raise ValueError("Inputted file {} is not a valid .json file extension".format(uncompressed_json_file.split(".")[1]))
-    with open(uncompressed_json_file, 'rb') as f:
-        data = f.read()
-    with open(uncompressed_json_file, 'wb') as f:
-        f.write(brotli.compress(data))
-
-    os.rename(uncompressed_json_file, os.path.splitext(uncompressed_json_file)[0] + '.br')
-
-    return os.path.splitext(uncompressed_json_file)[0] + '.br'
+    with open(output_file, 'wb') as f:
+        f.write(brotli.compress(binary_json))
 
 
 def decompress_json(compressed_json_file):
--- a/ale/isd_generate.py
+++ b/ale/isd_generate.py
@@ -194,10 +194,34 @@ def file_to_isd(
     return
 
 def write_json_file(isd_string, json_file):
+    """
+    Writes out json file with inputted isd string.
+    
+    Parameters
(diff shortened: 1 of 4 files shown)


ale/isd_generate.py
```diff
@@ -190,6 +193,71 @@ def file_to_isd(
...
+    ----------
+    uncompressed_isd_file : str
+        .json file path
+
+    Returns
+    -------
+    str
+        Compressed .br file path
+    """
+    if not uncompressed_isd_file.split(".")[1] == 'json':
```
Kelvinrr, line 225: might be more robust to use `os.path.splitext` instead

ale/isd_generate.py
```diff
@@ -190,6 +193,71 @@ def file_to_isd(
...
+    isd_string : str
+        The ISD as a JSON formatted string
+
+    json_file : str
+        Output json file path
+    """
+    with open(json_file, 'w') as fp:
+        json.dump(isd_string, fp, cls = AleJsonEncoder)
+
+def compress_isd(uncompressed_isd_file):
```
Kelvinrr, line 211: maybe a semantics, but the function seems to compress any json file, not just ISDs. We could add an extra step that verifies that the file is at least json with a json.load or similar, we could probably call this `compress_file` or `compress_json` if we keep the extension check.

ale/isd_generate.py
```diff
@@ -190,6 +193,71 @@ def file_to_isd(
...
+        data = f.read()
+    with open(uncompressed_isd_file, 'wb') as f:
+        f.write(brotli.compress(data))
+
+    os.rename(uncompressed_isd_file, os.path.splitext(uncompressed_isd_file)[0] + '.br')
+
+    return os.path.splitext(uncompressed_isd_file)[0] + '.br'
+
+
+def decompress_isd(compressed_isd_file):
```
Kelvinrr, line 237: same here

jlaura:
LGTM! General question - does one have to read/write the intermediary uncompressed file or can one just work with the compressed ISD. It looks like the workflow right now would be write uncompressed, compress. Then read compressed, uncompress and write to disk, do something with the data. Ideally, could one just instantiate a sensor model using the compressed ISD without having to use all the disk space to uncompress? Maybe that is a knoten issue?

Kelvinrr:
@jlaura It kind of does seem like a knoten/SET issue, but if we are going to support compressed ISDs we might as well provide a basic API so people dont have to implement their own. Maybe a follow up task could be to add a `CompressedJson` object that subclasses dict and read/writes in the compressed format. 

Something else to consider: add a flag to `generate_isd` to export compressed ISDs?

ale/isd_generate.py
```diff
@@ -57,6 +60,13 @@ def main():
         action="store_true",
         help="Display information as program runs."
     )
+    parser.add_argument(
+        "-c", "--compress",
+        action="store_true",
+        help="Output a compressed isd json file with .br file extension. "
+             "Ale uses the brotli compression algorithm. "
+             "To decompress an isd file run: python -c \"import ale.isd_generate as isdg; isdg.decompress_json('/path/to/isd.br')\""
+    )
```
jlaura, line 69: Cool! So using this, one could write directly to compressed. Then a SET could read the compressed ISD straight to memory and use it?

ale/isd_generate.py
```diff
@@ -185,11 +196,81 @@ def file_to_isd(
...
+    if not os.path.splitext(uncompressed_json_file)[1] == '.json':
+        raise ValueError("Inputted file {} is not a valid .json file extension".format(uncompressed_json_file.split(".")[1]))
+    with open(uncompressed_json_file, 'rb') as f:
+        data = f.read()
+    with open(uncompressed_json_file, 'wb') as f:
+        f.write(brotli.compress(data))
+
+    os.rename(uncompressed_json_file, os.path.splitext(uncompressed_json_file)[0] + '.br')
+
+    return os.path.splitext(uncompressed_json_file)[0] + '.br'
```
jlaura, line 247: Does this have to write and then re-write? No way from brotli to compress into memory to skip the intermediary file? Perhaps using something like [StringIO](https://docs.python.org/3/library/io.html#io.StringIO) or cStringIO (both standard library afaik.

ale/isd_generate.py
```diff
@@ -185,11 +196,81 @@ def file_to_isd(
...
+    -------
+    str
+        Decompressed .json file path
+    """
+    if not os.path.splitext(compressed_json_file)[1] == '.br':
+        raise ValueError("Inputted file {} is not a valid .br file extension".format(compressed_json_file))
+    with open(compressed_json_file, 'rb') as f:
+        data = f.read()
+    with open(compressed_json_file, 'wb') as f:
+        f.write(brotli.decompress(data))
```
jlaura, line 269: I am confused by the decompression here. Is this read the compressed and then write the decompressed to disk? Would it make more sense to simply pipe the decompressed to stdout? That would be more bash-like.

jlaura (approved):
Clarifying questions for me. If they are pedantic, ignore them and feel free to merge!

ale/isd_generate.py
amystamile-usgs: This is for situations where a user wants to revert to a readable JSON file saved locally. We could add a function to read the JSON from binary format, if you think that would be appropriate.

ale/isd_generate.py
amystamile-usgs: I updated this function to accept json data directly instead of reading from a file, eliminating the need to write the json data to a file.

ale/isd_generate.py
jlaura: Cool! I didn't mean for this comment to spill into more work over there! Much appreciated.

jlaura (approved):

Issue #603, opened 2024-05-20
Quaternion sign flip for LROC NAC


thareUSGS:
Oleg noticed an issue with some recent LROC NAC testing at the pole. The file M124343339LE.lev1.8bit.json seems to have a problem where the quaternions (used for orientation) have inconsistent signs. Example (line ~2193):

```
   [
        0.016525284863560585,
        -0.0734723194332759,
        0.9798210267646943,
        0.18514720831558124
      ],
    [
        0.0024593731395100016,
        0.06930709261721921,
        -0.9799803592542052,
        -0.18662522303458276
      ],
```

Oleg states, each quaternion alone is fine, as it defines the same rotation matrix with any sign. The problem is that the CSM linescan model interpolates between them, and that results in orientations being junk when the quaternions change sign.

He also prepared an image example which is the result after a temporary fix in ASP. Both of these are after mapprojection with ASP. The raw image does not have this artifact, it comes from the camera.

[M124343339LE.lev1.8bit.json](https://github.com/DOI-USGS/ale/files/15379649/M124343339LE.lev1.8bit.json)
<img width="484" alt="csm_mapproject_issue" src="https://github.com/DOI-USGS/ale/assets/3095138/e827d199-9f78-447f-973b-7c9f9cd0ccf8">

thareUSGS:
I can share the cube also (just let me know). Too big to post.

acpaquette:
@thareUSGS If you can post the cube label, that should be enough to get an ISD out of ALE. Oleg mentioning this is at the polls makes more sense

oleg-alexandrov:
There is no label, this has to come directly from the .img. That one is here: https://wms.lroc.asu.edu/lroc/view_lroc/LRO-L-LROC-2-EDR-V1.0/M124343339LE

If a sanity check with mapprojection is needed with the logic that USGSCSM now provides, the problematic area in the clip above is in the middle of the image (where it is brightest) and the artifact shows up along the entire image width.

AustinSanders:
from [naif docs](https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/cspice/qxq_c.html) 
``` 
m2q_c always returns a quaternion with scalar part greater than or equal to zero.
```
I'm pretty confident that this is the issue.  The sign of the scalar portion "should" flip at some point in the ISD, but instead it hits 0 and flips the rest of the signs in order to stay positive, e.g.:
 
```
[
[...]

PR #612, opened 2024-08-21
Update isd_generate.py to support optional function to ensure that instrument pointing quaternions do not change sign

thareUSGS:
Implement an optional function to ensure that instrument pointing quaternions do not change sign.

This is really more to continue the discussion under issue #603 . I understand that NAIF is supposed to return the quaternions with flipped signs, but this seems to be breaking our applications. ASP has been [updated](https://github.com/NeoGeographyToolkit/StereoPipeline/blob/222cb5fd64a6ea2fb723da964a131925e6ce7501/src/asp/Camera/CsmUtils.cc#L34) to account for this issue. This example function for isd_generate.py, following ASP's logic, could also be used to "clean" the ISD for other applications like SOCET GXP.

## Licensing

### DIFF
--- a/ale/isd_generate.py
+++ b/ale/isd_generate.py
@@ -280,7 +280,6 @@ def fix_quaternion_signs(usgscsm_str):
     usgscsm_json = json.loads(usgscsm_str)
     quats = usgscsm_json["instrument_pointing"]["quaternions"]
     
-    logger.info(f"fixing any quaternions that have a sign change")
     # First find the largest magnitude quaternion coefficient and its coordinate
     num_quats = len(quats)
     max_q = 0.0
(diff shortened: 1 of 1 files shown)


ale/isd_generate.py
```diff
@@ -89,6 +89,12 @@ def main():
         help="Generate nadir spice pointing, an isd that has pointing directly towards "
              "the center of the target body."
     )
+    parser.add_argument(
+        "-f", "--fix_quaternion",
+        action="store_true",
+        help="There are rare cases when a stream of quaternion flip their signs. This option will "
+             "fix those to be consistent. This takes after a NASA ASP function to do the same. "
+    )
```
Kelvinrr, line 97: Are there cases where you wouldn't want this on?

thareUSGS: Good question. I think @oleg-alexandrov would state this should always be run to fix for all images (that assumes I am doing this correctly). But perhaps. we have a flag that turns this function off instead of on.

oleg-alexandrov:
This must run always. This is fix for quartertions having two solutions for
any one rotation. In rare cases one may want the option of not doing this.

oleg-alexandrov:
To add, this fix must be deep in ALE, in the linescan driver. That some
quarternions flip sign is a limitation of the math. No tools will break if
the sign is regularized.

Nothing is gained by allowing a flip, as any one quarterion is equally good
with any sign.

ale/isd_generate.py
jlaura: Weighing in here with a +1 to @oleg-alexandrov's comment below. Always on.

Kelvinrr:
We need to create a new PR that moves this code into the ALE mixing

PR #617, opened 2024-09-17
Revert "Initial ApolloPanIsisLabelIsisSpiceDriver"

AustinSanders:
Reverts DOI-USGS/ale#616

Tests are failing as a result of #616 .  Because there's so little info in labels after apollopaninit, the driver is not currently restrictive enough to fail for other instruments.


Kelvinrr:
I think we should just get #618 working

PR #659, opened 2025-06-24
Update Windows CI Image

chkim-usgs:
[Windows 2019 image is being deprecated](https://github.com/actions/runner-images/issues/12045) very soon so we need to use a newer image.

## Licensing

### DIFF
--- a/.github/workflows/ci_testing.yml
+++ b/.github/workflows/ci_testing.yml
@@ -92,7 +92,7 @@ jobs:
           cmake --build . --target Docs --config Debug
 
   Build-and-Test-Win:
-    runs-on: windows-2019
+    runs-on: windows-2020
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci_testing.yml
```diff
@@ -92,7 +92,7 @@ jobs:
           cmake --build . --target Docs --config Debug
 
   Build-and-Test-Win:
-    runs-on: windows-2019
+    runs-on: windows-2020
```
acpaquette, line 95: @chkim-usgs Reading the blog post about it, we should switch to 2022 or 2025. I would suggest 2025 to get more time before we have to do this again. You might also have to change the visual studio version but just try setting the run-on to windows-2025

acpaquette (commented):
On line 120 I think we need to update the visual studio version to `Visual Studio 17 2022`

PR #662, opened 2025-07-11
Web Clean up

acpaquette:
Cleans up some of the web integration in ALE. These changes depend on this [PR](https://github.com/DOI-USGS/SpiceQL/pull/86) in SpiceQL

## Licensing

### DIFF
--- a/ale/transformation.py
+++ b/ale/transformation.py
@@ -281,7 +281,7 @@ class FrameChain(nx.DiGraph):
         quats_and_avs_per_frame = []
 
         logger.debug(f"Generate rotation times: {times}")
-        with ThreadPoolExecutor(max_workers=1) as executor:
+        with ThreadPoolExecutor(max_workers=3) as executor:
             futures = []
 
             for s, d in frames:
--- a/ale/transformation.py
+++ b/ale/transformation.py
@@ -128,8 +128,6 @@ class FrameChain(nx.DiGraph):
 
         constant_frames.extend(target_constant_frames)
 
-        print("SENSOR TIME DEP FRAMES:", sensor_time_dependent_frames)
-
         # Do this call so we know if we can get exact ck times for our data
         if exact_ck_times and len(ephemeris_times) > 1 and not nadir:
             try:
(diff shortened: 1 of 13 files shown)


ale/transformation.py
```diff
@@ -279,21 +278,25 @@ def generate_rotations(self, frames, times, time_bias, rotation_type, sensor_fra
...
-                        futures.append(executor.submit(get_ephem_data, times, "getTargetOrientations", 150, self.use_web, function_args))
-                    quats_and_avs_per_frame = [future.result() for future in futures]
+        logger.debug(f"Generate rotation times: {times}")
+        with ThreadPoolExecutor(max_workers=3) as executor:
+            futures = []
+
+            for s, d in frames:
+                if exact_ck_times and len(times) > 1:
+                    function_args = {"startEt": start_et+time_bias, "stopEt": stop_et+time_bias, "toFrame": d, "refFrame": s, "mission": mission, "searchKernels": self.search_kernels, "ckQualities": ["smithed", "reconstructed"], "fullKernelPath": False}
+                    futures.append(executor.submit(spiceql_call, "getExactTargetOrientations", function_args, self.use_web))
```
Kelvinrr, line 288: this code cant really run on none-web spice. Some kernel sets will need more than one CK, causing this to fail since kernels are furnished in the __enter__ call. Considering that GetTartgetOrientations only causes a speed up in the web condition, I think it's fine limiting to web.

ale/transformation.py
```diff
@@ -128,34 +128,33 @@ def from_spice(cls, sensor_frame, target_frame, center_ephemeris_time,
...
+                                                             "targetFrame": sensor_frame,
+                                                             "mission": mission,
+                                                             "searchKernels": frame_chain.search_kernels},
+                                                             use_web=frame_chain.use_web)
+
+                if len(times) == 0:
+                    exact_ck_times = False
+
             except Exception as e:
+                exact_ck_times = False
```
Kelvinrr, line 145: Similar to my other comment, Im not sure we want to spend time with the extra call if web==True. 

There might be a concern that if exact_times==False and web==True, and the sensor is a linescanner, things will take forever.

acpaquette: Sure, the extra call here is to handle if there are multiple cks that are loaded for the image. If the function fails, then we can't get exact ck times so we set `exact_ck_times` to false.

This then goes to the rotation generation that will run `getTargetOrientations` rather than `getExactTargetOrientations`. So you're right it may take a while for large line scanners. I think we need to address other ways of reducing those times rather than adding `if else` blocks that all over ALE to handle specific cases.

PR #663, opened 2025-07-15
Add warning when web=true

Kelvinrr:
We should commnicate when web=true that it has not been thoroughly tested for all instruments if it is released in a production build. 

It also shows the link so that people can create issues. 

## Licensing

- [ ] I dedicate any and all copyright interest in this software to the public domain. I make this dedication for the benefit of the public at large and to the detriment of my heirs and successors. I intend this dedication to be an overt act of relinquishment in perpetuity of all present and future rights to this software under copyright law.

### DIFF
--- a/ale/base/data_naif.py
+++ b/ale/base/data_naif.py
@@ -138,7 +138,7 @@ class NaifSpice():
                     web_prop = False
                 self._use_web = web_prop
             if self._use_web:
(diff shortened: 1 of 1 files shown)


acpaquette (approved):
Looks good, not sure if this needs a changelog entry since its somewhat user facing but it falls under the "Integrated SpiceQL" entry

PR #667, opened 2025-07-28
Online fixes

Kelvinrr:
Fixes to using some spiceql calls over web. 

## Licensing

### DIFF
--- a/ale/base/data_naif.py
+++ b/ale/base/data_naif.py
@@ -447,14 +447,12 @@ class NaifSpice():
             positions = []
             velocities = []
 
-            sun_lt_states = self.spiceql_call("getTargetStates",{"ets": times,
-                                                                 "target": "SUN",
-                                                                 "observer": self.target_name,
-                                                                 "frame": self.reference_frame,
-                                                                 "abcorr": "LT+S",
-                                                                 "limitCk": -1, 
-                                                                 "limitSpk" : 1,
-                                                                 "mission": self.spiceql_mission})
+            sun_lt_states = self.spiceql_call("getTargetStates", {"ets": times,
+                                                                  "target": "SUN",
+                                                                  "observer": self.target_name,
+                                                                  "frame": self.reference_frame,
+                                                                  "abcorr": "LT+S",
+                                                                  "mission": self.spiceql_mission})
             for sun_state in sun_lt_states:
                 sun_state = np.array(sun_state)
                 positions.append(sun_state[:3])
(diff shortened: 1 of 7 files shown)


ale/base/data_naif.py
```diff
@@ -523,8 +525,11 @@ def sensor_position(self):
                               "frame": "J2000",
                               "abcorr": self.light_time_correction,
                               "mission": self.spiceql_mission,
-                              "searchKernels": self.search_kernels}
-                    obs_tars = self.spiceql_call("getTargetStates", web=self.use_web, function_args=kwargs)
+                              "searchKernels": self.search_kernels,
+                              "limitCk": -1, 
+                              "limitSpk" : 1,
```
Kelvinrr, line 530: These are due to them not being interpreted correctly for some reason. Might be a bug with how SpiceQL web is handling default params?

ale/transformation.py
```diff
@@ -131,7 +131,7 @@ def from_spice(cls, sensor_frame, target_frame, center_ephemeris_time,
         # Do this call so we know if we can get exact ck times for our data
         if exact_ck_times and len(ephemeris_times) > 1 and not nadir:
             try:
-                times = spiceql_call("extractExactCkTimes", {"observStart": ephemeris_times[0] + inst_time_bias, 
+                ephemeris_times = spiceql_call("extractExactCkTimes", {"observStart": ephemeris_times[0] + inst_time_bias,
```
Kelvinrr, line 134: before times was not being used.

ale/transformation.py
```diff
@@ -279,41 +280,51 @@ def generate_rotations(self, frames, times, time_bias, rotation_type, sensor_fra
...
             quats = np.zeros((len(times), 4))
             avs = []
-            _quats = np.array(quats_and_avs)[:, 0:4]
+            _quats = np.asarray(quats_and_avs)[:, 0:4]
             logger.debug(f"Quats: {_quats}")
             for j, quat in enumerate(_quats):
-                quats[j,:3] = quat[1:]
-                quats[j, 3] = quat[0]
+                quats[j, 0] = quat[3]
+                quats[j, 1:4] = quat[0:3]
```
acpaquette, line 312: I don't know if this is right. Does this change how the elements in each quaternion are arranged? If the orientation calls aren't consistent with how the quaternions are returned that is an issue we need to fix

Kelvinrr: yeah, I was wondering this too, it seems to be changing the order of the quaternions. I think this is more a mismatch of how SPICE returns quats vs what ALE/CSM wants.

PR #693, opened 2026-03-31
get_naif_keywords, throw exception if not found

### LINKED ISSUE #671: Make naif_keywords explictly fail when key does not exist
It seems that when a NAIF keyword isn't in the naif_keyword dictionary, ALE stores a null instead of throwing. This probably has to do with how the props are being evaluated. 

```json
  "focal_length_model": {
    "focal_length": null
  },
  "detector_center": {
    "line": null,
    "sample": null
  },
  "focal2pixel_lines": null,
  "focal2pixel_samples": null
```

This causes CSM creation to fail and for users to dig through the logs to uncover the issue. It would be better to fail at ISD creation.

jrcain-usgs:
Fixes #671 

## Licensing


ale/base/data_isis.py
```diff
@@ -147,6 +147,11 @@ def rotate_state(table, rotation):
         rotated_vel = None
     return rotated_pos, rotated_vel, ephemeris_times
 
+def get_naif_keyword(self, ale_name, naif_key):
+    val = self.naif_keywords.get(naif_key, None)
+    if val is None:
+        raise LookupError('Could not find a value for {} using the NAIF Keyword {}'.format(ale_name, naif_key))
```
Kelvinrr: lets add what key are available for debugging. Also the standard error is KeyError

Kelvinrr:
appended on https://github.com/DOI-USGS/ale/pull/715

PR #697, opened 2026-04-10
Gdal tables

### LINKED ISSUE #670: Add support for ISIS Gtiffs/GDAL labels
It would be useful for people to run isd_generate on Gtiffs or other GDAL file types with ISIS labels. 

I think the simplist way to do this is to update the `isis_label` class to open using GDAL, similar to how ISIS does it in its code.

jrcain-usgs:
Fixes #670 

## Licensing


ale/base/data_isis.py
```diff
@@ -272,18 +364,26 @@ def sun_position_table(self):
...
+        #             binary_data = read_table_data(table, self._file)
+        #             self._sun_position_table = parse_table(table, binary_data)
+        #             return self._sun_position_table
+        #     # GDAL
+        #     t_name='SunPosition'
+        #     if isinstance(self.label[f"Table_{t_name}"], dict): 
+        #         self._sun_position_table = parse_table_json(self.label[f"Table_{t_name}"])
+        #         return self._sun_position_table
+        #     raise ValueError(f'Could not find SunPosition table on file {self._file}')
+        # return self._sun_position_table
```
acpaquette: Could these commented out sections of the properties be removed? I know that they are here so others can see what they used to be

ale/base/data_isis.py
```diff
@@ -16,6 +18,29 @@
 
 from ale.base.label_isis import IsisLabel
 
+def read_table(self, priv_name, table_name): 
+
+    if not hasattr(self, priv_name):
```
acpaquette: I'm torn about retaining this here. General functions that operate on a `self` existing outside of a class aren't ideal as they can create weird behavior patterns and inconsistent state.

While I think it prevents code repetition, I think we should be setting the table in the properties rather than handling that in this function

ale/base/data_isis.py
```diff
@@ -200,18 +284,7 @@ def inst_pointing_table(self):
         : dict
           Instrument pointing table
         """
-        if not hasattr(self, "_inst_pointing_table"):
```
acpaquette, line 203: W.r.t. my other comment, this should stay here, and we should set it from the output of `read_table`. It should look something like:
```Python
if not hasattr(self, "_inst_pointing_table"):
    self._inst_pointing_table = read_table("_inst_pointing_table", "InstrumentPointing")
return self._inst_pointing_table
```

ale/base/data_isis.py
```diff
@@ -224,18 +297,7 @@ def body_orientation_table(self):
         : dict
           Body orientation table
         """
-        if not hasattr(self, "_body_orientation_table"):
```
acpaquette, line 227: Same as above but with `_body_orientation_table`

ale/base/data_isis.py
```diff
@@ -248,18 +310,7 @@ def inst_position_table(self):
         : dict
           Instrument position table
         """
-        if not hasattr(self, "_inst_position_table"):
```
acpaquette, line 251: Same as above but with `_inst_position_table`

ale/base/data_isis.py
```diff
@@ -272,18 +323,7 @@ def sun_position_table(self):
         : dict
           Sun position table
         """
-        if not hasattr(self, "_sun_position_table"):
```
acpaquette, line 275: Same as above but with `_sun_position_table`

acpaquette (commented):
A couple of things to address

ale/base/data_isis.py
```diff
@@ -317,11 +369,22 @@ def _sclock_hex_string(self):
...
+            if isinstance(key, tuple) and re.match(regex, key[0]):
+                    # If the hex string is only numbers and contains leading 0s,
+                    # the PVL library strips them off (ie. 0000000000002040 becomes
+                    # 2040). Pad to 16 in case this happens.
+                    return str(key[1]).zfill(16)
+
+            # In gtiffs, looks like:
+            # "CLOCK_ET_-236_2":{
+            #     "0072174528:989000_COMPUTED":"4a1edaaeddcbbc41"
+            # }
```
acpaquette, line 383: I think this is a bug. I created an [issue](https://github.com/OSGeo/gdal/issues/14443) on GDAL but it might not be GDAL and may be an issue with JSON.

ale/base/data_isis.py
```diff
@@ -16,6 +18,24 @@
...
+def read_table(label, file, table_name): 
+
+    tables = []
+    if "Table" in label:
+        tables = label.getall('Table')
+
+    for table in tables:
+        if table['Name'] == table_name:
+            binary_data = read_table_data(table, file)
+            return parse_table(table, binary_data)
```
acpaquette: Can this encompassed by a check for PVL? I think we can add something like:
```python
isinstance(label, pvl.PVLModule)
```

acpaquette (commented):
Nearly there, just one more change to the`read_table` function

ale/base/data_isis.py
jrcain-usgs: Shall I leave this in there until the bug is fixed?

Kelvinrr (approved):

PR #698, opened 2026-04-15
[2.0.0] Remove isis_formatter and usgscsm_formatter

### LINKED ISSUE #653: Remove old unused formatter logic
ALE has three files:

 - ale/formatters/usgscsm_formatter.py
 - ale/formatters/isis_formatter.py
 - ale/formatters/formatter.py

These are mutated copies of the same thing, and in practice only the last one, formatter.py is used when creating ISD. 

The former two need to go, as they make it hard to understand how the logic works and where one should change or what code to read.

jrcain-usgs:
Fixes #653 

## Licensing


ale/drivers/__init__.py
```diff
@@ -84,10 +78,10 @@ def load(label, props={}, formatter='ale', verbose=False, only_isis_spice=False,
...
             property to specify an explicit set of kernels and load order.
 
-    formatter : {'ale', 'isis', 'usgscsm'}
-                Output format for the ISD. As of 0.8.0, it is recommended that
-                the `ale` formatter is used. The `isis` and `usgscsm` formatters
-                are retrained for backwards compatibility.
+    formatter : {'ale'} ('isis', and 'usgscsm' are deprecated)
+                Output format for the ISD. As of 1.2.0, the
+                `ale` formatter is always used. The parameter is retained 
+                for backwards compatibility.
```
acpaquette, line 84: Everything looks good but a breaking change like this would require a version tick to ALE 2.0.0. In ISIS we usually hold any breaking change for the yearly release, I don't know what we should do for ALE. With the spiceql changes and the gdal additions we could likely do a 2.0.0. @Kelvinrr thoughts?

Kelvinrr: we should probably hold off on this for this reason. At least we should see if there are more API breaking changes we can pile on.

oleg-alexandrov: The code removed is ancient and not used by anybody. Likely ALE 2.0.0 bump is needed simply because of adding the ShadowCam lens distortion which breaks the enums. 

I'd say the risk of this is very low. Also, likely next ISIS need to have a hard pin to this as well, and with usgscsm and spiceql too, all because of ShadowCam. There's no good reason for the user to want independent versions  for these.

oleg-alexandrov:
This should be merged, presumably because so many things broke as is in ALE, with inclusion of shadowcam lens distortion model, that this minor breaking change that removes dead code is small by comparison.

oleg-alexandrov:
This PR must be merged and the dead code has to go. This situation has been around for 3-4 years by now, and any time one studies the ALE code one is misled by three ways of doing things, out of which only one has ever worked since that time. By now it is inevitable there will be a USGSCSM 2.0 or some major bump in light of the newly added CaSSIS and KPLO ShadowCam lens distortion models. Having this drag on is just burden on developers with no upside.

oleg-alexandrov:
For the record, the current approach of making an ISD has been the default since 2020-03-31, shipped in 0.8.0 on
 2020-04-12. Nobody uses the other approaches even if they make use of ALE, unless they explcitily ask for the old format and not contented with the default, which I don't see any reason for.

oleg-alexandrov:
Closed in favor of https://github.com/DOI-USGS/ale/pull/737.

PR #700, opened 2026-04-16
Method to return driver from label

### LINKED ISSUE #633: function to return a driver for a label
Add a simple function similar to loads/load, that instead of returning the ISD, returns the driver that can be used to create an ISD.

jrcain-usgs:
Fixes #633 

With this PR, you can use ale.load to return a driver:
```py
# Using Cube 1, find the driver
my_driver = ale.drivers.load(loc1, return_driver=True)

# Use the driver on Cube 2
with my_driver(loc2) as driver:
    my_isd = to_isd(driver)

# Print ISD of Cube 2 for inspection.
my_isd
```

## Licensing


ale/drivers/__init__.py
```diff
@@ -54,7 +54,8 @@ def default(self, obj):
         return json.JSONEncoder.default(self, obj)
 
 
-def load(label, props={}, formatter='ale', verbose=False, only_isis_spice=False, only_naif_spice=False):
+def load(label, props={}, formatter='ale', verbose=False, only_isis_spice=False, only_naif_spice=False, return_driver=False):
```
Kelvinrr, line 57: idk how I feel about this, having the return type change based on a flag seems prone to error. Seems it would be better to isolate the code that finds the driver into a new function, then call that within load. So people who want a driver can explicitly call that function instead.

jrcain-usgs: I separated out a few functions for easier distinction (`load()` and `get_driver_from_label()`)- finding a driver that works and generating the ISD seem pretty intertwined though, so most of the core code is still in the same core function (`load_from_label()`).

ale/drivers/__init__.py
```diff
@@ -55,6 +55,76 @@ def default(self, obj):
...
+    Returns
+    -------
+    dict
+         The ISD as a dictionary
+
+    See Also
+    --------
+    :func:`~ale.drivers.load_from_label`
+    """
+    return load_from_label(label, props, formatter, verbose, only_isis_spice, only_naif_spice)['isd']
```
Kelvinrr: maybe the make this a tuple?

Kelvinrr (approved):

PR #703, opened 2026-05-01
Get Kernels from MKs (with path correction to SPICEROOT)

### LINKED ISSUE #529: Metakernels lookup for MSL
There is a good chance I filed something like this before, but I cannot look it up in this repo or the ALE repo. Apologies for repeating it, if that's the case.

Trying to load MSL metakernels runs into the error:

The first file '/usgs/cpkgs/isis3/data/msl/kernels/lsk/naif0012.tls' specified by KERNELS_TO_LOAD in the file /home/oalexan1/projects/isis3data/msl/kernels/mk/msl_v01.tm could not be located.

That is because some files have a hard-coded path pointing to some filesystem local to USGS which does not exist on user's system. Manually editing the offending .tm file to make the location be on my system solves this, but ideally this should work out of the box.

jrcain-usgs:
When Metakernels were auto-loaded from SPICEROOT, they would often still have their default directory, `/usgs/cpkgs/isis3/data`, which does not exist on most user's systems, and must be manually changed by the user.

This PR checks auto-loaded metakernels for correct kernel paths (in both the mk's default paths, and SPICEROOT), and returns a list of kernels paths that exist, or fails with an error explaining which kernel is missing from where.

Fixes #529 

## Licensing


Kelvinrr:
Needs changelog and tests for new function.

oleg-alexandrov:
I added the semi-related https://github.com/DOI-USGS/ale/pull/705. I had some fix for this particular issue as well but then took it out not to have a conflict. That one addresses something else.

All these show up with Chandrayaan-2. This particular thing is very long-standing.

ale/kernel_access.py
```diff
@@ -20,6 +23,149 @@
...
+        check_spiceroot = False
+        logger.debug("new_root is not a string.  Likely, ALESPICEROOT is not set.\nCannot check ALESPICEROOT for corrected metakernel paths.")
+
+    # Check if mk is a file, show its contents
+    extension = os.path.splitext(metakernel)[1]
+    if extension.lower() != '.tm':
+        raise ValueError(f'File {metakernel} is not a metakernel (does not have the .tm file extension).')
+
+    with open(metakernel, 'r') as old_mk_file:
+        mklines = old_mk_file.readlines()
```
acpaquette: I think rather then reading all the lines then reparsing them with sections and things, I think we can just parse as we go. Here is something that I think would work but needs to be extended:
```python
with open(file, 'r') as fp:
    path_values = []
    line_data = fp.readline()
    while(not re.search(r'PATH_VALUES\s*=', line_data)):
        line_data = fp.readline()
    while(not re.search(r'PATH_SYMBOLS\s*=', line_data)):
        path_values.extend(re.findall("'(.*?)'", line_data))
        line_data = fp.readline()
```

ale/kernel_access.py
```diff
@@ -20,6 +23,149 @@
...
+    logger.debug(f"Listed Kernels in metakernel: {listed_kernels}") 
+        
+    if len(listed_kernels) == 0:
+        raise ValueError(f"No kernels were found listed in this metakernel: {metakernel}")
+
+    # Check default mk paths for kernels
+    for kernel in listed_kernels:
+        default_kernel = kernel
+        for symbol, path in default_paths.items():
+            default_kernel = re.sub(r'\$'+symbol, path , default_kernel)
```
acpaquette: Rather than do this expansion manually, you'll want to load the path symbols into `os.environ`. Something like:
```
for symbol, path in default_paths.items():
    os.environ[symbol] = path
```

Then you can use `os.path.expandvars` for all of the kernels and check if they exist

ale/kernel_access.py
```diff
@@ -20,6 +23,149 @@
...
+            spiceroot_kernel = kernel
+            for symbol, path in spiceroot_paths.items():
+                spiceroot_kernel = re.sub(r'\$'+symbol, path , spiceroot_kernel)
+            if os.path.isfile(spiceroot_kernel):
+                spiceroot_kernels.append(spiceroot_kernel)
+            else:
+                missing_spiceroot_kernels = True
+                raise FileNotFoundError(f"""One or more kernel was missing from the default path in the metakernel,
+                                        so ALE looked under the ALESPICEROOT path for kernels, 
+                                        but could not find this kernel there: {spiceroot_kernel}""")
```
acpaquette, line 149: Similar idea here from previous comment

acpaquette (commented):
Initial review

ale/kernel_access.py
```diff
@@ -20,6 +23,152 @@
...
+        raise ValueError(msg)
+    else:
+        for index, key in enumerate(path_symbols):
+            default_paths[key] = path_values[index]
+            if check_spiceroot:
+                spiceroot_paths[key] = re.sub(old_root, new_root, path_values[index])
+
+    # Set env vars (mk default)
+    for symbol, path in default_paths.items():
+        os.environ[symbol] = path
```
Kelvinrr: not a good idea to modify the users env like this

ale/kernel_access.py
```diff
@@ -20,6 +23,152 @@
...
+            logger.warning(f"Could not find kernel in paths from metakernal: {default_kernel}.  Looking under ALESPICEROOT if set...")
+            missing_default_kernels = True
+            break
+
+    # If kernels missing from default, Check spice_root paths for kernels
+    if missing_default_kernels and check_spiceroot:
+
+        # Set env vars (spiceroot)
+        for symbol, path in spiceroot_paths.items():
+            os.environ[symbol] = path
```
Kelvinrr: same here

ale/kernel_access.py
```diff
@@ -20,6 +23,152 @@
...
+
+    # Check default mk paths for kernels
+    for kernel in listed_kernels:
+        default_kernel = os.path.expandvars(kernel)
+        if os.path.isfile(default_kernel):
+            default_kernels.append(default_kernel)
+        else:
+            logger.warning(f"Could not find kernel in paths from metakernal: {default_kernel}.  Looking under ALESPICEROOT if set...")
+            missing_default_kernels = True
+            break
```
Kelvinrr, line 133: is this break supposed to be here?

Kelvinrr (changes requested):

ale/kernel_access.py
jrcain-usgs: @acpaquette requested that I use the os.environ and os.path.expandvars - this was added in response to his review. I can go back my original version I guess?

acpaquette: @jrcain-usgs Lets use pythons string replace instead. A bit less complicated than regex

oleg-alexandrov: Adam, thanks for commenting. Sorry for stepping on people's foot here. I hope this is merged in as it has been a while. Can ignore my other pull request.

ale/kernel_access.py
jrcain-usgs: yes, if a kernel is missing, it should stop looking in the same place for the other kernels.

ale/kernel_access.py
jrcain-usgs: Reverted back to simple string replacement here.

ale/kernel_access.py
jrcain-usgs: Reverted back to simple string replacement here.

ale/kernel_access.py
jrcain-usgs: Tested with this version, but ended up using simple string replacement, as to not disturb the user's environment.

Kelvinrr (approved):

Kelvinrr:
We should figure out how to integrate this into the MK ale.loads paradigm. So users dont have to use python to make it work.

oleg-alexandrov:
@Kelvinrr Thanks a lot for the review and @jrcain-usgs for the work. I validated that this works. As to integrating this  into the MK ale.loads paradigm, that would likely be nice, but the fix does what it intends. When going through the NAIF path the metakernel gets loaded properly. The other path is with -k when an ISIS cub is loaded and meta kernels are not needed. I can't think of a use case where the bug would still be triggered (so unless the user does something custom).

PR #707, opened 2026-05-15
Default Reduction to None

acpaquette:
## Licensing


acpaquette:
@oleg-alexandrov After some discussion with @Kelvinrr , we considered linear reduction as the default to be a breaking change. The main reason is that if older versions of ISIS pick up the next ALE release with linear reduction as the default it would lead to different spice data produced on `spiceinit`'d cubes

oleg-alexandrov:
I thought ALE version was pinned to each version of ISIS no?

If the linear reduction is not the default, that is very problematic, as the cameras are very large. It is not possible to do properly bundle adjustment with many cameras, as almost all the time is spent in loading the cameras.  

I think the spice data produced with full samping vs resampling should be highly similar. That is because all the resampling does is reduce somewhat the extremely high sampling rate. The orientation of a sensor in orbit varies much slower than once per line.

oleg-alexandrov:
In other words, one can try to evaluate the effect of old ISIS with new ALE but results should be highly similar since the sampling has minimal effect.

By not having the sampling the default we will keep on propagating the huge camera bug for the foreseeable future.

acpaquette:
@oleg-alexandrov You're correct that most recent ISIS releases pin to the patch version of ALE that the ISIS release was built against. While this insulates the various versions of ISIS I have checked from unwanted data changes, it is still relavent. Users generating ISDs should be aware that they will be reducing the ephemeris data and losing some amount of accuracy depending on the size of the reduction.

Making linear reduction the default may not change much for CTX, but other instrusments like LRO LROC are effected see: https://github.com/DOI-USGS/ISIS3/issues/5224

While the ephemeris differences are minor there are significant pixel differences when comparing the projected products, with a maximum pixel difference of 121. I confirmed that this would occur when sub sampling every 10th ephemeris time.

ALE was originally applying a general linear sampling of every 64th ephemeris time for line scanners which changed the resulting projections of LROC images, especially NAC. The quaternions are obtained in a different way for LROC (since it loads 2 cks) vs most other instruments that only need one. For LROC NAC, a quaterion is loaded at each line in ALE. Then in ISIS, that set of quaternions is reduced [here](https://github.com/DOI-USGS/ISIS3/blob/40bca6a6415bb6fa67a3f150f5345034ff1801cc/isis/src/base/objs/SpiceRotation/SpiceRotation.cpp#L2480). Both the hermite cache reduction and the quaternion reduction mentioned above should happen in ALE rather than in ISIS so that the CSM can benefit from the reduced ephemeris data as well.

If we can add both of those reductions into ALE, we can make a reduced ISD the default and likely tick ALE to a 2.0.0 version

oleg-alexandrov:
What this means is that the Chandrayaan drivers must get the resampling back internally. Otherwise we get no-good ISDs with 200k lines.

acpaquette:
That is an amicable compromise. I will make it the default for chandrayaan2 again in this PR

oleg-alexandrov:
This should apply to both TMC and OHRC I believe. There should be testcases at least for one of them.

acpaquette:
@oleg-alexandrov Does this PR look good to you?

ale/drivers/chandrayaan_drivers.py
```diff
@@ -521,6 +521,26 @@ def ephemeris_start_time(self):
...
+        -------
+        : ndarray
+            ephemeris times split based on image lines and ephem_sample_rate passed
+            through props
+        """
+        if not hasattr(self, "_ephemeris_time"):
+            reduction = self._props.get('reduction', 'none').lower()
+            if (reduction == 'none'):
+                self._props['reduction'] = 'linear'
+            super().ephemeris_time
```
oleg-alexandrov: This line is confusing. It should maybe explicitly say: self._ephemeris_time = super().ephemeris_time. Then later it returns as now.

ale/drivers/chandrayaan_drivers.py
```diff
@@ -699,6 +719,26 @@ def ephemeris_stop_time(self):
...
```
oleg-alexandrov: Same thing here. Should be an explicit assignment, even if perhaps redundant.

oleg-alexandrov:
I asked Claude for a review. It points out:

 1. --ephem_sample_rate is silently dropped unless --reduction Linear. In file_to_isd:
  if reduction:
      props["reduction"] = reduction
      if reduction == "Linear":
          props["ephem_sample_rate"] = ephem_sample_rate
  For Chandrayaan-2, the driver override force-flips reduction to 'linear' when the user passes the default
  --reduction None. But because the CLI saw reduction == "None", ephem_sample_rate was never written into props,  so the base falls back to default 10. A CH-2 user doing --ephem_sample_rate 5 will get rate 10 with no warning. Either always pass the rate, or detect this case and warn.
 
I did not test this, but looks plausible.

  2. CLI choices are case-sensitive ('None', 'Linear'), but internal code lowercases. --reduction linear errors at
  argparse before the code ever sees it. Suggest type=str.lower and lowercase choices, or accept any case.
 
  3. if reduction: is always true. Default 'None' is a truthy string, so props["reduction"] = "None" is always set.
  Works because downstream uses .lower(), but reads as if the branch is conditional when it isn't.

oleg-alexandrov (approved):
Looks good. Since the Chandra tests did not change results, presumably the fix kicks in for them.

PR #709, opened 2026-05-19
Kplo shadowcam driver

oleg-alexandrov:
This adds a ShadowCam driver. It very closely emulates the corresponding ISIS camera model.

Releasing this will require releasing also USGSCSM, where the ShadowCam lens distortion is implemented.

This was tested by comparing the CSM camera with the ISIS implementation, using the ASP cam_test tool that shoots rays to ground with one moel (ISIS or CSM) and back to camera with the other one and vice versa. 

The pixel error discrepancy (ISIS vs CSM) is at most 0.01 pixels, which is acceptable. Looks like subtle differences in ISIS vs CSM about how camera position and direction are interpolated accounts for the difference.

I also mappojected a clip with cub vs csm camera and it looks good.

Tested with PDS datasets M074289249SE and M074296291SE.

Visual inspection was done with preexisting mapprojected images in PDS. Those are a little shifted, both relative to each other for these two datasets, and with what I get, but that is expected, as we don't know exactly their modality. 

Unit tests added.

## Licensing


ale/drivers/kplo_drivers.py
```diff
@@ -0,0 +1,170 @@
...
+
+    @property
+    def detector_center_line(self):
+        # Match LRO LROC NAC convention: leave BORESIGHT_LINE unshifted (only
+        # the SAMPLE direction gets the -0.5 ISIS-to-CSM half-pixel offset).
+        # LROC NAC ISIS driver only overrides detector_center_sample with -0.5
+        # and inherits the base detector_center_line = BORESIGHT_LINE.
+        # Applying -0.5 to LINE produces a systematic 0.5 px line residual in
+        # cam_test (cube vs CSM ISD) on ShadowCam.
+        return float(self.naif_keywords['INS{}_BORESIGHT_LINE'.format(self.ikid)])
```
acpaquette: By default the NaifSpice mix-in defines `detector_center_line` as:
```Python
  if not hasattr(self, "_detector_center_line"):
      self._detector_center_line = self.naif_keywords['INS{}_BORESIGHT_LINE'.format(self.ikid)]
  return self._detector_center_line
```
So I think this property can just be removed from the driver

ale/drivers/kplo_drivers.py
```diff
@@ -0,0 +1,170 @@
...
+    @property
+    def odtk(self):
+        od_k = self.naif_keywords['INS{}_OD_K'.format(self.ikid)]
+        if isinstance(od_k, (list, tuple)):
+            return [float(v) for v in od_k]
+        return [float(od_k)]
+
+    @property
+    def detector_center_sample(self):
+        return float(self.naif_keywords['INS{}_BORESIGHT_SAMPLE'.format(self.ikid)]) - 0.5
```
acpaquette: This looks similar enough to what's in the NaifSpice mixin with the -0.5.

It should probably be:
```Python
return super().detector_center_sample - 0.5
```

acpaquette (commented):
Overall looks good! A couple of small changes to address

ale/drivers/kplo_drivers.py
oleg-alexandrov: Good point. Done. Local tests did not change results. Let us see what the regression does.

ale/drivers/kplo_drivers.py
```diff
@@ -0,0 +1,164 @@
...
+    @property
+    def usgscsm_distortion_model(self):
+        return {"kplo_shadowcam": {"coefficients": self.odtk}}
+
+    @property
+    def odtk(self):
+        od_k = self.naif_keywords['INS{}_OD_K'.format(self.ikid)]
+        if isinstance(od_k, (list, tuple)):
+            return [float(v) for v in od_k]
+        return [float(od_k)]
```
acpaquette: I see a number of these checks, this is rather weird coming from the naif keywords group as the keyword is unlikely to change from either a single value into an interable (list, tuple, etc.).

If anything we might want an error if the object coming from the naif keywords group is not what is expected.

Was there a reason for these checks?

oleg-alexandrov: That was some defensive programming. There are places in the LRO driver for example where one considers naif_keywords['INS{}_BORESIGHT_LINE'] either a value or an array.

But for ShadowCam this seems unnecessary so I went back to assuming a simple value. I tested offline with isd_genrate and added another test that goes through motions of querying those.

oleg-alexandrov:
I fixed a merge conflict in the changelog. The tests are running.

acpaquette (approved):

PR #712, opened 2026-05-27
Auto-correct stale metakernel paths using ALESPICEROOT or ISISDATA

oleg-alexandrov:
This is a fix on top of @jrcain-usgs's pull request https://github.com/DOI-USGS/ale/pull/703 that addresses @Kelvinrr's points.

The implementation was adjusted to read the .tm file that has the hard-coded /usgs/cpkgs/isis3/data path, make a temporary copy replacing that with ALESPICEROOT (or ISISDATA if ALESPICEROOT is not set), and load the corrected .tm.

This avoids any changes to the actual run-time environment and fixes the issue at the core, which is that a lot of metakernels shipped by USGS have the hard-coded path above.

This is a long-standing issue for at least 3 years, and while there are documented workarounds the user is supposed to make, fixing this is quite important as the new Chandrayaan-2 data is able to make use of metakernels as well. This problem also affects MSL, LRO, and MRO.

## Licensing

### DIFF
--- a/ale/kernel_access.py
+++ b/ale/kernel_access.py
@@ -3,6 +3,7 @@ from glob import glob
 from itertools import groupby, chain
 import os
 from os import path
+from tempfile import NamedTemporaryFile
 import re
 import warnings
 from collections.abc import Iterable
@@ -12,6 +13,7 @@ import pvl
 import json
 
 from ale import spice_root
+from ale import logger
 from ale.util import get_isis_preferences
 from ale.util import get_isis_mission_translations
 from ale.util import read_pvl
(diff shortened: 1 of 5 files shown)


ale/kernel_access.py
```diff
@@ -20,6 +22,58 @@
...
+        new_root = os.environ.get('ISISDATA')
+        if new_root is not None:
+            warnings.warn(f"ALESPICEROOT not set, using ISISDATA={new_root}")
+
+    with open(metakernel, 'r') as f:
+        text = f.read()
+
+    if isinstance(new_root, str) and old_root in text:
+        corrected = text.replace(old_root, new_root)
+        tmp = NamedTemporaryFile(mode='w', suffix='.tm', delete=False)
```
Kelvinrr, line 68: we should have an option to clean this file up somehow.

oleg-alexandrov: I am not sure how. The current approach assumes a hard-coded known path that is replaced. 

The alternative would be to have such files always with relative paths end ensure the other pull request I put in which got merged can read such things. Looks like it should, as I tested, but when we produce new tm files need to test again.

In either case, looks like @jrcain-usgs is on it in the original pull request, so I will stay out of the way. This will need to be closed unmerged when the other one makes it in. I will test that one as well when is up.

oleg-alexandrov: To add, overwriting user's tm files is likely not a good idea, and many times those are shared with others and may not have the right permissions. So some internal fix is likely the only option.

Kelvinrr: personally, I am not a fan of either implementation right now, they both add a lot of complexity. I think it'll be better to simply not support MKs

oleg-alexandrov: Kelvin, for MSL we do not have alternatives, I think. This is documented as a valid use case. 

I would advise merging these in as they are bug fixes, then have a larger discussion about what to do going forward. There are likely a few interested parties and decisions to make.

Kelvinrr:
closing in favor of https://github.com/DOI-USGS/ale/pull/703

PR #713, opened 2026-05-27
Fix nadir velocity axis computation to match ISIS

oleg-alexandrov:
This is a rebase of @acpaquette's #625 on current main, with updated test gold data.

The nadir velocity axis was computed by comparing `trans_x[0]` (the offset) against `trans_x[1]` (a directional component), which is incorrect. The fix compares `trans_x[1]` against `trans_x[2]` (the two directional components), matching the ISIS implementation.

The nadir codepath is only triggered when `nadir=True` is passed in props, which is used by Chandrayaan-1 M3 and MSL Mastcam.

## Licensing


ale/base/data_naif.py
```diff
@@ -675,7 +675,7 @@ def frame_chain(self):
                                                         mission=self.spiceql_mission, 
                                                         searchKernels=self.search_kernels, 
                                                         useWeb=self.use_web)[0][trans_x_key]
-                if (trans_x[0] < trans_x[1]):
+                if (trans_x[1] < trans_x[2]):
```
acpaquette, line 679: @oleg-alexandrov I looked at this a second time and the naif keyword gotten in ISIS is `"INS{self.ikid}_ITRANSX` rather than `"INS{self.ikid}_ITRANSL`. We should likely do the same here on line 674

oleg-alexandrov:
Thanks for catching this. Pushed `f1cf34e`, which switches the keyword from `INS<ikid>_ITRANSL` to `INS<ikid>_TRANSX` (keeping the `[1] < [2]` comparison this PR already fixed).

One small clarification on the comment: the keyword is **`TRANSX`**, not `ITRANSX` — there is no `INS..._ITRANSX` pool keyword. ISIS reads `TRANSX` in `SpiceRotation::setEphemerisTimeNadir` (`SpiceRotation.cpp`, ctor lines ~121-135):

```cpp
QString key = "INS" + toString(frameCode) + "_TRANSX";
SpiceDouble transX[2];
gdpool_c(key.toLatin1().data(), 1, 2, &number, transX, &found); // start=1 -> TRANSX[1], TRANSX[2]
p_axisV = 2;
if (transX[0] < transX[1]) p_axisV = 1;                          // i.e. TRANSX[1] < TRANSX[2]
```

Because `gdpool_c` starts at element 1, ISIS is comparing `TRANSX[1]` (d x / d sample) against `TRANSX[2]` (d x / d line) — which is exactly the `[1] < [2]` this PR introduced, just applied to the wrong keyword. `TRANSX` and `ITRANSL` are different transforms, so for some detectors they disagree. Chandrayaan-1 M3 (`INS-86520`) is one:

- `TRANSX  = [-0.64, -4.0055e-05,  0.0270]`  -> `[1] < [2]` **true**  -> axis **1**  (ISIS)
- `ITRANSL = [23.70,  37.037,     -0.05495]` -> `[1] < [2]` **false** -> axis **2**

So the index fix alone, while still reading `ITRANSL`, actually selected the wrong axis (2) for M3; reading `TRANSX` restores axis 1, matching ISIS.

### Validation

I built the same M3 image two ways — an ISIS camera (nadir pointing) and an ALE-generated CSM model — and measured the angle between their look directions at every pixel with ASP's `cam_test`. With both cameras driven by the **same** spacecraft SPK so positions match:

| ISIS nadir cube vs | camera center diff | look-direction diff (median) |
|---|---|---|
| **ALE with the fix** (`TRANSX`, axis 1) | 0.0001 m | **0.0024 rad (~0.14 deg)** |
| ALE with the bug (`ITRANSL`, axis 2) | 0.0001 m | 0.173 rad (~9.9 deg) |

The fix agrees with ISIS ~70x better than the bug. The residual 0.14 deg is ISIS's own nadir interpolation (the cube stores the nadir rotation at two endpoints and interpolates across the lines, while ALE computes it per line), not an ALE discrepancy.

### How to reproduce, step by step

Requires ISIS (with a full `chandrayaan1` area in `$ISISDATA`), this ALE branch, and ASP (`cam_test`). The M3 cropped test scene already ships in the ISIS repo under `isis/tests/data/chan1m32isis/forwardAscending/`.

```bash
# 1. Ingest the cropped M3 scene
chan1m32isis \
  from=.../chan1m32isis/forwardAscending/M3T20090630T083407_V03_L1B_cropped.LBL \
  loc=.../M3T20090630T083407_V03_LOC_cropped.IMG \
  obs=.../M3T20090630T083407_V03_OBS_cropped.IMG \
  to=m3.cub

# 2. Make ISIS use nadir pointing (forces SpiceRotation::setEphemerisTimeNadir)
spiceinit from=m3.cub cknadir=yes ckrecon=no ckpredicted=no cksmithed=no

# 3. ALE nadir ISD from the SAME cube (so both sides use the same kernels/SPK)
isd_generate -N m3.cub        # -> m3.json

# 4. Compare the two camera models geometrically
cam_test --image m3.cub --cam1 m3.cub --session1 isis \
         --cam2 m3.json --session2 csm --sample-rate 1
```

With the fix the "camera direction diff norm" collapses to ISIS's interpolation floor (sub-0.2 deg); reverting `data_naif.py` to `ITRANSL` reproduces the ~8-10 deg disagreement.

### Note on the second copy

There is a near-identical nadir block in `ale/base/type_sensor.py` (`Cahvor.frame_chain`), reached only by MSL Mastcam. I left it untouched on purpose: MSL is a CAHVOR framer with no `INS..._TRANSX` keyword (its `focal2pixel` comes from the CAHVOR model), so the `TRANSX` form is inapplicable there and there is no ISIS `setEphemerisTimeNadir` equivalent to match against. Happy to revisit if you'd prefer it handled here too.

---
*Full disclosure: this analysis, the fix, and the validation above were done by Claude (an AI assistant from Anthropic) working on Oleg's machine, with Oleg directing. Flagging it so the cam_test recipe and numbers can be checked independently rather than taken on faith.*

oleg-alexandrov:
CI is green after this. Switching the keyword to `TRANSX` flips the nadir velocity axis for a few instruments, so I regenerated the Chandrayaan-1 M3 and MRFFR nadir gold ISDs and updated the MRO CTX `test_nadir_sensor_orientation` expectation to match. All Linux/macOS/Windows checks pass.

acpaquette:
@oleg-alexandrov Can this get a changelog? I will merge this after that

oleg-alexandrov:
Done.

acpaquette (approved):

PR #714, opened 2026-05-29
Hirise pds3 driver

oleg-alexandrov:
This is a careful effort to sort out https://github.com/DOI-USGS/ale/pull/702.

The goal of that seems worthy: A PDS3 label driver for HiRISE that enables ISD generation directly from raw PDS EDR files without requiring an ISIS cube.

That is useful because HiRISE cubes are huge and hard to make, and bypassing that can greatly speed up processing. It also parallels nicely the existing CTX driver duo.

That pull request was obviously incomplete. I used Claude to fix this. The goal here is to leverage the AI to help ensure very thorough testing and careful reviewing, rather than add to the developer workload.

Here is what is changed.

Synched up to main. The SpiceQL lookups now use the pyspiceql.strSclkToEt(...) / translateNameToCode(...) / findMissionKeywords(...) calling convention already used by the sibling MroHiRiseIsisLabelNaifSpiceDriver.

Addressed all of @acpaquette's review comments:

image_samples override removed. The base Pds3Label.image_samples already returns the correct value.

detector_center_line / detector_center_sample is kept at 0 for parity with MroHiRiseIsisLabelNaifSpiceDriver.

spacecraft_name simplified. HiRISE EDR labels carry only INSTRUMENT_HOST_NAME (no SPACECRAFT_NAME), so this now resolves via name_lookup[super().platform_name], matching how MroCrismIsisLabelNaifSpiceDriver does it.

platform_name override removed. The base Pds3Label.platform_name already returns INSTRUMENT_HOST_NAME.

## Testing

Two test classes added to tests/pytests/test_mro_drivers.py, using the same scaffolding as the existing tests - conftest.get_image_label, the test-data directory layout, and mocked pyspiceql calls:

- test_hirise_pds_naif - per-property unit tests for the new driver, modeled directly on the existing test_ctx_pds_naif and test_hirise_isis_naif.

- test_hirise_pds_vs_isis_equivalence - instantiates the PDS3 driver and the ISIS-label driver MroHiRiseIsisLabelNaifSpiceDriver on the *same* observation and asserts they go through the same motions: every ISD-relevant property agrees between the two drivers (instrument id, sensor name, clock count, un-binned rate, exposure duration, image samples/lines, sensor frame, detector center, ephemeris start time, CCD ikid). spacecraft_name is excluded by design - the PDS3 driver maps it to the NAIF short name MRO while the ISIS driver keeps the full MARS RECONNAISSANCE ORBITER.

- A trimmed PDS3 EDR label fixture (data/PSP_001446_1790_BG12_0/PSP_001446_1790_BG12_0_pds3.lbl) is added - it is the *same observation* as the existing HiRISE ISIS-label fixture and kernels, so the two drivers are directly comparable. It is trimmed the same way as the existing CTX _pds3.lbl fixture, with all retained values verbatim from the archive label (HiRISE PDS node).

There are 19 new tests.

## Licensing

### DIFF
--- a/ale/drivers/mro_drivers.py
+++ b/ale/drivers/mro_drivers.py
@@ -569,20 +569,19 @@ class MroHiRisePds3LabelNaifSpiceDriver(LineScanner, Pds3Label, NaifSpice, Radia
     @property
     def spacecraft_name(self):
         """
+        HiRISE PDS3 EDR labels carry INSTRUMENT_HOST_NAME rather than
+        SPACECRAFT_NAME, so resolve the name from the base platform_name
+        (which returns INSTRUMENT_HOST_NAME for a PDS3 label).
+
         Returns
         -------
         : str
           Spacecraft name for NAIF lookups
         """
         name_lookup = {
-            'MARS_RECONNAISSANCE_ORBITER': 'MRO',
-            'MARS RECONNAISSANCE ORBITER': 'MRO',
-            'MRO': 'MRO',
+            'MARS RECONNAISSANCE ORBITER': 'MRO'
         }
-        # PDS3 HiRISE labels use INSTRUMENT_HOST_NAME, not SPACECRAFT_NAME
-        raw = self.label.get('SPACECRAFT_NAME',
-                             self.label.get('INSTRUMENT_HOST_NAME', ''))
-        return name_lookup[raw]
+        return name_lookup[super().platform_name]
 
     @property
     def sensor_name(self):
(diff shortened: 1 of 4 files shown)


ale/drivers/mro_drivers.py
```diff
@@ -535,6 +535,282 @@ def platform_name(self):
...
+        the ISD usable within ISIS but is not the true USGSCSM boresight line.
+        ISIS itself sets the detector origin/offset in the HiRise camera model
+        rather than reading these from the ISD.
+
+        Returns
+        -------
+        : float
+          Detector line of the principal point
+        """
+        return 0
```
Kelvinrr, line 765: Adam mentioned this in the previous PR, but this is consistent with the existing driver, if we want to fix it we can do it in another PR.

Kelvinrr (approved):
Thanks for taking this on, we were on the fence about taking it over so glad someone else made the decision

PR #716, opened 2026-06-02
Fix KeyError for MEX HRSC stereo channels in spiceql_mission

oleg-alexandrov:
Generating a CSM ISD for the Mars Express HRSC stereo channels currently fails. isd_generate finds the correct driver (MexHrscIsisLabelNaifSpiceDriver) but it throws before producing an ISD, so ALE falls through to "No viable Driver for Label". Exercising the driver directly shows the real error:

    File ".../ale/base/data_naif.py", line 887, in spiceql_mission
        return spiceql_mission_map[self.instrument_id]
    KeyError: 'MEX_HRSC_S1'

Root cause: the HRSC head has nine filter channels, and each has its own instrument_id (the ISIS Archive/DetectorId, e.g. MEX_HRSC_S1, MEX_HRSC_S2, MEX_HRSC_RED, and so on). The spiceql_mission_map only contains MEX_HRSC_IR, so every other channel, including both stereo channels used for DEM generation, raises a KeyError in spiceql_mission. This is not a kernel problem and not a SpiceQL problem; it never reaches a SpiceQL call.

Fix: override spiceql_mission to return hrsc in both HRSC head drivers (the PDS3 and ISIS label variants), so all nine channels resolve to the same mission. This mirrors what the KPLO and Mariner drivers already do, which override spiceql_mission rather than relying on the map.

I added regression tests for both drivers in test_mex_drivers.py. Without the fix they fail with KeyError MEX_HRSC_S1; with the fix they pass. The full test_mex_drivers.py passes, 54 tests.

Reproduction (ISIS and ALE), following the HRSC example in the Ames Stereo Pipeline docs:

    # Fetch one observation's two stereo channels
    wget https://pds-geosciences.wustl.edu/mex/mex-m-hrsc-3-rdr-v4/mexhrs_4000/data/1995/h1995_0000_s13.img
    wget https://pds-geosciences.wustl.edu/mex/mex-m-hrsc-3-rdr-v4/mexhrs_4000/data/1995/h1995_0000_s23.img

    # Ingest and spiceinit
    hrsc2isis from=h1995_0000_s13.img to=h1995_0000_s13.cub
    hrsc2isis from=h1995_0000_s23.img to=h1995_0000_s23.cub
    spiceinit from=h1995_0000_s13.cub ckpredicted=true
    spiceinit from=h1995_0000_s23.cub ckpredicted=true

    # Fails on main, works with this PR.
    # Here must set -k to furnish the kernels the cube was spiceinit'd with. The bare
    # "isd_generate <cube>" command requires a prebuilt SpiceQL inventory.
    isd_generate -k h1995_0000_s13.cub -v h1995_0000_s13.cub
    isd_generate -k h1995_0000_s23.cub -v h1995_0000_s23.cub

On main, isd_generate ends with "No viable Driver for Label" for both channels; with this PR it produces h1995_0000_s13.json and h1995_0000_s23.json. The stereo-channel DetectorId values MEX_HRSC_S1 and MEX_HRSC_S2 were verified on real spiceinit'd cubs.

Prepared with Claude Code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ release.
 ### Fixed
 - Fixed undefined behavior in Rotation::toRotationMatrix by normalizing the quaternion before converting it to a rotation matrix. See [Eigen 3.4.1 Docs](https://libeigen.gitlab.io/eigen/docs-3.4/classEigen_1_1QuaternionBase.html#a8cf07ab9875baba2eecdd62ff93bfc3f) [#711](https://github.com/DOI-USGS/ale/pull/711)
 - Nadir velocity axis is now computed from the `INS<ikid>_TRANSX` keyword with a `[1] < [2]` index comparison, matching ISIS's `SpiceRotation::setEphemerisTimeNadir`. [#713](https://github.com/DOI-USGS/ale/pull/713)
+- Fixed `KeyError: 'MEX_HRSC_S1'` when generating an ISD for the MEX HRSC stereo channels (and any non-IR filter channel). The HRSC drivers now override `spiceql_mission` to return `hrsc`, matching the pattern used by the KPLO and Mariner drivers, instead of looking up the filter-specific instrument id in `spiceql_mission_map`, which only lists one channel. [#716](https://github.com/DOI-USGS/ale/pull/716)
 
 ## [1.2.0] - 2026-05-20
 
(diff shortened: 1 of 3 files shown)


oleg-alexandrov:
Tests pass.

Kelvinrr (approved):

PR #726, opened 2026-07-02
Add Lagrange interpolation of quaternions to ale::Orientations

oleg-alexandrov:
This adds order-8 Lagrange interpolation of quaternions to ale::Orientations, matching ISIS SpiceRotation. 

ISIS reconstructs sensor pointing between cache nodes with an order-N Lagrange window. When ISIS delegates rotation interpolation to ALE, ale::Orientations reconstructed with a two-node SLERP instead, which is too coarse, and inconsistent with quaternion interpolation in USGSCSM as well.

The orientations are converted to quaternions when set, and made sign-continuous at that time. That is because while a quaternion and its negative represent the same rotation, a sign jump between adjacent nodes gives bad results when interpolating the components with Lagrange. This sign handling has been problematic in the past. 

The existing behavior is unchanged. The new path is off by default and is selected with the LAGRANGE_ROTATION interpolation type. The caller, most likely ISIS, sets that flag to get the Lagrange reconstruction.

Tests are added for exact reproduction at node times, accuracy versus SLERP on a non-constant-rate rotation, and sign-flip invariance.

All tests pass.

Developed with Claude (AI) assistance. The code was thoroughly reviewed. Comments and API were adjusted as needed.

## Questions

This change covers ale::Orientations, which is the interpolation ISIS uses at runtime when it delegates to ALE. 

It is not clear whether anything is also needed on the ISD-generation side, that is, in the Python code. The current default formatter, to_isd, writes the raw pointing table (ephemeris_times and quaternions) straight from the frame chain, without resampling, and USGSCSM does its own Lagrange interpolation of that table when it reads the ISD. 

Could you confirm whether the C++ path here is sufficient, or whether there is a need for this logic in Python as well (a parallel implementation). 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ release.
 ## [Unreleased]
 
 ### Added
+- Added order-8 Lagrange interpolation of quaternions to `ale::Orientations`, matching ISIS SpiceRotation, selectable via the new `LAGRANGE_ROTATION` interpolation type. [#726](https://github.com/DOI-USGS/ale/pull/726)
 - Re-enabled and fixed the TGO CaSSIS driver, which now emits the CaSSIS rational distortion. Validated against ISIS to within ~0.013 pixel across 130 framelets of two real stereo pairs. [#720](https://github.com/DOI-USGS/ale/pull/720)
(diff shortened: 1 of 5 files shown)


Kelvinrr (approved):
we need to expose this in ALE and USGSCSM to reduce the ISD sizes as well.

PR #731, opened 2026-07-16
Hermite Position Reduction

acpaquette:
This PR adds the ability to reduce the number of ephemeris positions in ALE using the exsiting hermite spline creation code in the `States` class in the C++ side of ALE

## Licensing


ale/driver_verification.py
```diff
@@ -49,8 +50,9 @@ def run_spiceinit_isis(image_path):
     
     # Run spiceinit with ISIS
     try:
-        subprocess.run(['spiceinit', f'from={image_path}']) # I believe this is where the crashes are coming from
-    except:
+        subprocess.run(['spiceinit', f'from={image_path}', "spkpredicted=true"])
+    except Exception as e:
+        logger.error(e.what())
```
Kelvinrr: Maybe im misremembering, but I dont think python functions have a `.what` built in? I think you do `logger.error(e)`

setup.py
```diff
@@ -13,6 +14,18 @@
 # prerequisite: setuptools
 # http://pypi.python.org/pypi/setuptools
 
+conda_path = os.environ["CONDA_PREFIX"]
```
Kelvinrr: minor, but might be worth throwing an exception showing that it's expected to run in a conda env.

ale/drivers/tgo_drivers.py
```diff
@@ -108,116 +108,3 @@ def detector_center_line(self):
...
-    def sensor_position(self):
-        """
-        CaSSIS sets LT_SURFACE_CORRECT with LIGHTTIME_CORRECTION=LT+S, so ISIS
-        applies the surface light-time correction. The shared
-        NaifSpice.sensor_position samples the target body at the raw ephemeris
-        time in that branch, but ISIS samples it at the surface-light-time
-        adjusted time (ephem - obs_tar_lt + radius_lt); the body moves along its
-        orbit during that interval, which otherwise leaves a constant
-        tens-of-meters camera-center bias versus ISIS. This override applies that
-        for CaSSIS only, so the shared path is unchanged for every other sensor.
```
Kelvinrr, line 260: why remove from TGO?

ale/base/data_naif.py
```diff
@@ -628,10 +633,46 @@ def sensor_position(self):
...
+                a, b, c = self.target_body_radii
+                coord = spice.surfpt([0, 0, 0], middle_nadir_lv, a, b, c)
+                radius = np.linalg.norm(coord)
+                # Ensure the radius is below the spacecraft (largely for landed sensors)
+                radius = min(radius, middle_magnitude - 0.0001)
+
+                # Compute the altitude based on the radius of the ground coordinate
+                # minus the magnitude of the spacecraft position
+                altitude = (middle_magnitude - radius) * 1000
+                tol = self.pixel_size * altitude / self.focal_length / 100.
```
Kelvinrr, line 670: probably need comments on conversions here. Altitude is being converted from KM to meters, I assume, but I do not know what deviding the tol by 100 is doing. Altitude is in M but tol is in some other unit?

ale/base/data_naif.py
```diff
@@ -628,10 +633,46 @@ def sensor_position(self):
...
             self._ephem = ephem
+            
+            reduction = self._props.get('reduction', 'none').lower()
+            if reduction == 'hermite' and len(self._position) > 3:
+                logger.debug("Applying hermite reduction to positions")
+                state_list = [ale_c.State(i) for i in np.append(self._position, self._velocity, axis=1)]
+                states = ale_c.StatesFromStateVec(self._ephem, state_list, 1)
+
+                # Get middle position (it will be in meters)
+                middle_state = states.getState(self.center_ephemeris_time, 1)
```
Kelvinrr, line 647: It's used in the alt compute later down the line where its multiplied by 1000, doesn't this mean it's in KM not M?

tests/pytests/data/isds/lro_isis_isd.json
```diff
@@ -9988,7 +9988,7 @@
   "instrument_position": {
     "spk_table_start_time": 302228504.36824864,
     "spk_table_end_time": 302228558.33807606,
-    "spk_table_original_size": 25,
+    "spk_table_original_size": 2,
```
Kelvinrr, line 9991: maybe I forgot what this value was, but if it's the number of ephems, it seems the actual number in the ISD did not change.

ale/base/data_naif.py
```diff
@@ -573,15 +575,18 @@ def sensor_position(self):
...
+                ssb_tars_kwargs = {**ephem_kwargs,
+                                   "target": target,
+                                   "observer": "SSB",
+                                   "frame": "J2000",
+                                   "abcorr": "NONE",
+                                   "mission": self.spiceql_mission}
+                if self.instrument_id == "TGO_CASSIS":
+                    adjusted_time = ephem - obs_tar_lts + radius_lt
+                    ssb_tars_kwargs["startEt"] = adjusted_time[0]
+                    ssb_tars_kwargs["stopEt"] = adjusted_time[-1]
```
Kelvinrr, line 588: Is this to avoid overwriting the entire function? Is there a clean way to use ALE's paradigm instead of an if statement?

Kelvinrr (commented):
Biggest concern is the unit question, just some clarity needed there.

ale/drivers/tgo_drivers.py
acpaquette: This is related to adding the TGO_CASSIS specific check in the `sensor_position` property. Currently, we are maintaining two implementations of the light time correction `sensor_position` code, one in the base `Driver` class the other in the `TGO_CASSIS` concrete driver. Currently, there is only one driver but it would need the duplicate implementation if any new drivers are added for `TGO_CASSIS` that use the `NaifSpice` mixin.

ale/base/data_naif.py
acpaquette: I don't think this would be a property that a user would pass in. The way I see handling it without an explicit check is to add a new property on the base `NaifSpice` mixin that is a boolean for `adjusted_time` and default it to `false`. Then the concrete drivers could override it as necessary. It would only be used by concrete Cassis drivers that use the `NaifSpice` mixin

ale/base/data_naif.py
acpaquette: Dividing by 100 is just setting the tolerance to 1/100th of the computed pixel resolution

tests/pytests/data/isds/lro_isis_isd.json
acpaquette: This is a weird case where we are using a cropped LRO cube that has been reduced to one line but retains the `InstrumentPosition` table as if it was the full cube. So the change to the formatter is changing the `spk_table_original_size` in the resulting ISD.

If we wanted to get the actual value in there from the ISIS table:
```
  Name                 = InstrumentPosition
  StartByte            = 117778
  Bytes                = 1400
  Records              = 25
  ByteOrder            = Lsb
  CacheType            = HermiteSpline
  SpkTableStartTime    = 302228504.36825
  SpkTableEndTime      = 302228558.33808
  SpkTableOriginalSize = 816.0
  Description          = "Created by spiceinit"
  Kernels              = $lro/kernels/spk/fdf29r_2009182_2009213_v01.bsp
```
It should be `816.0` but I can't think of an easy way to transfer it

Kelvinrr: I figured it was something like that.

ale/base/data_naif.py
```diff
@@ -628,10 +633,52 @@ def sensor_position(self):
...
+                state_list = [ale_c.State(i) for i in np.append(self._position, self._velocity, axis=1)]
+                states = ale_c.StatesFromStateVec(self._ephem, state_list, 1)
+
+                # Get middle position (it will be in kilometers)
+                middle_state = states.getState(self.center_ephemeris_time, 1)
+                middle_position = np.asarray([middle_state.position.x, 
+                                              middle_state.position.y, 
+                                              middle_state.position.z])
+
+                # Compuet a normalized nadir look vector
```
Kelvinrr: Compuet -> Compute

Kelvinrr (approved):

PR #734, opened 2026-07-27
Fix Viking Orbiter VIS CSM camera: detector center and IsisSpice driver

oleg-alexandrov:
The ALE Viking driver returned a detector center of (0, 0) instead of the array center at sample 602, line 528 that the ISIS VikingCamera model uses. This placed the CSM boresight in the wrong location and offset every pixel by roughly 800 pixels from the ISIS camera, which made the Viking CSM camera unusable.

Separately, VikingIsisLabelIsisSpiceDriver did not define the focal length, detector center, or sensor model version at all. As a result, isd_generate with the only_isis_spice option produced an ISD with a null focal length and a null detector center, which usgscsm could not build a camera from.

This change makes both Viking drivers return the detector center at sample 602, line 528, and adds the focal length, detector center, and sensor model version to the IsisSpice driver, matching the NaifSpice driver and the ISIS camera. The values follow ISIS VikingCamera.cpp, which calls SetDetectorOrigin(602.0, 528.0) for all Viking configurations and sets the per-camera focal length.

Validation used campt on a Viking Orbiter 1 frame at about 63 meters per pixel, comparing the ISIS camera and the CSM camera on the same cube. See more in follow up comment below.

The remaining difference grows toward the frame edge and is the Viking VIS optical distortion. ISIS models this with a reseau distortion map, while the distortion-free CSM frame model does not. We do not plan to implement the reseau grid distortion in usgscsm, so this small residual is expected, and the CSM camera is a distortion-free approximation of the ISIS camera.

This updates the Viking driver unit tests, adds a unit test class for the IsisSpice driver, updates the golden ISDs, and adds a changelog entry.

Done with Claude (AI) assistance.


oleg-alexandrov:
Viking cub and csm file for testing the replication of the fix are attached.   

[vo1_f913a15.cub.txt](https://github.com/user-attachments/files/30408430/vo1_f913a15.cub.txt)
[vo1_f913a15.json](https://github.com/user-attachments/files/30408440/vo1_f913a15.json)

Rename the cub file to remove the txt extension.

To reproduce the check, compare the ISIS camera and the CSM camera on the same pixel with campt. First the ISIS camera, with no CSM:

    campt from=vo1_f913a15.cub sample=602 line=528

Then the CSM camera, reading the ISD directly through the CSM parameter:

    campt from=vo1_f913a15.cub csm=vo1_f913a15.json sample=602 line=528

After the fix, both commands report essentially the same ground point for that pixel, a planetocentric latitude of about -6.88 and an east longitude of about 289.27. At the detector center the two agree to about 7 pixels, and toward the frame corner to about 15 pixels. Before the fix the CSM command was off by about 800 pixels. The remaining difference is the Viking VIS optical distortion, which ISIS models and the distortion-free CSM frame model does not.

acpaquette (approved):
Looks good!

PR #434, opened 2022-01-11
GalileoSsi Driver and Tests

acpaquette:
Adds GalileoSsi driver and tests to ale

PR #736, opened 2026-08-21
Fix single-coefficient distortion odtk to always return a list

oleg-alexandrov:
`isd_generate` produces an LRO NAC ISD that USGSCSM cannot load. `cam_test` and `bundle_adjust` fail with `Could not parse the lrolrocnac distortion model coefficients` and no camera is built.

Cause: the distortion coefficient is written to the ISD as a scalar instead of a list. The driver `odtk` returns a single-valued NAIF keyword unchanged, and such a keyword now comes back as a scalar rather than a 1-element list, so a scalar lands in the ISD. USGSCSM reads the coefficients as an array and throws. 

This bug showed up when `naif_keywords` moved to SpiceQL's `findMissionKeywords` (#621), which returns a single-valued keyword as a scalar; the earlier lookup returned a 1-element list, so `odtk` never needed to wrap it.

The same unwrapped pattern is present in three drivers, all fixed here: LRO NAC (`lrolrocnac`), Dawn FC (`dawnfc`, from `RAD_DIST_COEFF`), and Galileo SSI radial distortion (from `K1`). The KPLO ShadowCam driver already wraps its coefficient in a list, so it was unaffected. Galileo SSI has emitted a scalar since its original driver (#434); it went unnoticed because that path is not exercised through USGSCSM. The fix only makes the coefficient a list, matching the `RadialDistortion` contract that every other radial driver already follows.

Fix: each of these `odtk` properties coerces to a list. Added a scalar-keyword case to the LRO NAC and Dawn FC `odtk` tests, and updated the Galileo SSI `odtk` test, which fed a scalar and asserted a scalar, to assert a list.

The test failures are in the C++ tests, not in this change. On macOS the C++ tests no longer compile: a freshly resolved googletest now requires C++14 (`error: C++ versions less than C++14 are not supported`), which the test build does not set. On Linux they build but the `EvaluateCubicHermiteFirstDeriv.InvalidVelocities` test aborts. This PR changes only Python driver code and adds no C++, every Python test passes on all platforms (Windows is green), and this C++ is identical to `main`, whose last CI run passed, so this is CI toolchain drift and not a result of these changes.

Disclaimer: this work was done with AI assistance (Claude).

- [x] I dedicate any and all copyright interest in this software to the public domain.


oleg-alexandrov:
Update on the test failures: they are in the C++ tests, not in this change. 

On macOS the C++ tests no longer compile because a freshly resolved googletest now requires C++14 (`error: C++ versions less than C++14 are not supported`), which the test build does not set. 

On Linux they build but the `EvaluateCubicHermiteFirstDeriv.InvalidVelocities` test aborts. This PR changes only Python driver code and adds no C++.

Every Python test passes on all platforms (1029 passed, Windows green), and this C++ is identical to `main`, whose last CI run passed. So these failures are CI toolchain drift that will affect `main` too, independent of these changes.

oleg-alexandrov:
I fixed an edit conflict (in changelog). All tests pass.

Kelvinrr (commented):
I think this looks good, did you check if this is exhaustive and there are not other missions with the same issue?

oleg-alexandrov:
Claude went through every odtk in ale/drivers/. The scalar-not-a-list bug can only happen where a   driver passes a NAIF OD_K-type keyword straight through and that keyword may hold a single coefficient. That's exactly the three now fixed (LRO NAC, Dawn FC, Galileo SSI).

  All the others are safe by construction:
  - Cassini ISS and MGS MOC → return hardcoded multi-element lists.
  - KPLO ShadowCam → already wraps as [float(...)].
  - OSIRIS-REx and LRO WAC → slice/iterate a multi-coefficient OD_K, so always a list.

oleg-alexandrov:
Anything even remotely related to distortion was reviewed and fixed preemptively. Additional tests added.

oleg-alexandrov:
This needed a retrigger for the tests to pass.

Kelvinrr (approved):

PR #737, opened 2026-08-25
Remove isis_formatter and usgscsm_formatter- #698

oleg-alexandrov:
This is a reimplemenation of https://github.com/DOI-USGS/ale/pull/698 to remove a merge conflict.

This address a 6-year old dead code issue that should have been fixed 2 releases ago. 

That PR was being given a year-long strategic silence treatment, which is not good for any active repository. 

## Licensing


Kelvinrr:
There is a semantic versioning issue here in that this is technically API breaking. Since everything we do uses semantic versioning, we would need to do a 2.0 release for this. Which we can do, we would just need keep that in mind when we next do a release. We need to do a release now, what's there now out there, and then we can we can merge this and make a 2.0 next time.

Kelvinrr:
once https://github.com/DOI-USGS/ale/pull/736 is merged, we can do a release then merge this.

oleg-alexandrov:
Is there a plan for this, say before end of year. 

Strictly speaking it isn't as if this is a rush, but the whole thing has been going on for a while, and clearing the slate may be a good thing, when the time is right.

Kelvinrr:
@oleg-alexandrov I think it would be good to get that other PR pre-2.0, so once the other PR merges, we can merge this then release (like the same day or day after this is merged), then merge this one so we can have a clean break.

Kelvinrr (approved):

PR #739, opened 2026-08-29
Fix the Cassini ISS drivers so the CSM camera matches ISIS

oleg-alexandrov:
Made the three Cassini ISS drivers produce a CSM camera that matches the ISIS camera:

- ISIS-SPICE driver: emit the radial optical distortion (was none).
- ISIS-SPICE and NAIF-SPICE drivers: subtract 0.5 from the detector center (ISIS 0.5-based to CSM 0-based).
- NAIF-SPICE focal_length: no longer raises AttributeError when INS_FOCAL_LENGTH is defined.
- NAIF-SPICE ephemeris_start_time: take the scalar from utcToEt (was a tuple).

Validated against ISIS with cam_test on a Cassini NAC image. The pixel error is 5e-4 pixel.  

The ISIS-SPICE, NAIF-SPICE and PDS3 drivers now agree with each other and with ISIS. 

Updated the ISIS-SPICE golden ISD and the ephemeris unit tests, and added a focal_length regression test. USGSCSM needs no change.

Developed with assistance from Claude (AI).

This was manually edited for readability.

## Licensing
This project is mostly composed of free and unencumbered software released into the public domain, and we are unlikely to accept contributions that are not also released into the public domain. Somewhere near the top of each file should have these words:

> This work is free and unencumbered software released into the public domain. In jurisdictions that recognize copyright laws, the author or authors of this software dedicate any and all copyright interest in the software to the public domain.

- [x] I dedicate any and all copyright interest in this software to the public domain. I make this dedication for the benefit of the public at large and to the detriment of my heirs and successors. I intend this dedication to be an overt act of relinquishment in perpetuity of all present and future rights to this software under copyright law.

PR #740, opened 2026-08-31
Emit NEAR MSI and Clementine NIR radial optical distortion so the CSM cameras match ISIS

oleg-alexandrov:
Both the NEAR MSI and Clementine NIR drivers used no optical distortion, so their CSM cameras omitted the single-parameter radial distortion the ISIS cameras apply. Each driver now emits it, reusing the existing USGSCSM *RADIAL* model, and subtracts 0.5 from the detector center for the ISIS 0.5-based to CSM 0-based convention. This is the same pattern as the Cassini ISS driver in #739. USGSCSM is unchanged.

It is worth noting that for both of these cameras, just as for the existing Cassini ISS, Dawn and Hayabusa AMICA drivers, the lens distortion coefficient has the opposite sign from the coefficient in the ISIS camera. The USGSCSM radial model subtracts its polynomial where the ISIS model adds it, so the driver emits the negated coefficient and the two models then agree. This is the established pattern that this PR conforms to, and in practice the ISIS and CSM cameras give the same result, which is validated with cam_test.

On real images, cam_test between the ISIS and CSM cameras drops to sub-pixel: about 1e-8 px for NEAR MSI and about 1e-7 px for Clementine NIR.

How to reproduce, NEAR MSI (Clementine NIR is the same flow with *clem2isis*):

```
msi2isis \
  from=$ISISTESTDATA/isis/src/near/apps/msi2isis/tsts/bilinear/input/m0126865998f4_2p_iof.lbl \
  to=msi.cub
spiceinit from=msi.cub
isd_generate -k msi.cub msi.cub
cam_test --image msi.cub --cam1 msi.cub --session1 isis --cam2 msi.json --session2 csm --sample-rate 50
```

The *-k msi.cub* furnishes the kernels recorded by spiceinit.

Done with Claude/AI assistance.


oleg-alexandrov:
Any updates on this? Looks small and self-contained.

ale/drivers/clementine_drivers.py
acpaquette: This should likely be changed to use `RadialDistortion` rather than no distortion. Then update the driver to override the `odtk` property rather than the `usgscsm_distortion_model` property like in the changes to the MSI driver

acpaquette (commented):
Small thing to address, overall changes look good!

ale/drivers/clementine_drivers.py
oleg-alexandrov: Implemented: switched the driver to the *RadialDistortion* mix-in with an *odtk* override, matching the MSI driver. Tests pass.

acpaquette (approved):

